PR #21, opened 2020-12-20
Fix renaming files to same name with different case

stefnotch:
This makes it possible to rename files like `cat` to `cAT`. This change only improves the handling of files.

### DIFF
--- a/Source/Editor/Modules/ContentDatabaseModule.cs
+++ b/Source/Editor/Modules/ContentDatabaseModule.cs
@@ -464,12 +464,6 @@ namespace FlaxEditor.Modules
                 MessageBox.Show("Cannot move folder. Target location already exists.");
                 return;
             }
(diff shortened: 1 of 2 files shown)


mafiesto4:
I'm not sure if I tested it correctly but I cannot rename `cat` material to `cAT`. Does it work only for other files like scripts? Maybe the engine blocks renaming for binary files?
image

stefnotch:
That's odd, it should work for all file types. Does the output log say anything about why the rename failed (C++ side, I think)?

mafiesto4:
Okay the issue I had was related to Win32 error `ERROR_NOT_SAME_DEVICE` (`0x11`) for moving file using temporary proxy file when the project was on a different drive than system. Fixed by 45e86f7591055c37c45f814554b3b3fad5feaa51.

PR #24, opened 2020-12-21
Fix double extensions when user specifies one during asset creation

klukule:
When creating C# asset and manually specifiying extension (`TestScript.cs`) resulting file would be `.cs.cs` this PR should avoid any duplicate extensions.

Although side consequences of this might be some wrongful extension replacement. So we might want to replace (or not append) the extension if file already contains specific extension. Although currently selected method was cleanest and shortest solution.


mafiesto4:
Thanks!

PR #29, opened 2020-12-21
Fix wrong engine path with RegisterEng. as Admin.

jb-perrier:
Fix RegisterEngineLocation registering wrong path when launched as administrator.


mafiesto4:
Thanks!

Issue #65, opened 2020-12-26
New "Json Type" does not detect any acceptable types.


seapulse:
Not sure whats going on here yet. However, I've noticed that the lambda attached to` _element.CustomControl.CheckValid` in 
 `GenericJsonCreateEntry.Options` never gets called.

**Steps to reproduce:**

1. Create new project.
2. Follow instructions [here](https://docs.flaxengine.com/manual/scripting/tutorials/custom-asset.html) to create custom asset type:
![image](https://user-images.githubusercontent.com/22482205/103158688-c403ce80-477d-11eb-80fc-28774daf44ea.png)
3. Attempt to create new asset type in content window and acknowledge there is no detected acceptable types:
![image](https://user-images.githubusercontent.com/22482205/103158680-acc4e100-477d-11eb-8b32-1d5b20fe1cd4.png)
![image](https://user-images.githubusercontent.com/22482205/103158683-afbfd180-477d-11eb-8759-ecf2f177efc2.png)

**Flax version:**
1.0.6214

mafiesto4:
Fixed by #66

PR #66, opened 2020-12-26
Resolve issue where Values.Type resulted in System.Type

seapulse:
Potential fix regarding #65.


DaveTheGameDev:
Had this issue regarding json asset creation as well. This looks like it fixes the issue.

mafiesto4:
Yep, I can confirm that this fixes too. Thanks!

Issue #64, opened 2020-12-25
Clicking back into Flax using sub editor title bar makes window disappear until defualt layout restored


nocxtious:
**Issue description:**

Clicking back into Flax on the title bar (with native windows system enabled) of a sub editor window makes the window disappear and unable to get back to it unless you restore default layout.

**Steps to reproduce:**

1. Open or create a project using the native window system
2. Open a sub editor window (material editor/visual script editor)
3. Click out of the Flax program
4. Click back into Flax clicking on the title bar of the editor window

**Minimal reproduction project:**

1. Open or create a project
2. Enable "Use Native Window System"

**Flax version:**

1.0.6214 (last commit pulled: da389eea03ab0775e384d50ff65bf4d4792f3e47)

https://user-images.githubusercontent.com/72276215/103142292-01088c00-46be-11eb-9b72-ae15d5e67894.mp4

mafiesto4:
Fixed by #70

PR #70, opened 2020-12-28
Correct DockHintWindow drag offset if it was invalid when initializing

VNNCC:
Because of an issue on my side, I had to recreate this pull request.

Description: If the toMove window is unfocused the FlaxEngine.Input.MouseScreenPosition property will always return a vector that only contains zeros. #64

Fix: Perform a later update that will correct the drag offset using the current mouse position.


mafiesto4:
Thanks! 🥇

PR #72, opened 2020-12-28
Support for Visual Studio Code Insiders

VNNCC:
This will add support for the insiders edition from Visual Studio Code.


mafiesto4:
Thanks!

PR #94, opened 2021-01-05
Fixed typos

W2Wizard:


Source/Engine/Particles/Graph/CPU/ParticleEmitterGraph.CPU.h
```diff
@@ -100,7 +100,7 @@ class ParticleEmitterGraphCPU : public ParticleEmitterGraph<VisjectGraph<Particl
 public:
 
     /// <summary>
-    /// Determinates whenever this emitter uses lights rendering.
+    /// Determinate whenever this emitter uses lights rendering.
```
stefnotch, line 103: Shouldn't this be `Determines`?

stefnotch (approved):
Looks great!

Source/Engine/Particles/Graph/CPU/ParticleEmitterGraph.CPU.h
W2Wizard: The word determinates itself is like a stronger word for determines, but its more used in a sense of defining in a final state.

Changed it to `Determine` in the latest commit.

PR #95, opened 2021-01-05
Fixed a ton of typos

VNNCC:


Source/Editor/Surface/Archetypes/Animation.StateMachine.cs
```diff
@@ -1046,7 +1046,7 @@ public void UpdateTransitions()
                         startPos += nrm;
                         endPos += nrm;
 
-                        // Swap fo the other arrow
+                        // Swap for the other arrow
```
stefnotch, line 1049: This is a tricky one, maybe it should be `to`

stefnotch (approved):
I looked over all the changes and it looks great!

Source/Editor/Surface/Archetypes/Animation.StateMachine.cs
VNNCC: That's right, changed it now thanks for the hint!

Issue #97, opened 2021-01-05
Actors not properly flushed on Scene unload


farmisen:
**Issue description:**

The editor crashes if you spawn a prefab, set it up to be destroyed with a delay then stop the game and modify and save the prefab. 

**Steps to reproduce:**

- run something like the following:
```
var bullet = PrefabManager.SpawnPrefab (BulletPrefab, bulletTransform);
Destroy (bullet, 5.0f);
```
- stop the game before the spawned actor got GCed
- edit the prefab
- save the prefab
- notice the error dialog:
![image](https://user-images.githubusercontent.com/229946/103615451-d9000080-4ee7-11eb-8cbd-481c9796b0c2.png)

**Flax version:**
1.0

farmisen:
Crash report:
[Log.txt](https://github.com/FlaxEngine/FlaxEngine/files/5768667/Log.txt)

mafiesto4:
Sorry for later reply. Actually, recently I added checks for prefab objects during prefab changes apply (8beb732cb9a1cd1d7032697e787549dbe4ccf828) so even if there are not yet GCed objects they will be skipped by prefab apply to prevent crash - force flush is fine but won't be needed now (I tested different cases of this bug).

PR #98, opened 2021-01-05
Fixes Issue #97: Actors not properly flushed on Scene unload

farmisen:
Ref: #97 

Calling ForceFlush instead of Flush on Scene unload, changing ForceFlush  to no rely on an arbitrary  duration to force evicting object of the pool.


mafiesto4:
Not needed since the fix for #97 was done the other way.

PR #101, opened 2021-01-07
Added open project directory menu item

W2Wizard:
Added a new menu item entry to open the project file directory in the file explorer.


Source/Editor/Modules/UIModule.cs
```diff
@@ -574,6 +575,11 @@ private void OpenProject()
             }
         }
 
+        private void OpenProjectDirectory()
+        {
+            FileSystem.ShowFileExplorer(Editor.Instance.GameProject.ProjectFolderPath);
```
VNNCC, line 580: Instead of using `Editor.Instance` you should use the read only `Editor` field from the inherited `EditorModule` class.

VNNCC (changes requested):

mafiesto4:
Sorry, but I think we should leave just open project via file picker option.

PR #103, opened 2021-01-08
Include golden ratio constant

W2Wizard:


GoaLitiuM:
Could use more digits than 5. Should be defined as something excessive like `1.6180339887498948482045868343656381177203091798057628621354486227f` and let the compiler cut off all the unnecessary digits.

PR #108, opened 2021-01-09
Adding Convert feature to Actor context menu.

jb-perrier:
You can now convert an actor to another actor type.

Children, transform, tag, layer, ect ... are saved but you will need to re-add your scripts ( probably a later pull request ).


mafiesto4:
Overall it's a very cool idea but I think we need to wrap it inside undo action and support moving scripts too.

jb-perrier:
Ok, i will take a look at this. Just let it open.

jb-perrier:
Now support undo actions & scripts.

Not totally clean, there is two errors on undo (but it works), i'm investigating this.

jb-perrier:
Looks good to me, ready for review.

mafiesto4:
Works great. I added some improvements and tweaks here: https://github.com/FlaxEngine/FlaxEngine/commit/96846ea2551561596f4abe3cf51d2bba0b3d9d68

PR #111, opened 2021-01-13
Added and refactored several extension methods for Random

W2Wizard:
Added and refactored the extension methods for Random to include bools, bytes as well as uniformly distributed random vectors.


Source/Engine/Utilities/Extensions.cs
```diff
@@ -404,5 +238,184 @@ public static void Shuffle<T>(this Random random, IList<T> collection)
...
+        /// <summary>
+        /// Gets a random <see cref="long"/>.
+        /// </summary>
+        /// <param name="random">An instance of <see cref="Random"/>.</param>
+        /// <returns>A random <see cref="long"/>.</returns>
+        internal static long NextLong(this Random random)
+        {
+            var numArray = new byte[8];
+            random.NextBytes(numArray);
+            return (long)(BitConverter.ToUInt64(numArray, 0) & 9223372036854775807L);
```
VNNCC, line 381: For the number you should probably use long.MaxValue to make it more clear.

Source/Engine/Utilities/Extensions.cs
```diff
@@ -404,5 +238,184 @@ public static void Shuffle<T>(this Random random, IList<T> collection)
...
+            do
+            {
+                // Create random float with a mean of 0 and deviation of ±1
+                output.X = NextFloat(random) * 2.0f - 1.0f;
+                output.Y = NextFloat(random) * 2.0f - 1.0f;
+                output.Z = NextFloat(random) * 2.0f - 1.0f;
+
+            } while (output.LengthSquared > 1 || output.LengthSquared < 1e-6f);
+
+            output *= (1.0f / (float)Math.Sqrt(output.LengthSquared));
```
VNNCC, line 314: This could be replaced with the `Vector3.Normalize()` method.

Source/Engine/Utilities/Extensions.cs
```diff
@@ -404,5 +238,184 @@ public static void Shuffle<T>(this Random random, IList<T> collection)
...
+            Vector3 output;
+
+            do
+            {
+                // Create random float with a mean of 0 and deviation of ±1
+                output.X = NextFloat(random) * 2.0f - 1.0f;
+                output.Y = NextFloat(random) * 2.0f - 1.0f;
+                output.Z = NextFloat(random) * 2.0f - 1.0f;
+
+            } while (output.LengthSquared > 1 || output.LengthSquared < 1e-6f);
```
VNNCC, line 312: You should create a new local variable to store the `output.LengthSquared` result, to prevent calculating it more than one time.
The number `1e-6f` could be replaced with the constant variable defined in `Mathf.Epsilon`.

Source/Engine/Utilities/Extensions.cs
```diff
@@ -404,5 +238,184 @@ public static void Shuffle<T>(this Random random, IList<T> collection)
...
+
+            return output;
+        }
+
+        /// <summary>
+        /// Generates a uniformly distributed random unit length vector point on a unit sphere in 2D.
+        /// </summary>
+        /// <param name="random">An instance of <see cref="Random"/>.</param>
+        /// <returns>A random <see cref="Vector2"/>.</returns>
+        public static Vector2 NextVector2(this Random random)
```
VNNCC, line 324: Same changes as in the `NextVector3`.

VNNCC (changes requested):

Source/Engine/Utilities/Extensions.cs
```diff
@@ -404,5 +238,184 @@ public static void Shuffle<T>(this Random random, IList<T> collection)
...
+        /// <param name="random">An instance of <see cref="Random"/>.</param>
+        /// <param name="trueRandom">Should the color be generated from a single random Hue value or separate values for each channel.</param>
+        /// <param name="randomAlpha">Randomize the alpha value.</param>
+        /// <returns>A nice random <see cref="Color"/>.</returns>
+        public static Color NextColor(this Random random, bool trueRandom, bool randomAlpha)
+        {
+            float alpha = randomAlpha ? NextFloat(random) : 1f;
+
+            if (!trueRandom)
+                return Color.FromHSV(NextFloat(random, 0f, 360f), 1f, 1f, alpha);
```
stefnotch, line 352: I'm torn about this one, it's super useful but it might almost deserve its own method?

stefnotch (changes requested):
I'm so sorry for nitpicking so much 😅. 

In general, I really like those changes, since they have so many smart things in them. Like, the `randomRoll` option for the quaternions does seem really convenient.

Source/Engine/Utilities/Extensions.cs
W2Wizard: Adopted the suggestion from @stefnotch instead for Vector2.

Source/Engine/Utilities/Extensions.cs
W2Wizard: Noted & done.

Source/Engine/Utilities/Extensions.cs
W2Wizard: Done.

Source/Engine/Utilities/Extensions.cs
```diff
@@ -404,5 +238,184 @@ public static void Shuffle<T>(this Random random, IList<T> collection)
                 collection[n] = value;
             }
         }
+
+        /// <summary>
+        /// Generates a random <see cref="bool"/>.
+        /// </summary>
+        /// <param name="random">An instance of <see cref="Random"/>.</param>
+        /// <returns>A <see cref="bool"/> thats either true or false.</returns>
+        public static bool NextBool(this Random random) => random.Next(2) == 1;
```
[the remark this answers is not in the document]
W2Wizard: Done.

Source/Engine/Utilities/Extensions.cs
```diff
@@ -404,5 +238,184 @@ public static void Shuffle<T>(this Random random, IList<T> collection)
...
+        public static bool NextBool(this Random random, float weight = 0.5f) => weight >= NextFloat(random);
+
+        /// <summary>
+        /// Generates a random <see cref="byte"/> value between min and max.
+        /// </summary>
+        /// <param name="random">An instance of <see cref="Random"/>.</param>
+        /// <param name="min">The minimum value.</param>
+        /// <param name="max">The maximum value.</param>
+        /// <returns>A <see cref="byte"/> between min and max.</returns>
+        public static byte NextByte(this Random random, byte min = 0, byte max = 2)
```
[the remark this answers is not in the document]
W2Wizard: Done.

Source/Engine/Utilities/Extensions.cs
```diff
@@ -404,5 +238,184 @@ public static void Shuffle<T>(this Random random, IList<T> collection)
...
+        }
+
+        /// <summary>
+        /// Generates a random <see cref="float"/>  value between min and max.
+        /// </summary>
+        /// <param name="random">An instance of <see cref="Random"/>.</param>
+        /// <param name="min">The minimum value.</param>
+        /// <param name="max">The maximum value.</param>
+        /// <returns>A random <see cref="float"/> between min and max.</returns>
+        public static float NextFloat(this Random random, float min = 0.0f, float max = 2.0f)
```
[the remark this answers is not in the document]
W2Wizard: Done.

Added for similar method for double as well to keep it consistent.

Source/Engine/Utilities/Extensions.cs
```diff
@@ -404,5 +238,184 @@ public static void Shuffle<T>(this Random random, IList<T> collection)
...
+        /// <returns>A nice random <see cref="Color"/>.</returns>
+        public static Color NextColor(this Random random, bool trueRandom, bool randomAlpha)
+        {
+            float alpha = randomAlpha ? NextFloat(random) : 1f;
+
+            if (!trueRandom)
+                return Color.FromHSV(NextFloat(random, 0f, 360f), 1f, 1f, alpha);
+
+            return new Color(
+                NextFloat(random, 0f, 255f),
```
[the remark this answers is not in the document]
W2Wizard: Done.

Source/Engine/Utilities/Extensions.cs
```diff
@@ -404,5 +238,184 @@ public static void Shuffle<T>(this Random random, IList<T> collection)
...
+            random.NextBytes(numArray);
+            return (long)(BitConverter.ToUInt64(numArray, 0) & 9223372036854775807L);
+        }
+
+        /// <summary>
+        /// Generates a random normalized 2D direction <see cref="Vector2"/>.
+        /// </summary>
+        /// <param name="random">An instance of <see cref="Random"/>.</param>
+        /// <returns>A random normalized 2D direction <see cref="Vector2"/>.</returns>
+        public static Vector2 NextDirection2D(this Random random)
```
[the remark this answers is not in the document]
W2Wizard: Done.

Source/Engine/Utilities/Extensions.cs
```diff
@@ -404,5 +238,184 @@ public static void Shuffle<T>(this Random random, IList<T> collection)
...
+        {
+            return Vector2.Normalize(new Vector2((float)random.NextDouble(), (float)random.NextDouble()));
+        }
+
+        /// <summary>
+        /// Generates a random normalized 3D direction <see cref="Vector3"/>.
+        /// </summary>
+        /// <param name="random">An instance of <see cref="Random"/>.</param>
+        /// <returns>A random normalized 3D direction <see cref="Vector3"/>.</returns>
+        public static Vector3 NextDirection3D(this Random random)
```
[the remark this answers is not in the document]
W2Wizard: Done.

Source/Engine/Utilities/Extensions.cs
W2Wizard: I'm not sure about it either, It can't be an overloaded method because then the parameters of those 2 functions would be equal as both would have the randomize alpha param.

So would it be worth it making a whole new method just for selecting from HSV instead?

Source/Engine/Utilities/Extensions.cs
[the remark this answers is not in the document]
W2Wizard: Done.

Might follow up with a commit for Vector3 to make it trig based instead.

Source/Engine/Utilities/Extensions.cs
W2Wizard: Done.

Source/Engine/Utilities/Extensions.cs
```diff
@@ -404,5 +238,160 @@ public static void Shuffle<T>(this Random random, IList<T> collection)
...
+        public static bool NextBool(this Random random, float weight = 0.5f) => random.NextDouble() < weight;
+
+        /// <summary>
+        /// Generates a random <see cref="byte"/> value between min and max.
+        /// </summary>
+        /// <param name="random">An instance of <see cref="Random"/>.</param>
+        /// <param name="min">The minimum value.</param>
+        /// <param name="max">The maximum value.</param>
+        /// <returns>A random <see cref="byte"/> between min and max.</returns>
+        public static byte NextByte(this Random random, byte min = 0, byte max = byte.MaxValue)
```
stefnotch, line 257: This is a sneaky off-by-one bug, since the maximum is exclusive. Meaning `random.Next(0, 255);` won't ever return 255.

Source/Engine/Utilities/Extensions.cs
stefnotch: I guess it might be worthwhile to make a whole new method for a random HSV color.

stefnotch (commented):
I really like those random methods now 😄

Source/Engine/Utilities/Extensions.cs
W2Wizard: Done.

Source/Engine/Utilities/Extensions.cs
W2Wizard: Fixed.

Source/Engine/Utilities/Extensions.cs
```diff
@@ -256,7 +256,7 @@ public static void Shuffle<T>(this Random random, IList<T> collection)
         /// <returns>A random <see cref="byte"/> between min and max.</returns>
         public static byte NextByte(this Random random, byte min = 0, byte max = byte.MaxValue)
```
stefnotch, line 257: This is tricky, since a byte cannot be larger than `255`, however, the usual convention is to have an *exclusive* upper bound. I guess this part of why C# doesn't have a random byte method with a range as a parameter? 🤔

W2Wizard: Could you elaborate, because if you look into the definition of the method you can see that I've adjusted the `max` part by incrementing it by one.

stefnotch: Yep, I meant that the method's inputs are off by one. 
Like, `random.NextByte(0, 2)` would return 0, 1 or 2.
Meanwhile, `random.Next(0, 2)` would return 0 or 1.

W2Wizard: Ah alright, should be it hopefully with this commit 😄

PR #143, opened 2021-01-23
Improve build times in development configuration

GoaLitiuM:
These changes will disable the slowest optimization flags in development builds, leading to much faster build times with no noticeable performance loss in editor or games. I believe the development builds should hit the perfect middle ground in both these factors, to be fast for iterative builds (like debug configuration) without sacrificing too much runtime performance (like release configuration).

Right now building the engine in development configuration takes as much time as it would take to build it in release configuration. The main culprits for these slow build times in this case are the Whole Program Optimization and Link-Time Code Generation flags.

This commit changes the default build flags for development builds but also exposes the Whole Program Optimization flag and prevents using linker optimization flags when LTCG is enabled (otherwise the linker would show tons of confusing warnings when optimizations and LTCG are both enabled at the same time).

### DIFF
--- a/Source/Tools/Flax.Build/Build/NativeCpp/CompileEnvironment.cs
+++ b/Source/Tools/Flax.Build/Build/NativeCpp/CompileEnvironment.cs
@@ -57,6 +57,11 @@ namespace Flax.Build.NativeCpp
         public bool Optimization = false;
 
         /// <summary>
+        /// Enables the whole program optimization.
+        /// </summary>
+        public bool WholeProgramOptimization = false;
+
+        /// <summary>
         /// Enables functions level linking support.
         /// </summary>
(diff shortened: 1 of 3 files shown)


mafiesto4:
I tested those changes on:
`i7 CPU 16GB RAM SSD`

Editor build times:

**Before**
Debug rebuild: 01:14s
Debug build after 1 file change: 00:04s
Development full rebuild: 01:38s
Development build after 1 file change: 00:55s

**After**
Debug rebuild: 01:18s
Debug build after 1 file change: 00:04s
Development rebuild: 02:31s
Development build after 1 file change: 00:04s

Debug are just ok (actually it's unchanged). Development full rebuild is slower (due to codegen during cpp files compilation) but the iterative link times are now blazing fast so let's merge it!

Issue #128, opened 2021-01-20
More High DPI Screen Bugs


Evildea:
Hi,

The following code suffers from High DPI issues:

Screen.CursorLock = CursorLockMode.Locked;

![image](https://user-images.githubusercontent.com/35228400/105252601-dd0f6c00-5bd1-11eb-852d-b55b0ead8bc8.png)

PR #145, opened 2021-01-23
Fixed issue High DPI Issue

Evildea:
Fixes the issue I lodged in this ticket: https://github.com/FlaxEngine/FlaxEngine/issues/128

### DIFF
--- a/Source/Editor/Surface/Archetypes/Material.cs
+++ b/Source/Editor/Surface/Archetypes/Material.cs
@@ -642,6 +642,22 @@ namespace FlaxEditor.Surface.Archetypes
                     NodeElementArchetype.Factory.Output(0, "XYZ", typeof(Vector3), 0),
                 }
             },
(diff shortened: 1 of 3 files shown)


stefnotch:
Yes, this does fix the issue in the editor. Basically, the `SetMousePosition` function currently expects a mouse position in the "local space" of the game window. This means it expects a DPI scaled position `Input::SetMousePosition(Screen::GetSize() * 0.5f * Platform::GetDpiScale());`.

However, in a built game, it calls a different method, which apparently does not expect a DPI scaled mouse-position. Right now, I'm not entirely sure how to correctly fix it. 
https://github.com/FlaxEngine/FlaxEngine/blob/049450e31b5f4bc45fac80a5d9ea744accd50e18/Source/Engine/Engine/Engine.cpp#L419-L426

Evildea:
I guess we'll need to wait for an actual fix.

PR #161, opened 2021-01-25
Added Remap node for all graphs

W2Wizard:
Added the remap node since it was listed on trello specifically for the material editor. However, I also went ahead and added it for all graphs since its a general purpose node. The node is under the math category.

**Shadergraph:**

https://user-images.githubusercontent.com/63303990/105702427-1286c300-5f0c-11eb-838e-aa7a1c20ab3d.mp4

**Visjectgraph:**
image

### DIFF
--- a/Source/Engine/Visject/VisjectGraph.cpp
+++ b/Source/Engine/Visject/VisjectGraph.cpp
@@ -378,13 +378,10 @@ void VisjectExecutor::ProcessGroupMath(Box* box, Node* node, Value& value)
         const Vector2 rangeA = tryGetValue(node->GetBox(1), node->Values[1]).AsVector2();
         const Vector2 rangeB = tryGetValue(node->GetBox(2), node->Values[2]).AsVector2();
 
+        auto mapFunc = rangeB.X + (inVal - rangeA.X) * (rangeB.Y - rangeB.X) / (rangeA.Y - rangeA.X);
+        
         // Clamp value?
-        if (node->Values[3].AsBool) 
-        {
-            value = Math::Clamp(rangeB.X + (inVal - rangeA.X) * (rangeB.Y - rangeB.X) / (rangeA.Y - rangeA.X), rangeB.X, rangeB.Y);
-            break;
-        }
-        value = rangeB.X + (inVal - rangeA.X) * (rangeB.Y - rangeB.X) / (rangeA.Y - rangeA.X);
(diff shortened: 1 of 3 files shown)


Source/Engine/Visject/VisjectGraph.cpp
```diff
@@ -371,6 +371,19 @@ void VisjectExecutor::ProcessGroupMath(Box* box, Node* node, Value& value)
...
+    case 48:
+    {
+        const float inVal  = tryGetValue(node->GetBox(0), node->Values[0]).AsFloat;
+        const Vector2 rangeA = tryGetValue(node->GetBox(1), node->Values[1]).AsVector2();
+        const Vector2 rangeB = tryGetValue(node->GetBox(2), node->Values[2]).AsVector2();
+
+        auto mapFunc = rangeB.X + (inVal - rangeA.X) * (rangeB.Y - rangeB.X) / (rangeA.Y - rangeA.X);
+        
+        // Clamp value?
+        value = node->Values[3].AsBool ? Math::Clamp(mapFunc, rangeB.X, rangeB.Y) : mapFunc;
```
mafiesto4, line 384: Instead of using `node->Values[3].AsBool` call `tryGetValue` like for other inputs to handle both default value and custom connected value.

Source/Engine/Visject/VisjectGraph.cpp
```diff
@@ -371,6 +371,19 @@ void VisjectExecutor::ProcessGroupMath(Box* box, Node* node, Value& value)
         if (value.Type.Type == VariantType::Enum)
             value.AsUint64 = value.AsUint64 | (uint64)tryGetValue(node->GetBox(1), Value::Zero);
         break;
+
```
mafiesto4, line 374: Missing `// Remap` comment to mark the node type impl section.

mafiesto4 (changes requested):

Source/Engine/Visject/VisjectGraph.cpp
W2Wizard: Applied

Source/Engine/Visject/VisjectGraph.cpp
W2Wizard: Applied

mafiesto4 (approved):

PR #173, opened 2021-01-27
Add Switch on Int node

W2Wizard:
### Switch on Int
This PR is for the switch on int node listed on trello, the archetype and all is setup however currently i'm facing issues with implementing the execution for the output pins.

I have a backup solution however I don't want to implement it only as a last result since its a pretty weak solution for just a single node.

Basically the main issue is that I am unsure how to exactly determine the right output pin, using the box ID won't work since it would collide with the other input pins since they can have a value that can be selected via the switch.

Another solution would be to use the name of the box but this cannot be done either as I can't retrieve the name :P

I will have to keep working on it until I figure out a solution.

https://user-images.githubusercontent.com/63303990/106056724-4706a480-60ef-11eb-9f1e-d3c1b9ab53e9.mp4

### DIFF
--- a/Source/Editor/Surface/Archetypes/Flow.cs
+++ b/Source/Editor/Surface/Archetypes/Flow.cs
@@ -168,18 +168,18 @@ namespace FlaxEditor.Surface.Archetypes
 
                 var outPinCount = (int)Values[0];
                 var outPinMin = 0;
-                var outPinMax = 100; // Should be enough or maybe too much?
+                var outPinMax = 101; // Should be enough or maybe too much?
 
                 var startIndex = (int)Values[2];
 
                 // Remove default, if null nothing will happen.
-                RemoveElement(GetBox(101));
+                RemoveElement(GetBox(102));
 
                 // Add all boxes
                 for (int i = 0; i < outPinCount; i++)
                     AddBox(true, i + 1, i, (startIndex + i).ToString(), new ScriptType(typeof(void)), true);
 
-                AddBox(true, 101, outPinCount, "Default", new ScriptType(typeof(void)), true);
+                AddBox(true, 102, outPinCount, "Default", new ScriptType(typeof(void)), true);
 
                 // Cull boxes
                 for (int i = outPinCount; i <= outPinMax; i++)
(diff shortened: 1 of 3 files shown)


stefnotch:
One alternative solution would be to go back to the drawing board and redesign the node. No idea if this is a good solution, just throwing ideas out there.

e.g.
image

mafiesto4:
I like the initial idea. Suggestions:
- Add double-click to edit label value for the switch case instead of using `Start`
- Using `+` button just add the case for the next item but prevents adding already added case
- Use `byte[]` for data storage in node `Values` array (example: `MethodOverrideNode` from https://github.com/FlaxEngine/FlaxEngine/blob/master/Source/Editor/Surface/Archetypes/Function.cs#L492)
- In C++ backend simply use `MemoryReadStream` to read switch data from `byte[]` (example: `Invoke Method` from https://github.com/FlaxEngine/FlaxEngine/blob/master/Source/Engine/Content/Assets/VisualScript.cpp#L618)
- Ensure to add version number in the beginning of the data blog to support upgrading it in future

W2Wizard:
https://user-images.githubusercontent.com/63303990/106364645-c636eb00-6330-11eb-8858-4502e0b0401b.mp4

Ok so I sort of got it working (as in the UI and a bit of backend) to the suggestion of @stefnotch but I have to be honest here for a sec and say that the way the whole Visject thing is implemented is really really giving me mixed thoughts. I'm not sure wether this was due to constraint in using mono maybe or something that just had to be managed like this.

Theres a great amount of things I would completely redo from the way the Frontend is implemented, in which all nodes are just mushed together into a single file and collected into an array, each file being the category in which the nodes reside. Instead of treating each node as a seperate class inherting from the base node archetype and using attributes to determine extra info like category.

To the backend where the node's implementation is determined by just a big massive switch statement (I would probably just handle the nodes function directly in the archetype itself maybe), both of which I don't believe are very future proof or a general good way of implementing nodes.

Maybe in the future Visject will change (since were still 1.0 here, and the whole surface stuff is probably still wip) but the way its currently implemented I would sort of be worried if its continued to being maintained like that :confused:, its probably not as a big of a priority right now since mafiesto has so many things to work on but I hope its atleast being considered.

W2Wizard:
I'll close for now and reopen later at some point.

stefnotch:
Yes, I do agree that the current implementation feels less than ideal and eventually deserves a second look. In fact, while writing a [custom visject plugin](https://github.com/FlaxCommunityProjects/flax-custom-visject-plugin), I stumbled upon the same issue. Back then, I ended up deciding that I'm not sure how to design the API better and just left it. But, if you eventually want to give it a shot, hit me up :)

Issue #165, opened 2021-01-26
Material Bug with Plane


Evildea:
There are two minor bugs with the material editor.

The first bug is that when you change the view to plane. The view is facing away from the plane making it seem like your material has just randomly disappeared. You can see this in the video I attached.

The second issue is that there seems to be a weird lighting artifact on the plane. This is probably caused by the way the light hits the plane, but it is very distracting to the user. This can also be seen in the attached video.

https://drive.google.com/file/d/1SspeN-28uQDZern4w3L0QLNTvfTE6yUh/view?usp=sharing

stefnotch:
The second one is apparently because the default plane has odd normal vectors

![image](https://user-images.githubusercontent.com/10220080/106055287-77e5da00-60ed-11eb-9a0e-64a5aa3b5b86.png)

mafiesto4:
Fixed by a226ff87623533989cab90d60efd91f66ca15408 and 50f6a01f83a36824d6bc011f1d186d6bda08621c

PR #175, opened 2021-01-27
Fix Issue #165: Weird normals on plane primitive

W2Wizard:
Fixes issue #165 via 2 ways:
- Removes weird normals, simply imported a fresh new plane from blender.
- Rotates the plane correctly so that it isn't facing the wrong direction anymore.

Plane had some weird normal thing on the top left: (Image by Stefnotch)
image

Its gone: (visual effect is just a shader)
image


W2Wizard:
Fixed by mafiesto

PR #185, opened 2021-01-30
Add Sign, Any, All and Blackbody shader nodes

W2Wizard:
Probably the last shader node PR? Who knows!

While I was adding the nodes I also wanted to establish some form of consistency, for all the previous nodes since it was a bit here and there, by making box inputs as `const auto` and other things like shader functions as `auto`.

This PR adds all these nodes which were listed on trello:
-Sign
-Any
-All
-Blackbody

image

Blackbody Example:

https://user-images.githubusercontent.com/63303990/106369608-29d21000-6353-11eb-8c8d-a02ad406547d.mp4

### DIFF
--- a/Source/Engine/Tools/MaterialGenerator/MaterialGenerator.Material.cpp
+++ b/Source/Engine/Tools/MaterialGenerator/MaterialGenerator.Material.cpp
@@ -435,6 +435,8 @@ void MaterialGenerator::ProcessGroupMaterial(Box* box, Node* node, Value& value)
         // Blackbody
     case 35:
     {
+        // Based on unity's implementation by using data gathered by Mitchell Charity.
+
         const auto temperature = tryGetValue(node->GetBox(0), node->Values[0]).AsFloat();
 
         // Value X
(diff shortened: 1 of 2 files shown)


Source/Engine/Tools/MaterialGenerator/MaterialGenerator.Material.cpp
```diff
@@ -385,28 +385,76 @@ void MaterialGenerator::ProcessGroupMaterial(Box* box, Node* node, Value& value)
...
+    {
+        const auto inValue = tryGetValue(node->GetBox(0), 0, Value::Zero);
+
+        value = writeLocal(ValueType::Bool, String::Format(TEXT("all({0})"), inValue.Value), node);
+        break;
+    }
+        // Blackbody
+    case 35:
+    {
+        // Based on unity's implementation by using data gathered by Mitchell Charity.
```
mafiesto4, line 438: You mean this reference: http://www.vendian.org/mncharity/dir3/blackbody/ ?

W2Wizard: Yeah thats the one. I found an implementation on Unity's docs and it was the only good & working source that I could find regarding this node, other sources were either on websites that were completely outdated with broken links.

mafiesto4: Ok, could you replace that comment with sth like:
`// Reference: Mitchell Charity, http://www.vendian.org/mncharity/dir3/blackbody/`

W2Wizard: Done 👍🏻

mafiesto4 (approved):

Issue #214, opened 2021-02-06
DebugDraw.DrawLines Crashes the Editor


Evildea:
**Issue description:**
Using the following code with a broken number of points will results in an editor crash:
`DebugDraw.DrawLines(Path, Transform.GetWorld(), Color.Red);`

![image](https://user-images.githubusercontent.com/35228400/107107148-64700700-6883-11eb-82b8-d5505c52c601.png)

It shouldn't crash the entire editor but simply throw an output error during run-time.

W2Wizard:
Issue can be closed, one down, plenty more to go!

PR #215, opened 2021-02-06
Fix issue #214: Debug draw with uneven amount of lines crashes

W2Wizard:
The lines array has to have an even amount of items since for example [0] & [1] count as a single line but when we have an uneven amount we can't draw [2] & [??] as there is no target.

Having an assertion for something like that is a bit of harsh way of ensuring that the array is even, instead simply logging an error and not drawing the lines is probably less annoying.

Also just some very slight amount of formatting to make it a bit more readable.

### DIFF
--- a/Source/Engine/Debug/DebugDraw.cpp
+++ b/Source/Engine/Debug/DebugDraw.cpp
@@ -644,7 +644,7 @@ void DebugDraw::DrawLines(const Span<Vector3>& lines, const Matrix& transform, c
 
     if (depthTest) 
         list = duration > 0 ? &DebugDrawDepthTest.DefaultLines : &DebugDrawDepthTest.OneFrameLines;
-    else 
+    else
         list = duration > 0 ? &DebugDrawDefault.DefaultLines : &DebugDrawDefault.OneFrameLines;
 
     list->EnsureCapacity(list->Count() + lines.Length());
--- a/Source/Engine/Debug/DebugDraw.cpp
+++ b/Source/Engine/Debug/DebugDraw.cpp
@@ -19,6 +19,7 @@
 #include "Engine/Graphics/RenderBuffers.h"
 #include "Engine/Animations/AnimationUtils.h"
 #include "Engine/Profiler/Profiler.h"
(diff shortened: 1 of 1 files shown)


stefnotch:
I suppose throwing an exception to the C# side might also work https://github.com/FlaxEngine/FlaxEngine/blob/7e4afbc4ff31b333523934c6272cd7e5ca6ae689/Source/Engine/Debug/DebugLog.cpp#L167-L210

Source/Engine/Debug/DebugDraw.cpp
```diff
@@ -629,18 +630,24 @@ void DebugDraw::DrawLine(const Vector3& start, const Vector3& end, const Color&
 
 void DebugDraw::DrawLines(const Span<Vector3>& lines, const Matrix& transform, const Color& color, float duration, bool depthTest)
 {
-    ASSERT(lines.Length() % 2 == 0);
+    if (lines.Length() % 2 == 0)
```
SilentCLD, line 633: Shouldn't this be `if (lines.Length() % 2 != 0)` ?

W2Wizard: No, you want the remainder to be zero for the division, if there is a remainder then it means the num of elements are uneven. As described, if this is the case you can have a start value but no end which is not valid.

stefnotch: Here you're checking if the line length is *invalid* and then throw an error accordingly.
Basically `if(odd) { throw }`. Which means that the remainder should not be zero...

SilentCLD: > if there is a remainder then it means the num of elements are uneven
Right, but you're checking if there isn't a remainder, so if the array size is correct it'll throw

W2Wizard: Oops I read too quickly!

stefnotch:
Can I just say thank you so much for actually taking your time to hunt down those issues and graciously provide fixes for them! 😄

PR #224, opened 2021-02-09
[C++] Added FLAXENGINE_API macro on SettingsBase class

Vizepi:
Using any type derived from `SettingsBase` class is causing the link to fail because of this missing symbol:
`Array<SettingsBase*> SettingsBase::Containers`

The link fails because the symbol is not exported in FlaxEngine.lib.
Adding `FLAXENGINE_API` macro before `SettingsBase` drives the compiler/librarian to embed the symbol in the library as expected.

### DIFF
--- a/Source/Engine/Core/Config/LayersTagsSettings.h
+++ b/Source/Engine/Core/Config/LayersTagsSettings.h
@@ -3,6 +3,7 @@
 #pragma once
 
 #include "Engine/Core/Config/Settings.h"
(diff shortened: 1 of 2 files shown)


Vizepi:
Oh and just so you know, I created the pull request with only Settings.h in it. I didn't know that pushing from local to remote while a PR is in progress was adding the file to the PR.

Vizepi:
HOLD ON before merging! This has to be discussed

Issue #153, opened 2021-01-24
[BUG] Physics layers matrix is incorrect


Vizepi:
**Issue description:**
When editing the Physics Settings, the Layers matrix is incorrect.
One of the row or column order should be inverted.
Maybe it is not incorrect code-wise, but it is for sure not displayed properly.

**Steps to reproduce:**
- Open a project
- Open Game/Content/Settings/Physics Settings.json
- See the matrix is not correct

**Flax version:**
1.0.6215

**More Infos:**
![image](https://user-images.githubusercontent.com/9052852/105643736-ee869b80-5e91-11eb-953c-7f0ff2d04503.png)
You can see on this image that it is not possible to edit layer interactions between *Layer 3* and *Player*, between *Layer 3* and *Bullets* or between *Bullets* and *Bullets*. The expected matrix is the following:
![image](https://user-images.githubusercontent.com/9052852/105643817-7076c480-5e92-11eb-862d-72c199a130d3.png)

mafiesto4:
Fixed by #227

PR #227, opened 2021-02-10
[EDITOR] Fixed layer matrix order ( issue #153 )

Vizepi:
Inverts the order of filling of the top right names so it matches the matrix.
I tested that the behavior is correct and that the right names are inverted:
image
Bottom cubes are lines, top cubes are column


Vizepi:
Details about the screenshot:
given layers A B Cand D, the bottom cubes have the folowing layers:
```
A A A A
B B B
C C
D
```
and the top cubes have layers
```
A B C D
A B C
A B
A
```
Top cubes starts 2 meters in the air and goes through or collide the bottom cubes.
We can see that the collision correspond to the layer matrix we have defined.

PR #232, opened 2021-02-11
Update README.md

stefnotch:
I think the FlaxVS plugin is required with Visual Studio for building Flax from the source. But I might be mistaken...


mafiesto4:
Nope, it's optional. Required only for C# debugging and starting Flax from VS with C#-only projects.
We use `VisualStudioInstance.HasFlaxVS` to detect if it's installed and skip generating project files for it: https://github.com/FlaxEngine/FlaxEngine/blob/master/Source/Tools/Flax.Build/Projects/VisualStudio/VisualStudioInstance.cs#L33.

PR #241, opened 2021-02-14
Fix UI - mouse capturing and tree node hovering

stefnotch:
This PR fixes some UI issues that I've stumbled upon while hunting down other UI issues

1. When you have a secondary monitor to the left, the mouse coordinates can be negative. Which causes issues for `StartMouseCapture(true)`
2. When hovering over TreeNodes, the `DefaultNodeOffsetY` caused some small spaces between the nodes where you couldn't hover. 
3. When hovering over TreeNodes, you couldn't hover below a node when you added a child to the node and removed it again.


mafiesto4:
Oh this fix for secondary monitor on the left is so helpful!

PR #242, opened 2021-02-15
Render2D Triangle Functions

jb-perrier:
Render2D Triangle funcs.


Source/Engine/Render2D/Render2D.h
```diff
@@ -339,4 +340,29 @@ DECLARE_SCRIPTING_TYPE_NO_SPAWN(Render2D);
...
     /// <param name="blurStrength">The blur strength defines how blurry the background is. Larger numbers increase blur, resulting in a larger runtime cost on the GPU.</param>
     API_FUNCTION() static void DrawBlur(const Rectangle& rect, float blurStrength);
+
+    /// <summary>
+    /// Draws vertices array.
+    /// </summary>
+    /// <param name="t">The texture.</param>
+    /// <param name="vertices">The vertices array.</param>
+    /// <param name="uvs">The uvs array.</param>
+    API_FUNCTION() static void DrawVertices(GPUTexture* t, Array<Vector2> vertices, Array<Vector2> uvs);
```
mafiesto4, line 350: I would name it `DrawTexturedTriangles` and use `const Span<Vector2>& vertices` +` const Span<Vector2>& uvs` to make it more optimized (calling from C# won't allocate memory but just pass span as a pointer to fixed C# buffer).

Source/Engine/Render2D/Render2D.h
```diff
@@ -339,4 +340,29 @@ DECLARE_SCRIPTING_TYPE_NO_SPAWN(Render2D);
...
+    
+    /// <summary>
+    /// Draws vertices array.
+    /// </summary>
+    /// <param name="vertices">The vertices array.</param>
+    /// <param name="colors">The colors array.</param>
+    /// <param name="useAlpha">If true alpha blending will be enabled.</param>
+    API_FUNCTION() static void DrawVertices(Array<Vector2> vertices, Array<Color> colors, bool useAlpha);
```
mafiesto4, line 358: I would name it `FillTriangles` and use `const Span<Vector2>& vertices` +` const Span<Color>& colors` to make it more optimized (calling from C# won't allocate memory but just pass span as a pointer to fixed C# buffer).

mafiesto4 (changes requested):

jb-perrier:
I'm on macos right now, so i'm unable to compile it. But it should be good.

jb-perrier:
Looks like not, in checking it tonight.

Issue #167, opened 2021-01-26
Tooltips break when not using the default theme.


Thev2Andy:
**Issue description:**
When using custom themes, the tooltip can glitch out.

**Steps to reproduce:**
1) Change the theme of the editor to the exact same as in the picture:
![theme](https://user-images.githubusercontent.com/69793166/105873213-ebabb800-6003-11eb-9d93-d966fa1574a5.png)
2) Hold the mouse on a file / icon to bring up the tooltip.

**Minimal reproduction project:**
Any project works.

**Flax version:**
Occurs on version 1.0.6125.

**Working tooltips:**
![working](https://user-images.githubusercontent.com/69793166/105873683-696fc380-6004-11eb-9556-581933694c08.png)

**Broken tooltips:**
![not working](https://user-images.githubusercontent.com/69793166/105874001-c10e2f00-6004-11eb-8b1a-4dde320d7efe.png)

mafiesto4:
Could you attach your editor options with that theme options? File: `%appdata%/Flax/EditorOptions.json`

Thev2Andy:
Sure:
[EditorOptions.zip](https://github.com/FlaxEngine/FlaxEngine/files/5877917/EditorOptions.zip)

stefnotch:
I suppose we might want to separate the shared tooltip object from the style?

Thev2Andy:
on other colors it works fine tho

stefnotch:
Investigated, it's because the background color of the shared tooltip has an alpha value of 1.

```json
"SharedTooltip": {
...
         "BackgroundColor": {
              "R": 0.06666667,
              "G": 0.06666667,
              "B": 0.06666667,
              "A": 1.0
            },
}
```

This leads to Tooltip.cs drawing that color over everything else https://github.com/FlaxEngine/FlaxEngine/blob/f1763840c936a643ed785ffe061415888087c5a5/Source/Engine/UI/GUI/Tooltip.cs#L217-L227


https://github.com/FlaxEngine/FlaxEngine/blob/f1763840c936a643ed785ffe061415888087c5a5/Source/Engine/UI/GUI/Control.cs#L435-L442

jb-perrier:
Solved by #246 .

PR #246, opened 2021-02-17
Tooltip background.

jb-perrier:
Remove the call to Control.Draw since background are already drawn by Tooltip.Draw.

Issue #167 .


stefnotch:
The alternate solution would be to move the `base.Draw()` further up. But yeah, this is a nice fix 👍

jb-perrier:
It wouldn't make sense since it will be over-drawn anyway.

PR #248, opened 2021-02-18
Add localization services to the editor

W2Wizard:
# Localization? What is this PR?

Firstly, I'd like to thank @VNNCC for helping me out with the build and deploy stuff!

This PR should add the ability to localize strings using JSON files in order to support almost any language! Currently the entire engine is hardcoded to english, and with this PR it would allow to support any language.

The internal `EditorLocalization` class provides a singleton `Manager` that allows access to the services provided by the class itself that can be used through the editor, however not for games. *A restart is required to apply a language change*. 

Additionally we had to make quite a bit of work to the `Builder.NativeCPP` to support all the moving of JSON files from A to B which includes creating a whole new type of dependency that allows for a bit more control over how they get handled. Finally we cleaned up some additional code here and there.

### TODOs:
- [ ] Simplify access instead of `EditorLocalization.Manager.GetValue()` it should be smth like `Localization.GetValue()`
- [ ] Swap to C++, change C# side with Internal function calls instead
- [ ] Use JSON Assets instead of JSON Files
- [ ] Introduce namespaces

### Languages
- [ ] English
- [ ] German
- [ ] Polish
- [ ] French
- [ ] Others...

### Conversions
It will take some additional time to replace all existing strings into localized ones.
Currenlty a whopping 0.01% is localized!

As of now, no language is fully implemented at all, this will take the most time.

## How would you go on about adding a localized string?
We can access the string instead depending on which language is selected:

```cs
// Currently for example the File context menu is setup like this.
MenuFile = MainMenu.AddButton("File");

// Instead we simply use the manager to return the value of the given key!
MenuFile = MainMenu.AddButton(EditorLocalization.Manager.GetValue("T_CM_FILE"));

```

To add a value we need to go the language files under `FlaxEngine->Editor->Localization->Languages` there you will find the appropriate languages supported by Flax. Next you simply add a new element:
image

All that is needed now is to just build and that should be really it.
Just gotta make sure you have selected the correct language under `Options->Interface` which shows all the supported languages.

image

Final result:
image


stefnotch:
That's super neat! Since [internationali(s|z)ation is hard](https://www.youtube.com/watch?v=0j74jcxSunY&lc=UggRBiSU9e12QHgCoAEC), thought I'd ask how cases like the following should be handled

1. EditorDisplay: Sets the group and then groups with the same name get merged
2. Tooltip
3. Name of the property, will be shown in the editor (with spaces)

image

W2Wizard:
> That's super neat! Since [internationali(s|z)ation is hard](https://www.youtube.com/watch?v=0j74jcxSunY&lc=UggRBiSU9e12QHgCoAEC), thought I'd ask how cases like the following should be handled
> 
> 1. EditorDisplay: Sets the group and then groups with the same name get merged
> 2. Tooltip
> 3. Name of the property, will be shown in the editor (with spaces)
> 
> image

I will most likely have to add an Internal secondary constructor for attributes that use text which needs to be localized and instead specify a key instead of directly the text itself.

A little example here:
image

Source/Editor/Editor.cs
```diff
@@ -297,8 +298,14 @@ internal void Init(bool isHeadless, bool skipCompile)
             _modules.Sort((a, b) => a.InitOrder - b.InitOrder);
             _isAfterInit = true;
 
+            // Options module has to init first for localization.
+            _modules[0].OnInit();
+
+            Log("Launching localization service...");
+            EditorLocalization.Manager.SelectLanguage(Options.Options.Interface.Language);
```
mafiesto4, line 305: Maybe initialize `EditorLocalization` after loading the options in `OptionsModule`?

Source/Editor/Modules/UIModule.cs
```diff
@@ -382,7 +383,7 @@ private void InitMainMenu(RootControl mainWindow)
             MainMenu.Parent = mainWindow;
 
             // File
-            MenuFile = MainMenu.AddButton("File");
+            MenuFile = MainMenu.AddButton(EditorLocalization.Manager.GetValue("T_CM_FILE"));
```
mafiesto4, line 386: We could use some utility to make those texts easier to replace, eg.
```cs
Loc.T("T_CM_FILE")
```

Source/Tools/Flax.Build/Build/Graph/TaskGraph.cs
```diff
@@ -82,6 +82,7 @@ public Task AddCopyFile(string dstFile, string srcFile)
             if (Platform.BuildPlatform.Target == TargetPlatform.Windows)
             {
                 task.CommandPath = "xcopy";
+
```
mafiesto4, line 85: Leftover line.

Source/Editor/Surface/VisjectSurface.ContextMenu.cs
```diff
@@ -221,7 +221,7 @@ public virtual void ShowSecondaryCM(Vector2 location, SurfaceControl controlUnde
 
             _cmFormatNodesConnectionButton = menu.AddButton("Format node(s)", () =>
             {
-                FormatGraph(SelectedNodes);
+                //FormatGraph(SelectedNodes);
```
mafiesto4, line 224: leftover.

Source/Editor/Modules/UIModule.cs
stefnotch: For what it's worth, one really funky alternative that I've seen (ab)uses extension methods. e.g.
```cs
"T_CM_FILE".Loc()
```

Source/Editor/Surface/VisjectSurface.ContextMenu.cs
W2Wizard: Yeah that one for some reason keeps giving an error in VS, got fed up and just commented it out.

mafiesto4:
Great stuff! I have some more notes to throw:
- we might move `Localization Manager` into C++ and Engine as a separate module to make it reusable for both game and editor
- instead of Localization subfolder with binaries (cool idea) we would prefer separate assets with localization data tables (eg. imported from `.po` files or created in Editor)
- to make it more scalable let's go with namespaces for localization texts (eg. `Editor.General.File`, `Game.Menu.Pause`)
- we will need some helper utilities to improve localization in code (eg. suggested `Loc.T(name)` but similar stuff for C++ localization)
- editor can keep this languages enum and pick the language based on user locale, although let's use local under the hood as enum is more limited

Finally, let's track this feature here <https://trello.com/c/tSRK555N/274-games-localization> -> feel free to suggest more stuff.

honzapatCZ:
Needed to say, this will add 1. step of troubleshooting to be, "Switch to English editor"

W2Wizard:
Ok so I completely managed to mess this up in just 2 seconds...

New draft :man_facepalming:

stefnotch:
@W2Wizard Alternative would be resetting the branch to before you created those two commits and force pushing to your fork. But as always, force pushing is dangerous business.

PR #254, opened 2021-02-20
Add some defaults to Quaternion helpers

honzapatCZ:


Source/Engine/Core/Math/Quaternion.cs
```diff
@@ -975,7 +975,7 @@ public static void LookAt(ref Vector3 eye, ref Vector3 target, ref Vector3 up, o
         /// <param name="target">The camera look-at target.</param>
         /// <param name="up">The camera's up vector.</param>
         /// <returns>The created look-at quaternion.</returns>
-        public static Quaternion LookAt(Vector3 eye, Vector3 target, Vector3 up)
+        public static Quaternion LookAt(Vector3 eye, Vector3 target, Vector3 up  = Vector3.Up)
```
jb-perrier, line 978: The default value must be a compile time constant, that's why it's not possible to set a custom value, C# limitation.
There is a workaround using nullable value but it's not very clean imo.

Source/Engine/Core/Math/Quaternion.cs
```diff
@@ -999,7 +999,7 @@ public static void RotationLookAt(ref Vector3 forward, ref Vector3 up, out Quate
         /// <param name="forward">The camera's forward direction.</param>
         /// <param name="up">The camera's up vector.</param>
         /// <returns>The created look-at quaternion.</returns>
-        public static Quaternion RotationLookAt(Vector3 forward, Vector3 up)
+        public static Quaternion RotationLookAt(Vector3 forward, Vector3 up = Vector3.Up)
```
jb-perrier, line 1002: The default value must be a compile time constant, that's why it's not possible to set a custom value, C# limitation.
There is a workaround using nullable value but it's not very clean imo.

Source/Engine/Core/Math/Quaternion.cs
```diff
@@ -1011,7 +1011,7 @@ public static Quaternion RotationLookAt(Vector3 forward, Vector3 up)
         /// <param name="forward">The forward direction. Direction to orient towards.</param>
         /// <param name="up">Up direction. Constrains y axis orientation to a plane this vector lies on. This rule might be broken if forward and up direction are nearly parallel.</param>
         /// <returns>The calculated quaternion.</returns>
-        public static Quaternion LookRotation(Vector3 forward, Vector3 up)
+        public static Quaternion LookRotation(Vector3 forward, Vector3 up = Vector3.Up)
```
jb-perrier, line 1014: The default value must be a compile time constant, that's why it's not possible to set a custom value, C# limitation.
There is a workaround using nullable value but it's not very clean imo.

Source/Engine/Core/Math/Quaternion.cs
honzapatCZ: fixed

Source/Engine/Core/Math/Quaternion.cs
honzapatCZ: fixed

Source/Engine/Core/Math/Quaternion.cs
honzapatCZ: fixed

jb-perrier:
A struct is a value type, you can't set it to null without declaring it at nullable : Vector3?

honzapatCZ:
reamde it with overloading

jb-perrier:
You have to wait mafiesto, i don't have any more rights than reviewing your code.

mafiesto4:
Looks good but it doesn't handle case if `forward` vector is actually `Up`, here is an example that calculates Up vector for `Actor.Direction` setter:

https://github.com/FlaxEngine/FlaxEngine/blob/68b4fdb9c3f6a23caaf8e4d509d307e0d6168b6e/Source/Engine/Level/Actor.cpp#L588-L603

It could be a utility method in Vector3, like: `CalculateUp(Vector3 forward)` to make it reusable.

honzapatCZ:
> Looks good but it doesn't handle case if `forward` vector is actually `Up`, here is an example that calculates Up vector for `Actor.Direction` setter:
> 
> https://github.com/FlaxEngine/FlaxEngine/blob/68b4fdb9c3f6a23caaf8e4d509d307e0d6168b6e/Source/Engine/Level/Actor.cpp#L588-L603
> 
> It could be a utility method in Vector3, like: `CalculateUp(Vector3 forward)` to make it reusable.

What should I handle? I just added overloads for defaults. As up is usually up

mafiesto4:
`Up` is relative to the forward in the transformations we have in math lib. In the proposed code if uses pass `forward=Vector3.Up` then the calculations will end up in some nans/inf because `forward==up`.

honzapatCZ:
> `Up` is relative to the forward in the transformations we have in math lib. In the proposed code if uses pass `forward=Vector3.Up` then the calculations will end up in some nans/inf because `forward==up`.

? I still dont understand, i jus added defalt where up is up. Or iam just not seeing the problem? Or is it like up is in math equivalent to forward or whatever? If yes then thats misskeading at least

mafiesto4:
I guess you could simply swap hardcoded `Vector3.Up` with `Dot(forward, Up) < 0.999f ? Up : Forward` so it won't break if someone passes `forward=Vecto3.Up` 🐱

Issue #247, opened 2021-02-17
HSV to RGB shader nodes


stefnotch:
**Issue description:**
It would be really useful to have `HSV to RGB` and `RGB to HSV` shader nodes. For example, generating a rainbow becomes quite easy that way.

Some simple references regarding how the functions could look can be found here
https://www.shadertoy.com/view/MsS3Wc
https://www.shadertoy.com/view/4lVBW1

**Flax version:**
Flax 1.0

W2Wizard:
I couldn't resist not doing this :D

Coming Soon™
![image](https://user-images.githubusercontent.com/63303990/108523783-9bf0a180-72ce-11eb-85c5-9c99f3994e91.png)

mafiesto4:
Fixed by #256

PR #256, opened 2021-02-20
Add RGB & HSV Conversion nodes

W2Wizard:
Implement suggestion #247 by @stefnotch.

This took me a while but here we go.

image


stefnotch:
Oh yeah, fun fact: Writing something like `fix: #247` in your pull request has the effect of auto-closing that issue when the pull request gets merged.

PR #257, opened 2021-02-20
Add Vector3 Clamp

honzapatCZ:
useful for eg, making sure moving in diagonal doesnt gain more speed then moving just horizontally or just vertically

### DIFF
--- a/Source/Engine/Core/Math/Vector3.cs
+++ b/Source/Engine/Core/Math/Vector3.cs
@@ -1001,20 +1001,32 @@ namespace FlaxEngine
         /// <param name="max">Max Length</param>
         public static Vector3 ClampLength(Vector3 vector, float min, float max)
         {
-            Vector3 retVect = new Vector3(vector.X, vector.Y, vector.Z);
-            if (retVect.LengthSquared > max * max)
+            ClampLength(vector, min, max, out Vector3 retVect);
+            return retVect;
+        }
+        /// <summary>
+        /// Makes sure that Length of the output vector is always below max and above min.
+        /// </summary>
+        /// <param name="vector">Input Vector.</param>
+        /// <param name="min">Min Length</param>
+        /// <param name="max">Max Length</param>
+        /// <param name="retVect">The Return Vector</param>
+        public static void ClampLength(Vector3 vector, float min, float max, out Vector3 retVect)
+        {
+            float lenSq = vector.LengthSquared;
+            if (lenSq > max * max)
             {
-                float scaleFactor = max / (float)Math.Sqrt(retVect.LengthSquared);
-                retVect.X = retVect.X * scaleFactor;
-                retVect.Y = retVect.Y * scaleFactor;
-                retVect.Z = retVect.Z * scaleFactor;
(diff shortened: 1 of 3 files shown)


stefnotch (commented):
Pretty nice addition, I can definitely see that function being useful.

Some minor nitpicks and things that could eventually be improved
- What should happen when the vector has a length of zero?
- The mathematics APIs usually use `ref` for such vectors and then have an overloaded method that calls the `ref` variant under the hood
- To keep the API consistent, someone is going to have to port this to `Vector2` and `Vector4`.

honzapatCZ:
I dont see problem with zero,
hmm ok
technically just Vector4 will suffice, but ok

Source/Engine/Core/Math/Vector3.cs
```diff
@@ -983,7 +983,41 @@ public static Vector3 Normalize(Vector3 value)
...
+            if (retVect.LengthSquared > max * max)
+            {
+                float scaleFactor = max / (float)Math.Sqrt(retVect.LengthSquared);
+                retVect.X = retVect.X * scaleFactor;
+                retVect.Y = retVect.Y * scaleFactor;
+                retVect.Z = retVect.Z * scaleFactor;
+            }
+            if (retVect.LengthSquared < min * min)
+            {
+                float scaleFactor = min / (float)Math.Sqrt(retVect.LengthSquared);
```
W2Wizard, line 1014: Perhaps retVect should store its LengthSquared value once instead.

mafiesto4: Yeah, cache `vector.LengthSquared` for the input and reuse it in code to optimize perf.

honzapatCZ:
Someone should squash my commits once it compiles and btw its no so good idea to edit code from web github. XD

stefnotch:
(By the way, if you clone your repo to your PC, you can squash the commits and then do a force push)

stefnotch:
Also, no idea if this is worth mentioning, another slightly different way of writing this would be
```cs
public static void ClampLength(ref Vector4 vector, float min, float max, out Vector4 retVect)
{
	float lenSq = retVect.LengthSquared;
	if (lenSq > max * max)
	{
		float scaleFactor = max / (float)Math.Sqrt(lenSq);
		retVect.X = vector.X * scaleFactor;
		retVect.Y = vector.Y * scaleFactor;
		retVect.Z = vector.Z * scaleFactor;
		retVect.W = vector.W * scaleFactor;
	}
	else if (lenSq < min * min)
	{
		float scaleFactor = min / (float)Math.Sqrt(lenSq);
		retVect.X = vector.X * scaleFactor;
		retVect.Y = vector.Y * scaleFactor;
		retVect.Z = vector.Z * scaleFactor;
		retVect.W = vector.W * scaleFactor;
	} 
	else 
	{
		retVect.X = vector.X;
		retVect.Y = vector.Y;
		retVect.Z = vector.Z;
		retVect.W = vector.W;
	}
}
```

honzapatCZ:
ok squashed

Issue #262, opened 2021-02-21
Problem with "Get Node Transform (local space)" and "Get Node Transform (model space)"


Evildea:
There is an issue with the Get Node Transforms.
They are missing their execution pins.

![image](https://user-images.githubusercontent.com/35228400/108615338-1b10e180-7457-11eb-91e7-a0b98a523652.png)

I tried to fix the source myself by changing the .CS to:
`NodeElementArchetype.Factory.Output(0, string.Empty, typeof(void), 1),`

I also updated the .CPP to reflect the new structure but clearly I'm doing something wrong as the transform never seems to work correctly.
![image](https://user-images.githubusercontent.com/35228400/108615448-4a741e00-7458-11eb-9819-4703fa4a4c98.png)

mafiesto4:
The idea behind the `Get Node Transform` nodes (local and model space) is that they don't use this white flow arrow which contains a bones transformations. Here is an example of usage:

![image](https://user-images.githubusercontent.com/3464853/108628180-cb7fe500-7459-11eb-869a-d5b68369936f.png)

The white pins are not execution flow signals like in Visual Script but the value flow like in Material Graph.

stefnotch:
Would it be reasonable to change the icon of the white pins in the animation graph, since their meaning is quite different?

mafiesto4:
I guess yeah, we could use a bone icon for this or an animated pose.

Evildea:
I like the idea of using an animated pose icon. I assumed it worked exactly like the visual scripting because they looked the same. Thanks for clearing that up.

stefnotch:
By the way, if anyone is interested in fixing this, here is the relevant part of the source code
https://github.com/FlaxEngine/FlaxEngine/blob/87a290b45c73d90b0eec5fa89a3a6d4a56603638/Source/Editor/Surface/Elements/Box.cs#L515-L519

Issue #127, opened 2021-01-20
Bugs with Get Node Transforms


Evildea:
These two nodes have the following issues:

1. Their bounding boxes are too small.
2. They have no secondary execute pin. This means you can execute them but not connect them in a sequence of nodes which practically makes them useless.

Bounding Box is too small:
![image](https://user-images.githubusercontent.com/35228400/105247392-d5979500-5bc8-11eb-8799-26dc5315faff.png)
![image](https://user-images.githubusercontent.com/35228400/105247404-daf4df80-5bc8-11eb-99d6-9764ff68f4a8.png)

Execution isn't possible:
![image](https://user-images.githubusercontent.com/35228400/105247623-2f985a80-5bc9-11eb-942b-a1577b9c7829.png)

mafiesto4:
Fixed by #131

Evildea:
@mafiesto4 This node still doesn't have an output pin: 
![image](https://user-images.githubusercontent.com/35228400/108613287-87ceb080-7444-11eb-8d9e-79364e9c70be.png)

mafiesto4:
But this node is auto-executed when you use the transform output box. It samples the incoming input bone.

PR #266, opened 2021-02-21
Fix missing pins for Get Node Transform

W2Wizard:
A commit removed the fix provided by #127, which caused an issue to popup again mentioned in #262.


W2Wizard:
Seems like thats not the only issue, when closing and re-opening the animation graph the connect is undone

mafiesto4:
See https://github.com/FlaxEngine/FlaxEngine/issues/262#issuecomment-782867521

PR #277, opened 2021-02-25
Add localization for Editor

W2Wizard:
# Localization? What is this PR?

`NOTE: I'll make sure to actually check on which branch I am next time I update my fork...`

Firstly, I'd like to thank @VNNCC for helping me out with the build and deploy stuff!

This PR should add the ability to localize strings using JSON files in order to support almost any language! Currently the entire engine is hardcoded to english, and with this PR it would allow to support any language.

The internal `EditorLocalization` class provides a singleton `Manager` that allows access to the services provided by the class itself that can be used through the editor, however not for games. *A restart is required to apply a language change*. 

Additionally we had to make quite a bit of work to the `Builder.NativeCPP` to support all the moving of JSON files from A to B which includes creating a whole new type of dependency that allows for a bit more control over how they get handled. Finally we cleaned up some additional code here and there.

### TODOs:
- [x] Simplify access instead of `EditorLocalization.Manager.GetValue()` it should be smth like `Localization.GetValue()`
- [ ] Swap to C++ a module, change C# side with Internal function calls instead 
- [ ] Use JSON Assets instead of JSON Files
- [x] Introduce namespaces

### Languages
- [ ] English
- [ ] German
- [ ] Polish
- [ ] French
- [ ] Others...

### Conversions
It will take some additional time to replace all existing strings into localized ones.
Currenlty a whopping 0.01% is localized!

As of now, no language is fully implemented at all, this will take the most time.

## How would you go on about adding a localized string?
We can access the string instead depending on which language is selected:

```cs
// Currently for example the File context menu is setup like this.
MenuFile = MainMenu.AddButton("File");

// Instead we simply use the manager to return the value of the given key!
MenuFile = MainMenu.AddButton(Localization.GetValue("T_CM", "FILE"));

```

To add a value we need to go the language files under `FlaxEngine->Editor->Localization->Languages` there you will find the appropriate languages supported by Flax. Next you simply add a new element:
image

All that is needed now is to just build and that should be really it.
Just gotta make sure you have selected the correct language under `Options->Interface` which shows all the supported languages.

image

Final result:
image

### DIFF
--- a/Source/Editor/Localization/LanguageFormats.cs
+++ b/Source/Editor/Localization/LanguageFormats.cs
@@ -15,16 +15,16 @@ namespace FlaxEditor.LocalizationServices
         /// <summary>
         /// The Polish language
         /// </summary>
-        Polish, //TODO: Add support
+        Polish,
 
         /// <summary>
         /// The German language
         /// </summary>
-        German, //TODO: Add support
+        German,
 
         /// <summary>
         /// The French language
         /// </summary>
-        French  //TODO: Add support
+        French
     }
 }
--- a/Source/Editor/Localization/LanguageFormats.cs
+++ b/Source/Editor/Localization/LanguageFormats.cs
@@ -1,0 +1,30 @@
+// Copyright (c) 2012-2021 Wojciech Figat. All rights reserved.
+
+namespace FlaxEditor.LocalizationServices
+{
+    /// <summary>
+    /// All supported languages the editor provides
+    /// </summary>
+    public enum Languages
+    {
+        /// <summary>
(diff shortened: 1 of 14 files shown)


mafiesto4:
Editor localization might come later in future. Currently we focus on game and have new localization system here:
https://github.com/FlaxEngine/FlaxEngine/tree/master/Source/Engine/Localization
with docs:
https://github.com/FlaxEngine/FlaxDocs/blob/1.2/manual/editor/localization/index.md
for `1.2` version. Thanks for helping to rush this features earlier in the Flax!

PR #283, opened 2021-02-27
Add Render2D Tint

jb-perrier:
Add Render2D Tint stack feature.

Default tint is White.

### DIFF
--- a/Source/Engine/Render2D/Render2D.cpp
+++ b/Source/Engine/Render2D/Render2D.cpp
@@ -812,11 +812,11 @@ void Render2D::PopClip()
     OnClipScissors();
 }
 
-void Render2D::PushTint(const Color& tint)
+void Render2D::PushTint(const Color& tint, bool inherit)
 {
     RENDER2D_CHECK_RENDERING_STATE;
     
-    TintLayersStack.Push(tint * TintLayersStack.Peek());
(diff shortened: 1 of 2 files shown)


Source/Engine/Render2D/Render2D.cpp
```diff
@@ -795,6 +812,25 @@ void Render2D::PopClip()
     OnClipScissors();
 }
 
+void Render2D::PushTint(const Color& tint)
+{
+    RENDER2D_CHECK_RENDERING_STATE;
+    
+    TintLayersStack.Push(tint);
```
stefnotch, line 819: Should nested tints override the current tint or should they be combined?

jb-perrier: I think it should just override the previous one. Combining multiple tints would give dark results. But i can add it if needed.

mafiesto4: Yeah, let's make it inherited from the current tint stack even if it would make a dark result. For instance, one container could use it to fade pixels to back a little, and another child would use alpha tint to be less transparent so in the inherited case it will work fine.
Simply multiple tint by the current tint from the stack when adding it there.

jb-perrier: Pushed 🥳

PR #284, opened 2021-02-27
Editor viewport improvements

jb-perrier:
Changes :

- Camera Orientations are now Viewpoints ( Blender way ).
- Support Perspective and Orthographic projections.
- Viewpoints use currently selected actors, world origin otherwise.
- Add Mouse Wheel sensitivity in Viewport option.
- Mouse wheel change Ortho scale.
- Change move speed (RMB+Wheel) / Move using wheel cannot happen in the same time now.
- Focus actor ( F shortcut ) : Support Orthographic, preserve camera orientation.


jb-perrier:
Ready for review, imo some logic in EditorViewport class can be moved to FPsCamera/ViewportCamera. But that's probably for another day.

PR #297, opened 2021-03-03
Update .gitignore

stefnotch:
When you execute the [`PackageAll.bat`](https://github.com/FlaxEngine/FlaxEngine/blob/master/PackageAll.bat) file, Flax creates some `Package_` folders in the root directory. This excludes those folders from git.

It also removes some things from the `.gitignore` that seem like they could be removed without any issues

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ Cache/
 Binaries/
 Logs/
 Source/*.csproj
+/Package_*/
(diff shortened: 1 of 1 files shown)


honzapatCZ:
Note that Package_* is also created with PackagePlatforms/PackageEditor not just PackageAll, but PackageAll is the only useful thing

Issue #301, opened 2021-03-04
Property drawing order in inspector should respect order as in source + inheritance property order


honzapatCZ:
This was talked over on discord, specially that its not easy
I know that but its a real problem
AND
Inheritance
If you inherit something, it makes sense that your new properties are drawn at the start/end else it gets put somewhere in between which is, atleast, unpractical.
If reflection provides incosistent results, which I doubt, could we like make own parser?

UPDATE:
Looks like MemberInfo.MetadataToken could be the way to go, gonna investigate that and possibly make a PR

Evildea:
I second this. It's rather annoying having all your properties come in to the editor in alphabetical order.

honzapatCZ:
> I second this. It's rather annoying having all your properties come in to the editor in alphabetical order.

Yep, please up the PR itself too

honzapatCZ:
Btw @mafiesto4  looks like unity uses just plain GetMembers, without any sorting etc., you shouldn't rely on that, but its not mandatory
https://answers.unity.com/questions/822926/how-does-unity-sort-fields-in-the-order-of-their-d.html

mafiesto4:
#792

PR #302, opened 2021-03-04
Declaration sorting of script members

honzapatCZ:
Its not perfect, handles
fields and auto generated properties
I believe there is no built-in way to sort normal properties as this information is absent from the resulting assembly
related: #301

### DIFF
--- a/Source/Editor/CustomEditors/Editors/GenericEditor.cs
+++ b/Source/Editor/CustomEditors/Editors/GenericEditor.cs
@@ -173,8 +173,16 @@ namespace FlaxEditor.CustomEditors.Editors
                             return string.Compare(Display.Group, other.Display.Group, StringComparison.InvariantCulture);
                     }
 
-                    // By name
-                    return string.Compare(Info.Name, other.Info.Name, StringComparison.InvariantCulture);
+                    // By declaration order
+                    if (Info.MetadataToken > other.Info.MetadataToken)
+                        return 1;
+                    else if (Info.MetadataToken < other.Info.MetadataToken)
+                        return -1;
+                    else
+                    {
+                        // By name
+                        return string.Compare(Info.Name, other.Info.Name, StringComparison.InvariantCulture);
+                    }
                 }
 
                 return 0;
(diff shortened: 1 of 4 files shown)


honzapatCZ:
After merge, dont close this nor the issue, as this needs more improvements, note that its still better than the current alphabetical ordering which is just useless

stefnotch:
Ideally this would be a conditional setting, maybe one that can be enabled by switching on an editor option.
(Like mafiesto4 suggested on Discord)

honzapatCZ:
> Ideally this would be a conditional setting, maybe one that can be enabled by switching on an editor option.
> (Like mafiesto4 suggested on Discord)

What about voting from users?. 
Cuzz like get this, currently you get the alphabet which you:
1) don't care, so you wont care about declaration order either
2) use order attributes to change the order however you want, which you can still do

stefnotch:
I had those sort of things in mind when I suggested the setting 😉 If it proves to be really popular and we figure out how to make the code really bulletproof, it could become a default setting.

For now, as you noticed, it has issues with normal properties and I'm not quite sure if we should make something potentially unstable a default.

Source/Editor/CustomEditors/Editors/GenericEditor.cs
```diff
@@ -173,16 +173,16 @@ public int CompareTo(object obj)
                             return string.Compare(Display.Group, other.Display.Group, StringComparison.InvariantCulture);
                     }
 
-                    // By declaration order
-                    if (Info.MetadataToken > other.Info.MetadataToken)
-                        return 1;
-                    else if (Info.MetadataToken < other.Info.MetadataToken)
-                        return -1;
-                    else
+                    if(Editor.Instance.Options.Options.General.ScritpMembersOrder == Options.GeneralOptions.MembersOrder.Declaration)
```
stefnotch, line 176: Typo here, it should probably say `Script`

honzapatCZ: XD fixed

mafiesto4:
Thanks!

PR #303, opened 2021-03-04
Image should care just about image stuff

honzapatCZ:
MouseOverColor on Image is only used on single thing and without reason overcomplicates the process of image.
The single thing was PlatformWindow, which I fixed by making Platform image, which could further be replaced with GameButton from my FlaxGameUI Iam going to hopefully finish soon(tm)
HidenColor is also a thing which image doesnt have to care about, so I removed it, again a thing that's more expected on a button


RuanLucasGD:
@mafiesto4  Can this modification be accepted?

Issue #299, opened 2021-03-04
Controls Properties should show based on anchors


honzapatCZ:
Controls Location should hide based on anchors, get replaced with left/right/top/bottom offset.
Gonna look at that myself, shouldn't be that hard.

Basically the logic is as follows:
If X anchors(min, max) are equal you wanna show X Position and Width, else offset from left and right
If Y anchors(min, max) are equal you wanna show Y Position and Height, else offset from top and bottom

Result is uncluttered UI

PR #309, opened 2021-03-06
Better UI Control editor

honzapatCZ:
Editor now shows controls like location, size and offsets dependant on anchor configuration
related #299 
image


Evildea:
We totally need this! This will make my life so much easier :D

honzapatCZ:
> We totally need this! This will make my life so much easier :D

Mafi promised to check it out this weekend so maybe even 1.1

PR #324, opened 2021-03-09
Character Controller Value Dependency fix

honzapatCZ:
Step Offset can never be higher than height of character controller
This way if one changes above the other, the other gets adjusted.
-> No crashes just because of that YAY

### DIFF
--- a/Source/Engine/Physics/Colliders/CharacterController.cpp
+++ b/Source/Engine/Physics/Colliders/CharacterController.cpp
@@ -47,6 +47,9 @@ void CharacterController::SetHeight(const float value)
 
     _height = value;
 
+    if (GetStepOffset() > value)
+        SetStepOffset(value);
+
     UpdateSize();
     UpdateBounds();
 }
@@ -77,6 +80,9 @@ void CharacterController::SetStepOffset(float value)
         return;
 
     _stepOffset = value;
+
+    if (GetHeight() < value)
+        SetHeight(value);
 
     if (_controller)
         _controller->setStepOffset(value);


Source/Engine/Physics/Colliders/CharacterController.cpp
```diff
@@ -78,6 +81,9 @@ void CharacterController::SetStepOffset(float value)
 
     _stepOffset = value;
 
+    if (GetHeight() < value)
+        SetHeight(value);
```
stefnotch, line 85: Wait, why would modifying the step offset change the *height*?

honzapatCZ: Because step offset can never be higher than height, if they would be, PhysX throws error, engine asserts that and you get a crash. If you managed to save before that you can easily get soft-locked and need to manually edit the scene file to resolve the issue.

stefnotch: Dunno, maybe this is just personal preference but I'd almost expect the step offset to get clamped to the height.

honzapatCZ: That was the way it inicially was, but on the other hand, you wouldnt know why the value doesnt change, this way you see what changed because of that.

jb-perrier: It can be specified in the tooltip.

honzapatCZ: it is, but tooltip wont fix soft-lock in case this is invalid and you save the scene

mafiesto4:
Good catch to find this issue. I think we should use `Math::Min(_stepOffset, desc.height)` in `CreateActor` when setting up the actor and adjust code in `SetStepOffset` and `UpdateSize` to perform clamping too.
Editing other property in the setter is not the preferred way. Also, I guess we could apply scale to the step offset too as it's related to the character size.

honzapatCZ:
> Good catch to find this issue. I think we should use `Math::Min(_stepOffset, desc.height)` in `CreateActor` when setting up the actor and adjust code in `SetStepOffset` and `UpdateSize` to perform clamping too.
> Editing other property in the setter is not the preferred way. Also, I guess we could apply scale to the step offset too as it's related to the character size.

OK so should I clamp its own values in setter? ok

honzapatCZ:
I suppose this is already fixed??????

mafiesto4:
Yeah fixed in e6d5d5330e568a555e05b3dc194e30608cf0a106

Issue #333, opened 2021-03-11
Game crash after changing window mode to fullscreen from code


michalzembron:
**Issue description:**
Whenever I'm changing window mode by "Screen.IsFullscreen = true;" game crashes. I tried switching from dx11 to dx12 and the same thing happens (GTX 1050 notebook). Also I tried to force launch the game with dedicated graphics, but It didnt help at all and I checked this on another machine with GTX 1660 Ti.
![dx11](https://user-images.githubusercontent.com/59204659/110787790-14afa180-826e-11eb-9065-32ed3f024065.png)
![dx12](https://user-images.githubusercontent.com/59204659/110788028-5dfff100-826e-11eb-949a-87c236cacc14.png)

**Steps to reproduce:**
1. Create a scene with script attached to any component.
2. In OnStart write "Screen.IsFullscreen = true;".
3. Run Game Cooker and build the game.
4. Run the game and wait for crash.

**Minimal reproduction project:**
Any project with a basic scene.

**Flax version:**
1.0

Evildea:
I just tested this on my computer and also got the same error message:
![image](https://user-images.githubusercontent.com/35228400/110789935-887c9900-82c4-11eb-88e0-83a18983b691.png)

I'm using:
- DirextX 12
- Nividia GeForce GTX 1060

jb-perrier:
Should be solved by #336 . Keep us updated !

PR #336, opened 2021-03-11
Fix fullscreen dx11/dx12.

jb-perrier:
Fix the fullscreen crash on dx11/dx12.
Little tweak, we can now use ALT + ENTER to go fullscreen !

#333

### DIFF
--- a/Source/Engine/GraphicsDevice/DirectX/DX11/GPUSwapChainDX11.cpp
+++ b/Source/Engine/GraphicsDevice/DirectX/DX11/GPUSwapChainDX11.cpp
@@ -208,7 +208,7 @@ bool GPUSwapChainDX11::Resize(int32 width, int32 height)
         ASSERT(_swapChain);
 
         // Disable DXGI changes to the window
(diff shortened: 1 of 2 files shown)


honzapatCZ:
You sure it wasnt disabled intentionally? From the looks of it it seems like the code is specifically there to disable it.

jb-perrier:
Since fullscreen works great now, i don't see any reason to not enable ALT + ENTER, it's a very common way to switch to fullscreen.

honzapatCZ:
> Since fullscreen works great now, i don't see any reason to not enable ALT + ENTER, it's a common way to switch to fullscreen.

Well you removed some code which checks buffers or whatever, it would be good to know what that does.

jb-perrier:
SetFullscreenState trigger a WM_SIZE event that is hooked by the window resizing code. So we don't need to manually resize buffers here ( that's why switching to fullscreen was not working before ).

mafiesto4:
Tested and works great. Thnaks. I guess the previous behavior was trying to resize the window to the fullscreen size but Win32 layer handles this for us to some degree.

PR #339, opened 2021-03-13
AOT on Windows

honzapatCZ:
Note that this does not currently work, it generated the AOTed dlls and that's all

Why?: AOT on desktop can bring better peformance and a bit of security(atleast you will get rid of the quick scripty kids with DnSpy)
Note that this should be optional, bcz in some cases you want your game easily modable etc. etc.


honzapatCZ:
anyone who knows how to make it help, I have no idea

LCRW:
Not convinced this would actually improve performance, windows automatically NGens's .NET assemblies after the first few runs anyway, AOT is not mono specific like you think, NGen and ilasm have been about at least since .NET 1.1. Windows also adds a performance gain because the .NET Framework is already loaded by the system. Whereas mono aot would have to initialise the mono framework as a dependency. AOT has been available in dot net since release, it was just left undocumented for years. Some proof of performance gain would be good to see.

honzapatCZ:
> Not convinced this would actually improve performance, windows automatically NGens's .NET assemblies after the first few runs anyway, AOT is not mono specific like you think, NGen and ilasm have been about at least since .NET 1.1. Windows also adds a performance gain because the .NET Framework is already loaded by the system. Whereas mono aot would have to initialise the mono framework as a dependency. AOT has been available in dot net since release, it was just left undocumented for years. Some proof of performance gain would be good to see.

Well thats a question, what are the chances the flax target runtime is already loaded by the system? And performance gain? of course even flax has AOT on Mobile bcz performance gain

PR #355, opened 2021-03-17
Separate ContainerControl Self/Children draw code.

### LINKED ISSUE #312: Three more UI Bugs (Ordering, Auto-width, and Anchor)
**Issue description:**
image

Three UI Bugs
1. Draw Order of UI Elements doesn't appear right.
2. Auto-width and Auto-Height don't seem to affect parent sprite which results in text getting cut.
3. Anchor's sometimes seem to break the entire viewer.

**Steps to reproduce:**
To expose these three bugs, simply follow the steps in this film:
https://drive.google.com/file/d/1A4NLk9PpiK0-fhxN_8q8S4MjCleTn6Wu/view?usp=sharing

**Flax version:**
Branch 1.1

jb-perrier:
Enable us to change how `ContainerControl` and his children are rendered.
If you inherits from `ContainerControl` you can now override `DrawSelf` for custom drawing in place of `Draw`.

`ContainerControl.Draw` is now responsible for draw ordering, we can override it to create funky things like overlay.

Default order : 
image

Example from `ProgressBar` (now use `DrawSelf`): 
image
*Call `base.DrawSelf` in place of `base.Draw`.*

For simple `Control`, you can just use `Draw` like before.

**This is not a breaking change, custom `ContainerControl` that override `Draw` will continue to work, but i recommend you to modify them in order to use `DrawSelf`** 

*Fix #312 Image order issue

### DIFF
--- a/Source/Engine/UI/GUI/Common/Border.cs
+++ b/Source/Engine/UI/GUI/Common/Border.cs
@@ -5,7 +5,7 @@ namespace FlaxEngine.GUI
     /// <summary>
     /// Border control that draws the border around the control edges (inner and outer sides).
     /// </summary>
-    public class Border : Control
+    public class Border : ContainerControl
     {
         /// <summary>
         /// Gets or sets the color used to draw border lines.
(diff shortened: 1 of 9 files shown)


mafiesto4:
Side note: I had to revert the clipping change for children in 960a122550523ce7c02d71392c03cbe13dab6bf1 as it wasn't implemented in `DrawChdilren` what lead to issues for other UI controls like Panels.

PR #365, opened 2021-03-18
Select Prefabs First

honzapatCZ:
Selects prefabs first and then goes slowly one by one down
Video

### DIFF
--- a/Source/Editor/Gizmo/TransformGizmo.cs
+++ b/Source/Editor/Gizmo/TransformGizmo.cs
@@ -50,6 +50,44 @@ namespace FlaxEditor.Gizmo
         {
         }
 
+        /// <summary>
+        /// Helper function, recursivily finds the Prefab Root of node or null
+        /// </summary>
+        /// <param name="node">The node from which to start</param>
+        /// <returns></returns>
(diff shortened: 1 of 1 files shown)


Source/Editor/Gizmo/TransformGizmo.cs
```diff
@@ -50,6 +50,44 @@ public TransformGizmo(IGizmoOwner owner)
...
+                return null;
+        }
+
+        /// <summary>
+        /// Recursively walks up from the node up to ceiling node(inclusive) or selection(exclusive)
+        /// </summary>
+        /// <param name="node">The node from which to start</param>
+        /// <param name="ceiling">The ceiling(inclusive)</param>
+        /// <returns></returns>
+        public ActorNode RecursiveWalkUpAndFindSomethingBeforeSelectedOrSupplied(ActorNode node, ActorNode ceiling)
```
stefnotch, line 76: That's probably the longest method name I've ever seen 😂 

At the very least, we can drop the `Recursive`, since that's an implementation detail and replace the `Something` with `ActorNode`

honzapatCZ: renamed

PR #366, opened 2021-03-18
UICanvas Culling

jb-perrier:
UICanvas is now frustum culled using AABB.


mafiesto4:
Thanks, works like a charm. Probably easier to implement that writing that previous `TODO` comment xD

jb-perrier:
Haha, but maybe frustum colling was not a thing at the time.

PR #417, opened 2021-04-02
UI LocalLocation

honzapatCZ:
Now it even makes sense, uses Pivot and Anchors to determine location. Makes editing UI like 1000x easier.


mafiesto4:
Cool stuff! After we add some gizmo handles and visual hint to show control pivot and anchors editing UI will be even more smooth.

mafiesto4:
I added some cleanup in dff1374378012b89d6ae2eae04d43dcbbea4c767 to use fields directly (faster) and added check to prevent crashes when control has no parent set.

PR #419, opened 2021-04-02
Add Initializer lists support for arrays

W2Wizard:
Something I consider to be quite essential and kind of surprised this wasn't there so I added it.

Allows you do initilize arrays in C++ like this:

```cpp
// For instance in a .h file
API_FIELD() Array<int> foo = {1, 2, 3, 4};
```


mafiesto4:
This might be useful.
Also, keep in mid the compilation time when adding new includes. Hopefully for `initializer_list` it's not that bad.
(https://artificial-mind.net/projects/compile-health/)
image

(CL on PR checks doesn't show any regression in compile time, cool)

PR #420, opened 2021-04-03
Change FontReference to reference type

GoaLitiuM:
FontReference is mutating itself by storing the cached font for later reuse, but this is not working as intended because FontReference is returned as a value type in UIControls so the cached font is actually stored in a copy of the FontReference and not in the original property.

I am not entirely sure why the original code works in some cases, but without this change, the text in TextBox sometimes flickers when garbage collection is triggered (once every few seconds), and this can be verified by forcing garbage collection for every frame which returns in TextBox not rendering properly or at all. Again I'm not sure about the exact conditions when this occurs, but for me it happens when the font size of the TextBox is reasonably large (>100pt or so).

### DIFF
--- a/Source/Editor/Windows/Assets/FontWindow.cs
+++ b/Source/Editor/Windows/Assets/FontWindow.cs
@@ -135,7 +135,7 @@ namespace FlaxEditor.Windows.Assets
         /// <inheritdoc />
         protected override void UnlinkItem()
         {
-            _textPreview.Font = new FontReference();
+            _textPreview.Font = null;
 
             base.UnlinkItem();
         }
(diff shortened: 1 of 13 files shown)


stefnotch:
For completeness sake, two alternatives would be 
- making FontReference` a `readonly struct`
- using `ref struct`

GoaLitiuM:
Readonly struct would remove the caching though. I think I tried ref structs but those can't be used as properties.

mafiesto4:
Partially-merged in 3c2b1a232c35383c8e3e1292512e3d0619a73953 - I skipped `FLAX_EDITOR` as those change the UI behavior at runtime and UI looks different when created from code in cooked game. We still have to tackle fonts/styling issues in cooked game bit that's for another day.

PR #424, opened 2021-04-04
Better Game UI(Reorganization, hiding useless stuff, ui navigation, game specific ui elements, gui action(so called game events))

honzapatCZ:
1) this PR is big so I hope @mafiesto4  will have a good night read
2) this PR is based off of #417  as I want to work with the ui in fast way and the current absolute positioning obsolete

So this PR can be split to few groups
# 1a) Game Events
They allow you to graphically assign functions
(They are not that much tied to UI but I think they are nice addition in this dynamic kind of approach)(and yes I know the editor for them is a bit ugly, I will restructure it soon but atm I wanted to add it as its a dependency of GameButton)
image

# 1b) AlphaPanel
Similiar to TintPanel(if its already in 1.2 atleast it should be) with the option to ignore or inherit the parent tint alpha

# 1c) Horizontal/Vertical Panels resize children
One thing I expected from Horizontal/Vertical Panels is that they will takeup some space and then will dynamically change all children to fit their size, but i couldnt find so its there in their respective panels. (so now you can both set parent by children or set children by parent(not both at the same time))

# 2) I restructured the GUI folder a bit
Lot of current UIControls seems to be just about programatical layout, while the ui you want for your game will be build in the editor so lot of the stuff is not needed(atleast not needed to be shown).
So some stuff ended up in Layout(useful for programattic UI)
some in Common(shared, mostly some panels and Label etc.)
and new Folder got created Game(with stuff that only makes sense created through Editor)
Also the UIControl popup now searches only for ContainerControls as anything else is basically useless.

# 3) Some parts of the UI were abstracted away
I think its pretty important to have strong basic elements without any unneeded fuzz so Image and Label was stripped to their bare meat. I also added a semi-step Colorable which describes any ContainerControl which can have color(+its useful for the button which works with it).
Any functionality this had to do was abstracted away to HighlightableLabel(in case of the MouseOverColor) and ClickableImage(in case of onClick) and yes I retained all the connections to any code that relied on it

# 4) UI Gamepad/Keyboard Navigation
This is pretty much necessity so I implemented a pretty robust system for handling UI navigation
You have a UIInputSystem on the same Actor as UICanvas and together with it you have UIInputModule(extensible class for multiple input apis(if something like Rewired arises for Flax(Which I hope it wont))) with default implementation in FlaxUIInputModule.
Then you have a ISelectable which is an Interface which makes anyone inheriting it able to be navigated to/from.
This Iselestable is then impleneted in Selectable which is default implementation with alot of juicy features(allowed Input directions, override onNavLeft onNavRight etc....)
image

# 5) GameButton
As the current normal Button was moved to Layout folder there is need for a Button. This one called GameButton(As Iam not sure about name collision etc.)
This button works they way it catches events(Select/Deselct/Submit MouseEnter/Exit/Up/Down) and performs these operations on Image or Colorable(currently Label and Image)
This button also has OnClick though the implemented GameEvent which can be subscribed to from both GUI and Scripting
image

 # TLDR
 Of course everything is documented so dive in and see


mafiesto4:
1) PR is too big. It should be split into separate: Game Events, Navigation, UI Cleanup, UI Styling
2) Navigation should be in-built into GUI without need to add any scripts (editor needs to use this feature too for property editor tab navigation)
3) Renaming control types simply breaks existing docs/examples/scenes/games (we need type redirecting support for it (already on a roadmap) but this has been postponed since we don't want to rename stuff 3 months after engine release, we want engine API to be stable now)
4) Thx for spotting files naming typo for `HorizontalPanelElement` + `VerticalPanelElement` pair
5) Have you tried using those Game Events with C++ scripts or Visual Scripts? I wonder if we could handle those too (probably need to write some cpp code) but it would be nice to have proper scripting impl. Not just C# API feature.
6) UI Styling you've proposed is nice but we need sth better for game devs to define UI style and easily use it (eg. json asset with Button/Label/TextBox/Custom Style class so it can be reused and easily tweaked for a whole game/section)
7) Overall nice direction!

honzapatCZ:
Splitted into mulitple PRs. Closing

PR #430, opened 2021-04-07
Alphapanel

honzapatCZ:
I added an alpha panel and only showing ContainerControls in the SetTypePopup


Source/Editor/Modules/SourceCodeEditing/CodeEditingModule.cs
```diff
@@ -178,7 +178,7 @@ public ISourceCodeEditor SelectedEditor
         /// <summary>
         /// The control types collection (for game UI).
         /// </summary>
-        public readonly CachedTypesCollection Controls = new CachedTypesCollection(64, new ScriptType(typeof(Control)), IsTypeValidScriptingTypeControls, HasAssemblyValidScriptingTypes);
+        public readonly CachedTypesCollection Controls = new CachedTypesCollection(64, new ScriptType(typeof(ContainerControl)), IsTypeValidScriptingTypeControls, HasAssemblyValidScriptingTypes);
```
stefnotch, line 181: While I don't have an opinion regarding the `AlphaPanel`, this should probably be reverted to show all `Control`s

mafiesto4:
Merged, thanks.

PR #432, opened 2021-04-07
UI Abstractions

honzapatCZ:
I think both Label and Image should be strong base for just doing their thing anything special can be in other inheited classes.
I also add the concept of Colorable, which is shared between Image and Label, so they have common color property.


stefnotch (commented):
The `ClickableImage` abstraction feels a bit odd coming from you. Wouldn't a button with a image as a child be more in line with what you were suggesting?

honzapatCZ:
> The `ClickableImage` abstraction feels a bit odd coming from you. Wouldn't a button with a image as a child be more in line with what you were suggesting?

Yes, it would be and that would be my preffered approach. But that would mean changing all of that and Iam not even sure @mafiesto4  approves this way, yes if this PR gets accepted I can make all stuff there use image as child or whatever.

PR #434, opened 2021-04-07
UI Auto Navigation

honzapatCZ:
dependant on: #432
and on #433
though if needed it could be stripped to its bare minimum, though the example GameButton would be lacking features.
 
This is pretty much necessity so I implemented a pretty robust system for handling UI navigation
You have a UIInputSystem on the same Actor as UICanvas and together with it you have UIInputModule(extensible class for multiple input apis(if something like Rewired arises for Flax(Which I hope it wont))) with default implementation in FlaxUIInputModule.
Then you have a ISelectable which is an Interface which makes anyone inheriting it able to be navigated to/from.
This ISelestable is then impleneted in Selectable which is default implementation with alot of juicy features(allowed Input directions, override onNavLeft onNavRight etc....)
I know some of this stuff will have to get moved to like own actors etc
image


jb-perrier:
As mafiesto said "Navigation should be in-built into GUI without need to add any scripts (editor needs to use this feature too for property editor tab navigation)".

It should be integrated into the current stack and be configurable via console settings or smth.

honzapatCZ:
> As mafiesto said "Navigation should be in-built into GUI without need to add any scripts (editor needs to use this feature too for property editor tab navigation)".
> 
> It should be integrated into the current stack and be configurable via console settings or smth.

Ye, I get, thats why Iam saying it might need to be moved to own actor or something like that. The main thing is there is the Interface, there is the Default implemented Control, there is autonavigagtion, thats the hard work done. If its a script or an extension to Canvas, thats like no difference.

honzapatCZ:
Moving into #574

Issue #405, opened 2021-03-30
PxController::getUpDirection & PxController::setUpDirection not exposed through CharacterController


EnokViking:
Like the title suggests, these two methods have been left out from the CharacterController implementation.
https://docs.nvidia.com/gameworks/content/gameworkslibrary/physx/guide/Manual/CharacterControllers.html#up-vector
They're declared in PxControllerDesc.h

jb-perrier:
I've exposed these funcs, but before making a PR it probably need some more work and checks. Using SetUpDirection can make the character to overlap nearby geometry, so we need to use the `overlap recovery module` in order to keep things clean.

EnokViking:
> I've exposed these funcs, but before making a PR it probably need some more work and checks. Using SetUpDirection can make the character to overlap nearby geometry, so we need to use the `overlap recovery module` in order to keep things clean.

Given how it's implemented it makes sense to just expose that as a setting I feel. A call to PxControllerManager::setOverlapRecovery applies to all CCT's in the simulation. A possible solution to enable similar behaviour on a per CCT basis is to include a simple overlap check and resolve it with the translation vector returned by PxGeometryQuery::computePenetration.

mafiesto4:
Added in #435

PR #435, opened 2021-04-07
Expose Px Set/GetUpDirection.

jb-perrier:
Expose Px Set/GetUpDirection, serialization supported.

### DIFF
--- a/Source/Engine/Physics/Colliders/CharacterController.cpp
+++ b/Source/Engine/Physics/Colliders/CharacterController.cpp
@@ -92,7 +92,7 @@ void CharacterController::SetUpDirection(const Vector3& up)
 
 Vector3 CharacterController::GetUpDirection() const
 {
-    return _controller ? P2C(_controller->getUpDirection()) : Vector3::Up;
+    return _controller ? P2C(_controller->getUpDirection()) : _upDirection;
 }
 
 void CharacterController::SetMinMoveDistance(float value)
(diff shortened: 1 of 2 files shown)


Source/Engine/Physics/Colliders/CharacterController.h
```diff
@@ -141,6 +142,16 @@ DECLARE_SCENE_OBJECT(CharacterController);
...
 
+    /// <summary>
+    /// Sets the character up vector.
+    /// </summary>
+    API_PROPERTY() void SetUpDirection(const Vector3& up);
+
+    /// <summary>
+    /// Gets the character up vector.
+    /// </summary>
+    API_PROPERTY() Vector3 GetUpDirection() const;
```
mafiesto4, line 153: Move getter up to be first (in order). And add attributes like: `Attributes="EditorOrder(240), DefaultValue(true), EditorDisplay(\"Character Controller\")"`

Source/Engine/Physics/Colliders/CharacterController.cpp
```diff
@@ -82,6 +83,18 @@ void CharacterController::SetStepOffset(float value)
...
+void CharacterController::SetUpDirection(const Vector3& up)
+{
+    if (_controller)
+        _controller->setUpDirection(C2P(up));
+    _upDirection = up;
+}
+
+Vector3 CharacterController::GetUpDirection() const
+{
+    return _controller ? P2C(_controller->getUpDirection()) : Vector3::Up;
```
mafiesto4, line 95: Return `_upDirection` if no controller created.

mafiesto4 (changes requested):

mafiesto4:
Closes #405

Issue #453, opened 2021-04-13
3D Audio Inverted (XAudio2, OpenAL)


TaylerMauk:
**Issue description:**
3D audio placement is inverted on x-axis for XAudio2 and z-axis for OpenAL

**Steps to reproduce:**

1. Add an AudioSource to the scene and supply a sound to play
2. Add an AudioListener to the active camera
3. On play, observe correct placement for 2 of 3 dimensions

**Minimal reproduction project:**
[DemoProject.zip](https://github.com/FlaxEngine/FlaxEngine/files/6301382/DemoProject.zip)

**Flax version:**
1.1

mafiesto4:
Fixed by #454

PR #454, opened 2021-04-13
Fix 3D audio placement for XAudio2 and OpenAL

TaylerMauk:
Corrected inverted x-axis for XAudio2
Corrected inverted z-axis for OpenAL

See issue #453


mafiesto4:
Thanks!

PR #456, opened 2021-04-13
Overhaul Icons for the Editor

W2Wizard:
# New Icons
This PR is aimed at one of goals listed on trello for new icons. All Icons are from Icons8 (which is already mentioned on the About Window). The Icons are almost 99% Material Filled, which was also listed on trello, with a few iOS Glyphs icons (Save icon,...).

**In an ideal scenario** they would be all SVGs or an a alternative solution would be using an icon font, however for that to work, signed distance fields fonts would have to be implemented first.

### New (Black)
image

### Old
image

### Files
**Contains both Color & Non-Color**
[Final.zip](https://github.com/FlaxEngine/FlaxEngine/files/6305789/Final.zip)

Atlas .flax asset (so you don't have to redo all the sprite regions)
[IconsAtlas.zip](https://github.com/FlaxEngine/FlaxEngine/files/6306136/IconsAtlas.zip)


W2Wizard:
Only thing thats left out rn is the Items for Xbox & PS4, i'm not sure how to test for those Icons really.

stefnotch:
That is super awesome! 🥳 

We might need to add a license file or something for the icons. Mind adding a link to where you got the icons from? 😄

W2Wizard:
All Icons were collected from [Icons8](https://icons8.com/license).

honzapatCZ:
Ye so notice in about section. Up to mafi.

mafiesto4:
Looks more interesting! What's the icons group name from that _icons8.com_?

W2Wizard:
> Looks more interesting! What's the icons group name from that _icons8.com_?
> 
> Also, I though about keeping them a little more in _Material Design_ with more thin edges/fills (more lightweight look).

Perhaps if a more leightweight design is desired, perhaps [these](https://icons8.com/icons/fluent-systems-regular) fit more?
As they are also similar to what also on trello: https://streamlineicons.com/

I mainly focused on `Material Filled`, specifically the rounded variant, here is an overview as to what icon is from where: 

image

mafiesto4:
Thanks for the feedback! Let's keep slim&slick design for editor icons (to not go too far from the current UI style).
For me, the good choice would be: https://icons8.com/icons/fluent-systems-regular
What do you think?

And of course, using mostly a single pack would make the Editor look better. Also, we might think about upgrading the resolution of the icons to reduce blurriness on high-dpi displays. There is plenty of space in atlas so icons could be bigger.

Here is the original PSD file with icons atlas: https://drive.google.com/file/d/1_dvyq3ryeXg39Rb863rNPwbFS6V1FBi9/view?usp=sharing

W2Wizard:
I'll try out the `Fluent-Systems-Regular` and increase the resolution to 64px (2x).
Will ofcourse test it and send the result in here :)

Also what is you personal prefered choice regarding the settings files ? Colorful or just plain black? It seems that most people prefer the colorful version.

honzapatCZ:
Could we get the file icons colored? It would make the Content browser more interesting.

W2Wizard:
## New Icons
image

## File
[IconAtlasNew.zip](https://github.com/FlaxEngine/FlaxEngine/files/6472719/IconAtlasNew.zip)

honzapatCZ:
Noice, how does JSON asset look like?

PR #458, opened 2021-04-15
Game Plugin Editor Deinitiliazitation order

honzapatCZ:
Given that the GamePlugins are initialized last, I think they should be deinitialized first

### DIFF
--- a/Source/Editor/States/PlayingState.cs
+++ b/Source/Editor/States/PlayingState.cs
@@ -158,6 +158,8 @@ namespace FlaxEditor.States
         public override void OnExit(State nextState)
         {
             IsPaused = true;
+            
+            PluginManager.DeinitializeGamePlugins();
 
             // Remove references to the scene objects
             Editor.Scene.ClearRefsToSceneObjects();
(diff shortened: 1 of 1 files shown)


mafiesto4:
I think it should be opposite: in cooked build game plugins are initialized before scenes load and got deinitialization after scenes unload so it should be the same in the editor to mock the gameplay.
Note that common case is to use game plugins yo implement global gameplay systems or tools or SDKs integration so those should be available before scenes opening.

honzapatCZ:
> I think it should be opposite: in cooked build game plugins are initialized before scenes load and got deinitialization after scenes unload so it should be the same in the editor to mock the gameplay.
> Note that common case is to use game plugins yo implement global gameplay systems or tools or SDKs integration so those should be available before scenes opening.

OK, well the best would be to have it before a scene is loaded for the gameplay.(eg subscribing to OnSceneLoaded would trigger even for the first one) but should be deinitilized before the reload for editor happens(otherwise the dirty hack with isPlaying is needed)

mafiesto4:
I fixed it here: 3e3b0e795156b2c6a01dcd1bc956a04ca90446c7 together with initialize fix so Game Plugins are not initialized after editor-scenes are unloaded but before playmode scenes are loaded. Deinitialization happens in the opposite moment.

PR #465, opened 2021-04-19
Misc math changes

W2Wizard:
Add a function to easily get the unit direction between two points as well as some stuff here and there.


stefnotch:
Do we really need a function that just does `(b-a).Normalized`? It took me longer to figure out what the function is supposed to do based on its name than it took me to figure out what the code does. xD

Might just be me, but I'm almost more used to seeing lots of vector primitives instead of tons of unique functions

mafiesto4:
Yeah I also doubt a little the necessity of this util method.

PR #467, opened 2021-04-20
Better General Actor Editor

honzapatCZ:
Are you tired of having so much wasted space just by general stuff you will most likely change only once? Would you maybe want something special for the general category?
Look no further Better General Actor Editor is here.

Back to the future: Implements icon, which could in future be used for custom gizmos and stuff like that....

PS: Tested on default layout and works without problem

image


stefnotch:
I'm guessing you didn't mean to commit the `VolumetricFog.flax`.

Anyways, for completeness sake, here is a link to the Discord discussions https://discord.com/channels/437989205315158016/509056735844106251/834025619012321310

honzapatCZ:
> I'm guessing you didn't mean to commit the `VolumetricFog.flax`.
> 
> Anyways, for completeness sake, here is a link to the Discord discussions https://discord.com/channels/437989205315158016/509056735844106251/834025619012321310

ye that's a mistake, idk how stuff like this happens? any idea how to remove it?

stefnotch:
`git add .` is the fastest way to cause stuff like this.
Regarding removing it, I personally have no idea. But StackOverflow has the answer https://stackoverflow.com/q/12481639

honzapatCZ:
> `git add .` is the fastest way to inadvertently add files like this.
> Regarding removing it, I personally have no idea. But StackOverflow has the answer https://stackoverflow.com/q/12481639

I meant how can change like this happen if I havent modified the file. But OK, thanks

honzapatCZ:
Btw mafi if you are not convinced by for example the logo, if @W2Wizard creates some nice like actor/prefab or this could be like the actors icon?(joint, sky, audio source etc) and alternatively a place to select custom gizmos I guess

Source/Editor/CustomEditors/Dedicated/ActorEditor.cs
```diff
@@ -83,31 +112,93 @@ public override void Initialize(LayoutElementsContainer layout)
...
-                    break;
-                }
-            }
+        private void BetterGeneralEditor(LayoutElementsContainer group)
+        {
+            HorizontalPanelElement hor = group.HorizontalPanel();
+            hor.Panel.Margin = Margin.Zero;
+            hor.Control.Height = TextBoxBase.DefaultHeight * 2 + 2 + 2 + 2;
+
+            ImageElement img = hor.Image(Editor.Instance.Icons.Logo128);
```
stefnotch, line 124: Okay, after thinking about this for a bit, I realized that for a proper actor-image/actor-icon implementation, we would need to be able to store editor-only metadata for every actor. I haven't looked too closely into this, but I think Flax currently doesn't have a way of doing that.

honzapatCZ: Hmmm, well cant actor have like a SpriteHandle #IF FLAX_EDITOR or something like that?

PR #478, opened 2021-04-24
vkCmdDraw*Indirect impl.

jb-perrier:
Implementation of DrawInstancedIndirect & DrawIndexedInstancedIndirect.


mafiesto4:
Do we need this `drawCount `? What can be a use case for that?

jb-perrier:
It can be useful if we want to store multiple meshes in one vb. Using this draw count, we could render all of these "contained" meshes in one cmd. So it save the time/overhead where we need to loop and execute one cmd for each mesh ( using offset ).
Currently the RenderList doesn't support doing that (one cmd for multiple meshes in same buffer) but i'm working on.

In case of the " one buffer for all meshes " trello card that would be useful.

mafiesto4:
Okay cool! But for `one vb for all lod meshes` we can simply submit multiple draw calls like we do it now but use the same `GPUBuffer` with different `vertexBuffersOffsets` which is already in DrawCall.

Relevant code:
https://github.com/FlaxEngine/FlaxEngine/blob/master/Source/Engine/Graphics/Models/Mesh.cpp#L372-L381

mafiesto4:
I did testing on new indirect drawing support for Vulkan and it works nicely. Although I found a few crashes/bugs and fixed those here: cf799c2198ad0088918d656818676c3a7c7a86e2 and cbea62ceb9150d6c57f75d9e1cf8e3bad95b0357.

This PR is quite important as now particles can be simulated on GPU but I found some problems on Particles Features Tour:
image

Note that sparks are not falling now. Need to debug it with RenderDoc to find out why it doesn't work...

jb-perrier:
> Okay cool! But for `one vb for all lod meshes` we can simply submit multiple draw calls like we do it now but use the same `GPUBuffer` with different `vertexBuffersOffsets` which is already in DrawCall.

The whole point is to send one cmd instead of `n` cmds ( using offset require one cmd per offset ). But since this feature is already integrated let's keep that intact. The reason why i want to add this draw count, is that devs can use this to optimize their rendering code ( in this case RenderList will not try to batch this draw call and just execute it ). In my case using an array of indirect cmd will transforms smth like ~500 cmds ( one per offset, i want to render 500 chunks from my vertex buffer ) into 1 cmd (array of 500 indirect args, one cmd). 

But at this point another issue come up, there is no way to modify how the Mesh class create/populate his own draw call, we should think about that.
Because VertexBufferOffset / Indirect cmd features are useless if we can't correctly setup the draw call ( in an other way than the default one ). We can inherits from Mesh and override but since Model::SetupLODs will instantiate Mesh class ( and not the custom one )  i don't really know how to handle that.

GPU Particles \0/
> Note that sparks are not falling now. Need to debug it with RenderDoc to find out why it doesn't work...
Maybe another feature is missing ? I know that `Append/ConsumeStructuredBuffers` are missing ( can't find any docs about that in vulkan ). Will try to figure out what's happening here as i'm working on this PR.

jb-perrier:
Finally i've reverted changes about indirect draw count. Not pertinent enough, offsets are fine.

jb-perrier:
Ready for merge.

Issue #449, opened 2021-04-11
[Proposal] VisibleIfAttribute should have counterpart HideIfAttribute instead of invert parameter


Muchaszewski:
**Proposal description:**
Instead of having bool "invert" in `VisibleIfAttribute` its counterpart should be `HideIfAttribute`. This would be more obvious from code editor building.

Consider the following code:
```
[VisibleIf(memberName, false)]
public string Foo...
```
vs
```
[HideIf(memberName)]
public string Foo...
```

`HideIfAttribute` would be more in OOP spirit than undescriptive parameter for `VisibleIfAttribute`.

`inverted` can be left for compatibility reasons as well as added to HideIf - if someone choose to use it

mafiesto4:
Would be nice to have. Also, you can use the member name of the property and use custom expression there with negation or any other more cplex checking.

honzapatCZ:
Could this be closed if ShowIf has Invert?
@mafiesto4

PR #479, opened 2021-04-25
Add HideIf attribute

TaylerMauk:
Here is some initial code for a HideIf attribute implementation. See issue #449 

This works essentially the same way as VisibleIf, so maybe the logic should be abstracted prior to being merged to reduce the amount of redundant code. Thoughts?


mafiesto4:
Yeah please merge code logic to make it reusable for `VisiableIf` and `HideIf`. Will be easier to maintain and extend in future.

mafiesto4:
Ah I forget that `VisibleIfAttribute` has `Invert` property to simply do `HideIf` xD

```cs
public class TestVisibleIf : Script
{
    public bool ShowIt;

    [VisibleIf(nameof(ShowIt), Invert = true)]
    public int Value;
}
```

PR #504, opened 2021-05-06
External Assembly Folder

honzapatCZ:
Unclutering the build folder yay


mafiesto4:
What's wrong in having all binaries inside root folder? Have you tested all public platforms with those changes? (Windows, Linux, Android, UWP) I worry that separate `Assembly` folder might cause issues regarding 3rd party native binaries usage in games (eg. Mono search path). We used to have C# DLLs in a separate folder in a past but it was causing only problems and current flow is more intuitive.

PR #505, opened 2021-05-06
Rename exe on Windows and Fix logic on Linux

honzapatCZ:
I've implmented exe renaming the same way it's done on Linux(I know we discussed this, but I truly couldnt find it)
Other thing to note, I think the logic was inverted, the if statement should check if the names aren't equal, as if they were it would be pointless to rename.


mafiesto4:
This breaks games using C++ scripting because native DLL references `FlaxGame.exe` when linking. It works fow Linux because we have separate FlaxEngine shared lib and executable file is just a wrapper to call engine.
I see 2 solutions:
- use separate `FlaxGame.exe` and `FlaxEngine.dll` on Windows so we can rename exe and use dll for linking
- apply renaming in `Flax.Build` - can be tricky to do

honzapatCZ:
Hmmm, I though linking happened before deploy and in deploy those are all the final touch ups... weird.

PR #508, opened 2021-05-06
Add physics queries for convex meshes

GoaLitiuM:


mafiesto4:
Looks good, except `CollisionData` should be passed as pointer to those methods as it's an asset object.

GoaLitiuM:
Fixed and added validity checks for collision data.

PR #512, opened 2021-05-10
Add digit check at asset renaming.

### LINKED ISSUE #502: Creating a script with a number in it results in an error.
Creating a script with a number in it results in an error. Might want to stop User's from being able to do this at all.
image

**Flax version:**
1.1

jb-perrier:
Fix #502 by checking file name when the user rename it.

### DIFF
--- a/Source/Editor/Content/Proxy/ContentProxy.cs
+++ b/Source/Editor/Content/Proxy/ContentProxy.cs
@@ -74,6 +74,16 @@ namespace FlaxEditor.Content
         }
 
         /// <summary>
+        /// Determines whether the specified filename is valid for this proxy.
+        /// </summary>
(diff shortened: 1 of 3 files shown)


jb-perrier:
At first, i had implemented a virtual function `IsValidName` for proxy, so we can choose how names are verified per asset type. But for now it's probably overkill.

mafiesto4:
Hmm yeah let's do it via asset proxy. Some game projects use digits for map names prefix (eg. chapter number) to have nice ordering so we should restrict this first digit only for scripts.

jb-perrier:
Implemented, unfortunatly i had to add another check for `NewItem` since `ContentDatabase.GetProxy` doesn't work with it ( not sure if it's expected or not ).

PR #513, opened 2021-05-10
Makes The C# Script Template Doc Friendly

LCRW:


LCRW:
Added Proxies for C++ Json Asset (ISerializable) and C++ Function Library (Static class) and corresponding templates.

mafiesto4:
Thanks!

I improved it further by some minor code style tweaks and using shared base for Cpp Proxies in 9195e37a239fba1e25f99cc81601d1070a457ac0.

Issue #497, opened 2021-04-30
Visual Script with Int16 Parameter causes issues


stefnotch:
**Issue description:**
When you add a parameter of the type `Int16` to a visual script and then attach the visual script to an actor, Flax will loudly complain

![image](https://user-images.githubusercontent.com/10220080/116739267-c1e3a200-a9f3-11eb-9287-e1a10a4fd311.png)

**Steps to reproduce:**
1. Create a visual script
2. Add a `Int16` parameter 
3. Save the visual script
4. Attach the visual script to an actor (in the editor)
5. As soon as you attach it, the output log will start spamming this in the output log

```
[Error] Invalid type to box Structure System.Int16
[ 00:00:13.653 ]: [Warning] Failed to setup values and UI for item IndexFace
[ 00:00:13.653 ]: [Warning] Object reference not set to an instance of an object.
[ 00:00:13.656 ]: [Warning]   at FlaxEditor.CustomEditors.Editors.ShortEditor.GetValue (System.Object value) [0x00000] in C:\Flax\FlaxEngine\Source\Editor\CustomEditors\Editors\IntegerEditor.cs:216 
  at FlaxEditor.CustomEditors.Editors.SignedIntegerValueEditor.Refresh () [0x0000e] in C:\Flax\FlaxEngine\Source\Editor\CustomEditors\Editors\IntegerEditor.cs:148 
  at FlaxEditor.CustomEditors.CustomEditor.Initialize (FlaxEditor.CustomEditors.CustomEditorPresenter presenter, FlaxEditor.CustomEditors.LayoutElementsContainer layout, FlaxEditor.CustomEditors.ValueContainer values) [0x0002f] in C:\Flax\FlaxEngine\Source\Editor\CustomEditors\CustomEditor.cs:125 
  at FlaxEditor.CustomEditors.LayoutElementsContain...
  ```

**Flax version:**
Flax 1.2

LCRW:
Its in master also, looks like a threading issue maybe., Haven't been able to lock down error in debug.

LCRW:
Awesome work

jb-perrier:
Fixed in #515 .

PR #515, opened 2021-05-10
Int16 & Uint16 Scripting/Serialization support.

jb-perrier:
#497 , Uint16 & Int16 were not supported for serialization & interop.

For now as a draft, i have to review myself on Variant::Cast.

### DIFF
--- a/Source/Editor/Utilities/VariantUtils.cs
+++ b/Source/Editor/Utilities/VariantUtils.cs
@@ -55,7 +55,10 @@ namespace FlaxEditor.Utilities
 
             Int2,
             Int3,
-            Int4
+            Int4,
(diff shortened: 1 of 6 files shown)


Source/Editor/Utilities/VariantUtils.cs
```diff
@@ -55,7 +55,10 @@ internal enum VariantType
 
             Int2,
             Int3,
-            Int4
+            Int4,
+            
+            Int16,
+            Uint16
```
mafiesto4, line 61: Add `,` so we don't need to edit this line later on.

jb-perrier:
Rdy for merge.

PR #520, opened 2021-05-13
OnAwake before all OnStarts

honzapatCZ:
As proposed in discord. This calls all OnAwakes before all OnStarts.
Scripts can now properly setup themselfs and after that, they can talk to each other fully initialized.

### DIFF
--- a/Source/Engine/Level/Actor.cpp
+++ b/Source/Engine/Level/Actor.cpp
@@ -299,6 +299,7 @@ void Actor::SetParent(Actor* value, bool worldPositionsStays, bool canBreakPrefa
         // Fire events
         PostSpawn();
         {
+            Setup();
             SceneBeginData beginData;
             BeginPlay(&beginData);
             beginData.OnDone();
@@ -759,15 +760,7 @@ void Actor::PostLoad()
     // Use lazy creation for the managed instance, just register the object
     if (!IsRegistered())
         RegisterObject();
-    
-    // Fire events for scripting
-    for (auto* script : Scripts)
-    {
-        CHECK_EXECUTE_IN_EDITOR
-        {
-            script->OnAwake();
-        }
-    }
+
 }
 
 void Actor::PostSpawn()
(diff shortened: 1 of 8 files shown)


mafiesto4:
Cool, just 2 cents from my side:
- what about `PostSpawn` - called when someone spawns script at runtime - I guess it needs to call `OnAwake` too
- we gonna need to test this heavily!

honzapatCZ:
> Cool, just 2 cents from my side:
> 
> * what about `PostSpawn` - called when someone spawns script at runtime - I guess it needs to call `OnAwake` too
> * we gonna need to test this heavily!

Oh, so that was the difference between load and spawn, ok so I should add it to the other too? Ok.

mafiesto4:
`OnAwake` is also called by:
- `PrefabInstanceData::SynchronizePrefabInstances` 
- `Script::SetParent`

Those usages should be updated too.

honzapatCZ:
> `OnAwake`

I think those usages are OK as they are now, this change only applies to stuff that is batch spawned(loading level, spawning prefab with children) etc.

honzapatCZ:
Unfortunatly looks like another recursive function is needed as there is a bit more stuff happening after PostSpawn/PostLoad so scripts wont be able to for example destroy themself. Also on similiar note, if scripts are scene objects and have the BeginPlay why isn't OnStart triggered there? so Actors would just recurse that to children and scripts instead of having to managing scripts stuff too?

mafiesto4:
Note to self: probably move OnStart (and first OnEnable) to be called after the whole scene setup when loading a level. So physics and all objects are initialized properly. (test it deeply)

mafiesto4:
Thanks for your contribution. I implemented this by refactoring `PostLoad`/`PostSpawn` events into a single `Initialize` method so all `OnAwake` events are called before any `OnStart`/`OnEnable` which resolves the problem.

Changes: 3239150a610d38bccb9a93f786371ed7227fefda

Docs update: https://github.com/FlaxEngine/FlaxDocs/commit/134c73cf70e6224f5796b8f80771719926e9a6ea

Issue #448, opened 2021-04-11
[Discussion] UI Coordinate Spaces


stefnotch:
**Feature request/discussion:**
https://discord.com/channels/437989205315158016/437989205315158018/830820806207602748 and above

1. Using pixels could be discouraged in favor of some alternatives
   - e.g. `% of parent`, `% of screen`, `font size`, `normalized size`
   - Reasoning for this is that pixel sizes and screen resolutions are all but consistent
   - We pretty much need fancy layouting algorithms to handle cases like "squares should stay squares"
   - Related: https://gamedev.stackexchange.com/questions/2159/what-coordinate-system-to-use-to-handle-2d-ui
![image](https://user-images.githubusercontent.com/10220080/114310528-445efd00-9aeb-11eb-8eef-68e7b137a961.png)

**Flax version:**
Flax 1.1

LCRW:
Whilst we brushed on this in discord, maybe some sort of brief could be developed here in order for this to work, and be fine-tuned. UI needs an arbitrary pixel space, we need an engine setting for a default resolution, maybe this could be the actual bounds of the pixel space to keep things simpler.

honzapatCZ:
OK you are completely missing the idea of current UI system. I have some PRs that make the UI system more editor friendly like unity. But the concepts are actually in Flax. What are the real problems? 
percentages? use anchors
keep square? now thats a bit of a problem, Iam sure there's some component, but eventually there should be some RootControl wide scaler, where you would specify some UI dimensions and model stuff around that.

LCRW:
@honzapatCZ  This is a discussion on a feature spec, and how it should be implemented, one way is to integrate it in to the existing ui, but that is not the only option here. The UI stack could continue without even knowing about this feature, if required.

honzapatCZ:
> @honzapatCZ This is a discussion on a feature spec, and how it should be implemented, one way is to integrate it in to the existing ui, but that is not the only option here. The UI stack could continue without even knowing about this feature, if required.

Yeyeye I know, we discussed yesterday on discord. I still dont get much of the benefit, but thats up to you.

LCRW:
Ok here is how Unity approaches the problem [https://docs.unity3d.com/Packages/com.unity.ugui@1.0/manual/HOWTO-UIMultiResolution.html#scaling-with-screen-size](url)

mafiesto4:
[...]

PR #525, opened 2021-05-18
Custom UI scaling

honzapatCZ:
I suppose everything is right, iam just missing the ability to override size, so maybe make it like a property?

### DIFF
--- a/Source/Engine/UI/GUI/CanvasRootControl.cs
+++ b/Source/Engine/UI/GUI/CanvasRootControl.cs
@@ -54,6 +54,38 @@ namespace FlaxEngine.GUI
             }
         }
 
+        /// <summary>
+        /// Min scale value, anything below will get scaled up to fid this size
+        /// </summary>
+        public Vector2 MinCustomScaler = Vector2.Zero;
+        /// <summary>
+        /// Max scale value, anything above will get scaled down to fit this size
+        /// </summary>
+        public Vector2 MaxCustomScaler = Vector2.Maximum;
+        /// <summary>
+        /// Helper value to determine the custom scale;
+        /// </summary>
+        public Vector2 CustomScaler
+        {
+            get
+            {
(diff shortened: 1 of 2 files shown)


Source/Engine/UI/GUI/CanvasRootControl.cs
```diff
@@ -54,6 +54,38 @@ public override Control FocusedControl
...
+        /// <summary>
+        /// Helper value to determine the custom scale;
+        /// </summary>
+        public Vector2 CustomScaler
+        {
+            get
+            {
+                Vector2 outVal = Vector2.One;
+                Vector2 size = base.Size;
+                Debug.Log(size);
```
stefnotch, line 74: There is a stray `Debug.Log(size)` here and another one a few lines below

honzapatCZ: Ye it's not done and doesn't work 100%

stefnotch:
Relevant issue https://github.com/FlaxEngine/FlaxEngine/issues/448

Source/Engine/UI/GUI/CanvasRootControl.cs
stefnotch: Ah, fair enough. You could mark this pull request as a draft then.

Haridev669 (approved):
This is a great improvement

mafiesto4:
Added **Canvas Scaler** control in https://github.com/FlaxEngine/FlaxEngine/commit/d09dfd1f613c007911d999ac398ba82bc228a5df which implements this feature but also adds more functionalities (constant UI scale, DPI-based scaling, etc.). It should be fairly enough to properly scale game UI. I wrote some docs about it too:: https://github.com/FlaxEngine/FlaxDocs/blob/1.5/manual/ui/controls/canvas-scaler.md

PR #527, opened 2021-05-20
Command line -play argument.

jb-perrier:
`-play` as command line argument enable us to directly jump into play mode. The editor will skip script compilation.

A specific scene can be loaded using the scene guid : `-play 316098fa4b0cc2ead5196dad2827b437`

If the id is invalid or not provided the editor will play the scene based on editor settings ( Default or Last opened ).

Looks like it generates many `[Warning] GPU Async Task UploadBuffer (Running) has been canceled before a sync`.

### DIFF
--- a/Source/Editor/Editor.cpp
+++ b/Source/Editor/Editor.cpp
@@ -519,6 +519,11 @@ bool Editor::Init()
         return true;
     }
 
+    if (CommandLine::Options.Game.HasValue())
+    {
+        CommandLine::Options.SkipCompile.SetValue(true);
+    }
+    
     // If during last lightmaps baking engine crashed we could try to restore the progress
     ShadowsOfMordor::Builder::Instance()->CheckIfRestoreState();
 
@@ -534,6 +539,11 @@ bool Editor::Init()
     // Initialize managed editor
     Managed->Init();
 
+    if (CommandLine::Options.Game.HasValue())
+    {
+        Managed->RequestStartPlay();
+    }
+    
     return false;
 }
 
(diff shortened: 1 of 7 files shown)


jb-perrier:
I forgot to push some tweaks last day, draft until tonight.

Source/Engine/Graphics/Models/Mesh.cpp
```diff
@@ -403,6 +404,9 @@ void Mesh::Draw(const RenderContext& renderContext, MaterialBase* material, cons
 
 void Mesh::Draw(const RenderContext& renderContext, const DrawInfo& info, float lodDitherFactor) const
 {
+    if (!IsInitialized())
```
mafiesto4, line 407: It already checks for initialized state a few lines below.

mafiesto4:
Okay let's make it `-play` so it makes more sense. Good suggestion.

Source/Editor/Editor.cpp
```diff
@@ -519,6 +519,12 @@ bool Editor::Init()
         return true;
     }
 
+    // Skip compilation if play on start
+    if (CommandLine::Options.Game.HasValue())
```
mafiesto4, line 523: Does it work without compilation skipping? Because devs can already use `-skipcompile` arg (added by default to C++ projects debug cmd line) or disable initial compilation in Editor options.

jb-perrier: Ah yes ofc !

Source/Engine/Graphics/Models/Mesh.cpp
jb-perrier: My bad, how did I miss that.

PR #529, opened 2021-05-24
Only build bindings for project references with C#-only projects

GoaLitiuM:
This should skip the C++ #include parsing step of C++ references while building projects with only C#-script files.

In my tests, this change reduces the script compilation times from ~1.5 seconds down to ~0.8 seconds.


mafiesto4:
Might be a good idea. Have you tested game cooking? Does it work fine with C+ projects too?

GoaLitiuM:
Building a C++ project seems to work just fine, same with cooking the same game.

PR #532, opened 2021-05-25
Add how-to context for building Windows + Vulkan.

herocrab:
Small add to readme to clarify how to add Vulkan to Windows build.


stefnotch:
Maybe we should add that to the FlaxEngine documentation https://docs.flaxengine.com/manual/platforms/index.html

PR #550, opened 2021-06-13
Fix issues caused by the new Icons.

W2Wizard:
# Icon patch
With the new icons added by #456 some of the icons were either too small or had to little detail to be displayed properly.
This PR aims to fix some of the issues that came from that PR.

## Fixes
- [x] The buttons for adding new Functions and overrides are too small.
- [x] Timeline track icons are too small.
- [x] Debug log icons are unrecognizable.
- [x] Debug log tab icon is too small.
- [x] Visject Impulse pin has to be made a bit bigger
- [x] Add LocalizationSettings, SwitchSettings and Json Asset Icons.

## File

[IconsAtlas.zip](https://github.com/FlaxEngine/FlaxEngine/files/6644600/IconsAtlas.zip)

### Buttons

Getting rid of those tiny icons and using buttons instead makes it much more visible and obvious where functions and overrides are located.

https://user-images.githubusercontent.com/63303990/121814887-b252ac80-cc73-11eb-984f-7bed4e877fb2.mp4

### Timeline

The icons used on the Timeline track lacked pixels to be displayed clearly, now its very easy to see them and interact with them.

image

### Debug log

The icons used for the debug log window also had too little detail and were almost seethrough, not anymore.

image

### Misc

Made the Icons for the game cooker a bit bigger as well to in order to keep them symmetric with the buttons.

image

The new localization settings, json Asset & switch settings icons

image

image

### DIFF
--- a/Source/Editor/Content/Items/JsonAssetItem.cs
+++ b/Source/Editor/Content/Items/JsonAssetItem.cs
@@ -22,7 +22,7 @@ namespace FlaxEditor.Content
         public JsonAssetItem(string path, Guid id, string typeName)
         : base(path, typeName, ref id)
         {
-            _thumbnail = Editor.Instance.Icons.Document128;
+            _thumbnail = Editor.Instance.Icons.Json128;
         }
 
         /// <summary>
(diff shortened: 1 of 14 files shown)


W2Wizard:
The tab icon is slightly more visible now compared to before.
image

PR #566, opened 2021-07-09
Prevent file dialogs from changing the working directory

GoaLitiuM:
Fixes the issue whenever something was imported to the project, script compilation stops working due to missing files. This is caused by the sudden change of process working directory which the Windows file dialogs does by default and the build tool inherits the current working directory from the editor leading to this issue. This behavior can be easily disabled with the `OFN_NOCHANGEDIR` flag when constructing the dialog.


jb-perrier (approved):

jb-perrier:
I'm able to reproduce the issue, the fix works great.

PR #568, opened 2021-07-16
Rectangle distance

honzapatCZ:
Will be usefull for my UI nav

### DIFF
--- a/Source/Engine/Core/Math/Rectangle.cs
+++ b/Source/Engine/Core/Math/Rectangle.cs
@@ -284,6 +284,17 @@ namespace FlaxEngine
             Vector2 toExpand = Size * (scale - 1.0f) * 0.5f;
             return new Rectangle(Location - toExpand * 0.5f, Size + toExpand);
         }
+        
+        /// <summary>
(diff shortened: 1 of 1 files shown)


stefnotch:
Woah, that's a cool implementation! I wasn't aware of this mathematical trick previously.

For others who are confused, this is what the code does:
1. It calculates the distance between the centers 
2. Then, it subtracts the parts that are already covered by the rectangles

image
-- That stuff, but just along the X-axis. The code does it for both the X and Y axis.

honzapatCZ:
> Woah, that's a cool implementation! I wasn't aware of this mathematical trick previously.
> 
> For others who are confused, this is what the code does:
> 
> 1. It calculates the distance between the centers
> 2. Then, it subtracts the parts that are already covered by the rectangles
> 
> image
> -- That stuff, but just along the X-axis. The code does it for both the X and Y axis.

Ye I was thinking how to do it and I was only able to figure out the way with like if left top then xy, but I was sure there is some mathematical way, one stack overflow search later here it is. Also looking at that I simplified the multiplication and fixed the - to +.

PR #569, opened 2021-07-17
Improve detection of Rider installations

GoaLitiuM:
Installing Rider with JetBrains Toolbox does not update the paths in registry, so these needs to be manually detected from the default Toolbox installation directory.

This PR also adds support for detecting Rider for Unreal Engine installations.


jb-perrier (approved):

jb-perrier:
Works great ! Tested with Rider for Unreal.

PR #578, opened 2021-08-05
Order files/folders by name

thallard:
Now items can be sorted by their names.
I've tried to add the "Last Modified" option but I didn't find the way to get this parameter. :/

Here a little peak! :)

TEST1

### DIFF
--- a/Source/Editor/Windows/ContentWindow.cs
+++ b/Source/Editor/Windows/ContentWindow.cs
@@ -39,7 +39,7 @@ namespace FlaxEditor.Windows
         private TextBox _foldersSearchBox;
         private TextBox _itemsSearchBox;
         private ViewDropdown _viewDropdown;
-        private string _sortBy;
+        private string _sortBy = "Alphabetic Order";
 
         private RootContentTreeNode _root;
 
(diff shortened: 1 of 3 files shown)


jb-perrier:
I haven't found anything about a stored date per asset. A trick would be to use `File.GetLastWriteTime(item.Path);`.
But it's probably better to extands the api than relying on `System.IO.File`.

Source/Editor/Windows/ContentWindow.cs
```diff
@@ -230,6 +246,19 @@ private void OnFilterClicked(ContextMenuButton filterButton)
             _viewDropdown.OnClicked(i);
         }
 
+        private void OnSortByButtonClicked(ContextMenuButton button)
+        {
+            _buttonNameChecked = button.Text;
+            switch (button.Text)
```
mafiesto4, line 252: This could use `(SortType)button.Tag` instead of button text checking.

Source/Editor/Windows/ContentWindow.cs
```diff
@@ -215,6 +217,20 @@ private ContextMenu OnViewDropdownPopupCreate(ComboBox comboBox)
...
+            sortBy.ContextMenu.AddButton("Alphabetic Order", OnSortByButtonClicked).Tag = SortType.AlphabeticOrder;
+            sortBy.ContextMenu.AddButton("Alphabetic Reverse", OnSortByButtonClicked).Tag = SortType.AlphabeticReverse;
+            sortBy.ContextMenu.VisibleChanged += control =>
+            {
+                if (!control.Visible)
+                    return;
+                foreach (var item in ((ContextMenu)control).Items)
+                {
+                    if (item is ContextMenuButton button)
+                        button.Checked = _buttonNameChecked == button.Text;
```
mafiesto4, line 231: This could use `(SortType)button.Tag == _sortType` instead of comparison based on button text.

PR #583, opened 2021-08-07
Removed unnecessary NULL terminator in RunBuildTool() to fix #402

### LINKED ISSUE #402: Copying new log after clearing partially fails
**Issue description:**
When you clear the output log, then recompile the scripts and then try to copy the entire new log, only the first two lines get copied (or maybe the first 255 characters)

**Steps to reproduce:**
1. Open output log
2. Right click, clear
3. Recompile scripts
4. Select all 
5. Copy
6. Paste somewhere else

**Flax version:**
Flax 1.1

Zbyl:
Removed unnecessary NULL terminator in RunBuildTool() to fix #402: "Copying new log after clearing partially fails"


Zbyl:
Actually this redundant NULL was hiding an issue with StringView::GetText(): it can return non-null terminated string, which causes problem for example in WindowsPlatform::RunProcess().
I have created #584 for this problem.

PR #591, opened 2021-08-08
Add new (experimental) update mode

Erdroy:
This PR adds a totally new option for TimeSettings, where you can choose to force the draw to be perfectly synced with update.

Thanks to this mode, we can limit the FPS at which update and draw operates (without VSync) and make sure that we get draw called always at the same frame as update.

Two things:
* We can mark this feature as experimental, as it changes pretty much a lot (?).
* Also, we need to discuss wheter we want the paired mode as default - or not (?). 
I'd suggest to go for it, and make it default, so the update/draw is no longer confusing for new developers.

Draft until fully tested (draw DT/time/frames need some manual checks).

### DIFF
--- a/Source/Engine/Core/Config/TimeSettings.cs
+++ b/Source/Engine/Core/Config/TimeSettings.cs
@@ -1,0 +1,12 @@
+// Copyright (c) 2012-2021 Wojciech Figat. All rights reserved.
+
+namespace FlaxEditor.Content.Settings
+{
+    public partial class TimeSettings
+    {
+#if FLAX_EDITOR
+        // ReSharper disable once UnusedMember.Local
+        internal bool IsDrawFPSUsed => UpdateMode == TimeUpdateMode.UpdateAndDrawSeparated;
+#endif
+    }
+}
--- a/Source/Engine/Core/Config/TimeSettings.h
+++ b/Source/Engine/Core/Config/TimeSettings.h
@@ -5,12 +5,36 @@
 #include "Engine/Core/Config/Settings.h"
 
 /// <summary>
+/// Contains available time update modes for update and draw calls.
+/// </summary>
+API_ENUM(Namespace="FlaxEditor.Content.Settings") enum class FLAXENGINE_API TimeUpdateMode
+{
+    /// <summary>
+    /// Update and draw run separately - i.e. can run at different FPS, might get called at the same frame but this is not guaranteed.
+    /// Not recommended, might cause visible stutter and other graphical glitches when update is running less often than draw.
+    /// </summary>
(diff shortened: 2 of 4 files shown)


mafiesto4:
Okay so in terms of delta time stability for the Update using paired mode looks the same as using current separated mode with both update and draw at 60 FPS:
image

GoaLitiuM:
This affects editor usability more than stability, the effect is the same but there is only one value that needs to be changed in the editor when paired mode is enabled. In most typical scenarios, separated framerates are not needed so I think having this enabled by default makes sense.

Erdroy:
> This affects editor usability more than stability, the effect is the same but there is only one value that needs to be changed in the editor when paired mode is enabled. In most typical scenarios, separated framerates are not needed so I think having this enabled by default makes sense.

What value (that has to be changed) you are talking about?

GoaLitiuM:
This seems to cause both Update and Draw ticks to happen twice as often as specified in the combined `UpdateFps`. Following patch should fix it:
```patch
diff --git a/Source/Engine/Engine/Time.cpp b/Source/Engine/Engine/Time.cpp
index 1943418967a8742f7d8364a54a909788c361da48..ae4a6244c321e038089c1dd671acccc1577e03ad 100644
--- a/Source/Engine/Engine/Time.cpp
+++ b/Source/Engine/Engine/Time.cpp
@@ -106,7 +106,9 @@ bool Time::TickData::OnTickBegin(float targetFps, float maxDeltaTime, bool force
             NextBegin = time;
         }
 
-        if (targetFps > ZeroTolerance)
+        if (forceTick)
+            NextBegin = time + (1.0 / targetFps);
+        else if (targetFps > ZeroTolerance)
             NextBegin += (1.0 / targetFps);
     }
```

Zode:
Any word on this?
I would greatly appreciate if this was PR'd because I need to be able to do interpolation. Flax's update event is not per-frame like in Unity, which ends up breaking things :(

Zode:
@mafiesto4 maybe for 1.11? :pray: being able to do custom interpolation would be huge for KCC + CineBlend interoperability without the "frame behind" issue that comes with doing it in Update/LateUpdate (in Unity these are both synch'd to the frame so its not an issue there)

GasimoCodes:
@Zode told me to comment so here I am! We're pretty much trying to move objects with the camera but manually so its not tied in the hierarchy (with some smoothing like a spring arm). Right now it executes seemingly in random-ish intervals between either before or slightly after the transforms are updated so there is major noticeable jitter. This applies to Cineblend as well whenever anything physics is involved or you want to track transforms yourself.

PR #592, opened 2021-08-08
Fixed invalid uses of StringView::GetText(), where a null-terminated string was expected.

Zbyl:
Changes:
- Fixed many invalid uses of `StringView::GetText()`, where a null-terminated string was expected.  
  This should fix some hard to track bugs and random crashes.
- Renamed `GetText()` to `GetNonTerminatedText()` to reduce chance of identical bugs appearing in the future.
- Clarified some function comments to say where `nullptr`s are allowed, and where they are not.  
  This is important, as various functions have different conventions, and it's easy to make invalid assumptions.

Things to consider:
- I am was able to build the code for Windows and Linux platforms.  
  I might have missed some StringView::GetText() uses in other platforms. This would break compilation of those platforms. Help with that would be welcome.
- There are still quite a few places where `nullptr` is passed to function that explicitly defines it as an undefined behavior.  
  I have not touched those yet..

### DIFF
--- a/Source/Engine/Platform/Windows/WindowsClipboard.cpp
+++ b/Source/Engine/Platform/Windows/WindowsClipboard.cpp
@@ -25,7 +25,7 @@ void WindowsClipboard::SetText(const StringView& text)
 {
     const int32 size = (text.Length() + 1) * sizeof(Char);
     const HGLOBAL hMem = GlobalAlloc(GMEM_MOVEABLE, size);
-    Platform::MemoryCopy(GlobalLock(hMem), text.GetText(), size);
+    Platform::MemoryCopy(GlobalLock(hMem), String(text).GetText(), size);
     GlobalUnlock(hMem);
 
     OpenClipboard(nullptr);
--- a/Source/Engine/Platform/Windows/WindowsClipboard.cpp
+++ b/Source/Engine/Platform/Windows/WindowsClipboard.cpp
@@ -23,9 +23,12 @@ void WindowsClipboard::Clear()
 
 void WindowsClipboard::SetText(const StringView& text)
 {
-    const int32 size = (text.Length() + 1) * sizeof(Char);
-    const HGLOBAL hMem = GlobalAlloc(GMEM_MOVEABLE, size);
-    Platform::MemoryCopy(GlobalLock(hMem), String(text).GetText(), size);
+    const int32 sizeWithoutNull = text.Length() * sizeof(Char);
+    const HGLOBAL hMem = GlobalAlloc(GMEM_MOVEABLE, sizeWithoutNull + sizeof(Char));
+
+    Char* pMem = static_cast<Char*>(GlobalLock(hMem));
+    Platform::MemoryCopy(pMem, text.GetNonTerminatedText(), sizeWithoutNull);
+    Platform::MemorySet(pMem + text.Length(), sizeof(Char), 0);
     GlobalUnlock(hMem);
 
     OpenClipboard(nullptr);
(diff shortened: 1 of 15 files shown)


Source/Engine/Platform/Windows/WindowsClipboard.cpp
```diff
@@ -25,7 +25,7 @@ void WindowsClipboard::SetText(const StringView& text)
 {
     const int32 size = (text.Length() + 1) * sizeof(Char);
     const HGLOBAL hMem = GlobalAlloc(GMEM_MOVEABLE, size);
-    Platform::MemoryCopy(GlobalLock(hMem), text.GetText(), size);
+    Platform::MemoryCopy(GlobalLock(hMem), String(text).GetText(), size);
```
mafiesto4, line 28: If text to copy is large then allocating it into memory might be slower. Maybe we could do separate copy for text and manually insert null-terminator in the end?

Source/Engine/Platform/Windows/WindowsPlatform.cpp
```diff
@@ -1109,7 +1111,7 @@ int32 WindowsPlatform::RunProcess(const StringView& cmdLine, const StringView& w
 
     // Create the process
     PROCESS_INFORMATION procInfo;
-    if (!CreateProcessW(nullptr, const_cast<LPWSTR>(cmdLine.GetText()), nullptr, nullptr, TRUE, dwCreationFlags, (LPVOID)environmentStr, workingDir.HasChars() ? workingDir.Get() : nullptr, &startupInfoEx.StartupInfo, &procInfo))
+    if (!CreateProcessW(nullptr, const_cast<LPWSTR>(String(cmdLine).GetText()), nullptr, nullptr, TRUE, dwCreationFlags, (LPVOID)environmentStr, String(workingDir).GetText(), &startupInfoEx.StartupInfo, &procInfo))
```
mafiesto4, line 1114: For `workingDir` we should pass `nullptr` if it's empty so it picks the current process workspace.

Source/Tools/Flax.Build/Bindings/BindingsGenerator.Cpp.cs
```diff
@@ -166,7 +166,7 @@ public static string GenerateCppWrapperVariantToNative(BuildData buildData, Type
             if (typeInfo.Type == "String")
                 return $"(StringView){value}";
             if (typeInfo.IsPtr && typeInfo.IsConst && typeInfo.Type == "Char")
-                return $"((StringView){value}).GetText()";
+                return $"((StringView){value}).GetNonTerminatedText()"; // This is a bug, as we need a null-terminated strig here. Any idea how to fix it?
```
mafiesto4, line 169: I think it's not a problem as `Variant` if it's String type it contains null-terminator character so it's safe to access it. It will be null if string is empty.

Source/Engine/Platform/Windows/WindowsPlatform.cpp
Zbyl: Fixed in newest commit.

Source/Tools/Flax.Build/Bindings/BindingsGenerator.Cpp.cs
Zbyl: Added comments in Variant.h to document this property.

Source/Engine/Platform/Windows/WindowsClipboard.cpp
Zbyl: Improved as suggested.

mafiesto4 (approved):

Issue #75, opened 2020-12-29
Quaternion.LookAt(...) Reverse axis


Mummrik:
**Issue description:**
Rotation of my actor got reversed on X axis but working fine on the Z axis.

**Steps to reproduce:**
```
var position = _Some target position_
Actor.Rotation = Matrix.RotationQuaternion(Quaternion.LookAt(Actor.Position, position, Vector3.Up));
```

**Flax version:**
Flax engine 1.0.6214

stefnotch:
For completeness, this reproduces the same issue. 
`Actor.Orientation = Quaternion.LookAt(Actor.Position, position, Vector3.Up);`

A possibly related issue from another library is the following: https://github.com/toji/gl-matrix/issues/225

stefnotch:
So, the existing `Matrix.LookAt` (and `Quaternion.LookAt`) functions work for creating *view* matrices for a camera.
However, they don't work for creating *rotation* matrices for actors. (see https://github.com/toji/gl-matrix/issues/225 )

So, we'll need a new method and potentially rename the existing one. Any suggestions in that regard?

Mummrik:
What about ActorLookAt(...)?

I did just find out there is a Actor.LookAt(vec3), in my case i might just using the wrong function.

mafiesto4:
Actor has [Direction](https://docs.flaxengine.com/api/FlaxEngine.Actor.html#FlaxEngine_Actor_Direction) property for easier usage.

stefnotch:
Hm, there's LookRotation now. I wonder if that does what one expects when working with actors.

https://github.com/FlaxEngine/FlaxEngine/blob/fd3f10864b8c09289971d8cb2cfd64090d0cd026/Source/Engine/Core/Math/Quaternion.cs#L1023C28-L1023C40

PR #599, opened 2021-08-15
Add LookingAt to Actor

ScottLongley:
LookAt() modifies the Actor's rotation without returning anything and I needed a way to lerp that rotation, so LookingAt() just returns the rotation without modifying it.

### DIFF
--- a/Source/Engine/Level/Actor.cpp
+++ b/Source/Engine/Level/Actor.cpp
@@ -1425,9 +1425,23 @@ Actor* Actor::Intersects(const Ray& ray, float& distance, Vector3& normal)
 
 void Actor::LookAt(const Vector3& worldPos)
 {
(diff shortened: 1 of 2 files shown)


stefnotch:
Would the alternative of changing `LookAt` to return the orientation also work?

(Also, this is slightly relevant whenever talking about the lookat functions: https://github.com/FlaxEngine/FlaxEngine/issues/75 )

ScottLongley:
No, as someone may want to use that rotation data without affecting the Actor's orientation.

Consider this example that makes use of LookingAt:
Actor.Orientation = Quaternion.Lerp(Actor.Orientation, Actor.LookingAt(TargetPosition), Time.DeltaTime);

The only way to achieve this currently is to copy an Actor's orientation, rotate it with LookAt, copy the new orientation, rotate it back to it's original orientation, and then lerp between the two.

PR #601, opened 2021-08-15
Fixed bugs in string functions. Added tests for modified functions.

Zbyl:
Bugs fixed:
- `StringView::Compare()` did not perform lexicographical comparison.  
   It was also inconsistent with `String::Compare()` (which  worked correctly).  
   Fix incorrect documentation that stated that only -1, 0, or 1 can be returned.
- `String::Replace(Char)` incorreclty handled `StringSearchCase`.

Fixes of function semantics:
- `StartsWith()` and `EndsWith()` now return `true` when given empty prefix/suffix.  
  This is consistent with function description and also with C++ 20, C# and Python equivalents.  
  (Uses of those functions were reviewed, and no braking changes were found.)
- `String::Replace()` now ASSERTs when `searchText` is empty, as:
   - Searching for an empty string is not well defined.
   - It will always be a sign of a bug in the calling code.
   - Asserting in that case is consistent with behaviour of C# equivalent (C++ does not have any equivalent).  
    (Python equivalent in that case inserts replacement text between all characters, which is neither useful nor expected.)
- `String::Replace()` return value was inconsistent.  
   It could not be described as either of:
   - Number of occurences of searchText within the string.
   - Number of occurences of searchText that were replaced by replacementText, which resulted in actual change.

   The first definition is more useful and expected.
   The fix makes `Replace()` function slower in degenerate case when `searchText == replacementText` (for case sensitive search only).
   (Uses of `Replace()` were reviewed, and this degenerate case is never depended on.)

Additions:
- Added Catch2, header only test framework for native tests.  
  It is one of the most popular test frameworks for C++, and one that is very simple to use.
- Added tests for modified string functions.  
  (Code before fixes is failing on 21 out of 164 assertions.)
- Added LinkEnvironment.LinkAsConsoleProgram flag that allows for setting Windows subsytem to CONSOLE.


mafiesto4:
Thanks! Pretty nice unit testing 🎉

I did some renaming/moving of tests module. Left `Tests.Development` configuration only + disabled it's deployment when packaging Flax.

PR #608, opened 2021-08-20
Add margin to TilesPanel tile

ScottLongley:
This allows tiles inside a TilesPanel to be separated from each other using a Margin. The top margin of the first row and the left margin of the first column are ignored as I didn't want to inadvertently cause an issue, but I believe it solves the basic need of separated tiles.

Also contains a spelling fix inside GridPanel.


jb-perrier:
We have PanelWithMargins control for that, you inherits from it.

Source/Engine/UI/GUI/Panels/TilesPanel.cs
```diff
@@ -83,15 +111,15 @@ protected override void PerformLayoutBeforeChildren()
 
                 c.Bounds = new Rectangle(x, y, itemsWidth, itemsHeight);
```
mafiesto4, line 112: Shouldn't the x,y for Bounds rectangle be offset by Time Margin Left,Right?

ScottLongley: You're right, just pushed another commit with that line fixed.

PR #623, opened 2021-09-05
Fix engine ticking to catch up faster when falling behind

GoaLitiuM:
This addresses an issue when the engine is running slowly for a while, the framerate does not immediately get capped to the current effective framerate cap. For example, when the Flax Editor is not focused, the framerate is capped to 15fps, but when focus gets restored, framerate remains unlocked for a while before it gets capped again.


GoaLitiuM:
I forgot to add time skipping/frame dropping to physics ticking. This should also fix another weird case where high PhysicsFPS and unstable framerate would cause the physics to speed up when framerate becomes stable again.

PR #631, opened 2021-09-23
Fix getting toolchain

MiheevN:
To compile on Linux, the Toolchain was lost, it gave an error here:
https://github.com/FlaxEngine/FlaxEngine/blob/76a4d0652f2c64bc61cbf25ecfc713feafd7cdfb/Source/Tools/Flax.Build/Build/EngineTarget.cs#L163

Initially, null is passed - https://github.com/FlaxEngine/FlaxEngine/blob/76a4d0652f2c64bc61cbf25ecfc713feafd7cdfb/Source/Tools/Flax.Build/Build/NativeCpp/Builder.NativeCpp.cs#L957
and then the value is expected anyway.

It seems that the current fix brings the compilation back to normal.

Sorry for the formatting fix.

### DIFF
--- a/Source/Tools/Flax.Build/Build/NativeCpp/Builder.NativeCpp.cs
+++ b/Source/Tools/Flax.Build/Build/NativeCpp/Builder.NativeCpp.cs
@@ -251,22 +251,22 @@ namespace Flax.Build
             var modules = new List<Module>();
             switch (target.LinkType)
             {
-            case TargetLinkType.Monolithic:
-                // Include all modules
-                modules.AddRange(buildModules.Keys);
-                break;
-            case TargetLinkType.Modular:
-            {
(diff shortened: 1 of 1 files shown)


mafiesto4:
In some cases `platform.GetToolchain()` will crash in `BuildTargetNativeCppBindingsOnly` because `platform.HasRequiredSDKsInstalled` could `false` for some devs. We've got this issue reported few weeks ago and it was here: c4102ba8841cc9ddaf4364db7a091041eef11df7.

To overcome both problems I added `platform.TryGetToolchain()` here: e6ba2402b715db660117182149828542af269a70.
Does it work fine with this fix?

MiheevN:
Your code fix looks similarly functional.
But I don't have the strength to check yet.🥱

PR #663, opened 2021-10-30
Added NumThreadPoolThreads to Threading.h

intolerantape:
The new function exposes to script the number of worker threads available to the thread pool.

The number of threads in the thread pool was previously set in ThreadPool.cpp on line 54, within the function ThreadPoolService::Init(). I have promoted this value to a public function in the global namespace and exposed said function to the scripting API. The declaration of the new function is in Threading.h (immediately beneath IsInMainThread()) and its definition is in ThreadPool.cpp (also immediately beneath IsInMainThread()).

I wanted to know how many worker threads I was working with and couldn't find this value publicly exposed anywhere, so I added it. If the value is already publicly exposed and I missed it, please reject and disregard this PR.

### DIFF
--- a/Source/Engine/Threading/ThreadPool.cpp
+++ b/Source/Engine/Threading/ThreadPool.cpp
@@ -18,6 +18,11 @@ FLAXENGINE_API bool IsInMainThread()
     return Globals::MainThreadID == Platform::GetCurrentThreadID();
 }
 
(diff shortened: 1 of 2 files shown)


intolerantape:
PS Maybe Mafiesto can think of a better name? NumTaskThreads? ThreadPoolSize?

mafiesto4:
I wonder what might be an example use case of exposing this property?

PR #673, opened 2021-12-01
Multiple physics scenes

iainmckay:
Hi guys, I'm prototyping a game idea but needed to be able to create multiple physics scenes.

The goal here is to support multiple overlapping physical scenes for a game mechanic needed for my prototype. There are other uses such as forming part of a solution for making fast moving objects with complex contents such as spaceships.

This PR is still very early and not complete but I wanted to sanity check what I am doing and to ask if this is a feature that might be considered for inclusion in core or if I'm looking at maintaining my own fork (this I'm happy to do).

The changes primarily consist of moving the static functions of `Physics` to a new `PhysicsScene` class. I've kept the static functions but these simply wrap calls to the default scene for backwards compatibility.

Still to do:
- ~~When spawning or changing the parent of an actor, inherit the physics scene.~~
- ~~Support all physical based objects other than RigidBody and Collider~~

This is an example project using this feature. The scripts of interest are `CreatePhysicsScene.cs`, `MigratePhysicsActor.cs` and `SetPhysicsScene.cs`: [PhysicsWorld.zip](https://github.com/FlaxEngine/FlaxEngine/files/7635314/PhysicsWorld.zip)

### DIFF
--- a/Source/Engine/Level/Actor.cpp
+++ b/Source/Engine/Level/Actor.cpp
@@ -1820,11 +1820,13 @@ void Actor::SetPhysicsScene(PhysicsScene* scene)
     _physicsScene = scene;
 
     if (previous != _physicsScene)
+    {
         OnPhysicsSceneChanged(previous);
 
-    // cascade
-    for (auto child : Children)
-        child->SetPhysicsScene(scene);
+        // cascade
+        for (auto child : Children)
+            child->SetPhysicsScene(scene);
+    }
 }
 
 PhysicsScene* Actor::GetPhysicsScene()
--- a/Source/Engine/Level/Actor.cpp
+++ b/Source/Engine/Level/Actor.cpp
@@ -17,6 +17,7 @@
 #include "Engine/Debug/Exceptions/JsonParseException.h"
 #include "Engine/Graphics/RenderTask.h"
 #include "Engine/Graphics/RenderView.h"
+#include "Engine/Physics/Physics.h"
 #include "Engine/Profiler/ProfilerCPU.h"
 #include "Engine/Scripting/Scripting.h"
 #include "Engine/Serialization/ISerializeModifier.h"
@@ -77,6 +78,7 @@ Actor::Actor(const SpawnParams& params)
     , _transform(Transform::Identity)
     , _sphere(BoundingSphere::Empty)
     , _box(BoundingBox::Zero)
+    , _physicsScene(nullptr)
     , HideFlags(HideFlags::None)
 {
 }
@@ -1808,4 +1810,27 @@ void Actor::FromJson(const StringAnsiView& json)
     Deserialize(document, &*modifier);
     Scripting::ObjectsLookupIdMapping.Set(nullptr);
     OnTransformChanged();
+}
+
+void Actor::SetPhysicsScene(PhysicsScene* scene)
+{
+    ASSERT(scene);
+
+    auto previous = GetPhysicsScene();
+    _physicsScene = scene;
+
+    if (previous != _physicsScene)
+        OnPhysicsSceneChanged(previous);
+
+    // cascade
+    for (auto child : Children)
+        child->SetPhysicsScene(scene);
+}
+
+PhysicsScene* Actor::GetPhysicsScene()
+{
+    if (_physicsScene == nullptr)
+        _physicsScene = Physics::DefaultScene;
+
+     return _physicsScene;
 }
--- a/Source/Engine/Level/Level.cpp
+++ b/Source/Engine/Level/Level.cpp
@@ -22,6 +22,7 @@
 #include "Engine/Profiler/ProfilerCPU.h"
 #include "Engine/Scripting/Script.h"
 #include "Engine/Engine/Time.h"
+#include "Engine/Physics/Physics.h"
 #include "Engine/Scripting/ManagedCLR/MAssembly.h"
 #include "Engine/Scripting/ManagedCLR/MClass.h"
 #include "Engine/Scripting/ManagedCLR/MDomain.h"
@@ -164,6 +165,8 @@ bool LevelImpl::spawnActor(Actor* actor, Actor* parent)
         }
         if (parent == nullptr)
             parent = Level::Scenes[0];
+
+        actor->SetPhysicsScene(parent->GetPhysicsScene());
         actor->SetParent(parent, true, true);
     }
 
@@ -776,6 +779,8 @@ bool LevelImpl::unloadScene(Scene* scene)
     // Force flush deleted objects so we actually delete unloaded scene objects (prevent from reloading their managed objects, etc.)
     ObjectsRemovalService::Flush();
 
+    Physics::EndPlay();
+
     return false;
 }
 
@@ -1025,6 +1030,8 @@ bool Level::loadScene(rapidjson_flax::Value& data, int32 engineBuild, Scene** ou
     // Link scene and call init
     {
         PROFILE_CPU_NAMED("BeginPlay");
+
+        Physics::BeginPlay();
 
         ScopeLock lock(ScenesLock);
         Scenes.Add(scene);
--- a/Source/Engine/Level/Level.cpp
+++ b/Source/Engine/Level/Level.cpp
@@ -22,7 +22,6 @@
 #include "Engine/Profiler/ProfilerCPU.h"
 #include "Engine/Scripting/Script.h"
 #include "Engine/Engine/Time.h"
-#include "Engine/Physics/Physics.h"
 #include "Engine/Scripting/ManagedCLR/MAssembly.h"
 #include "Engine/Scripting/ManagedCLR/MClass.h"
 #include "Engine/Scripting/ManagedCLR/MDomain.h"
@@ -779,8 +778,6 @@ bool LevelImpl::unloadScene(Scene* scene)
     // Force flush deleted objects so we actually delete unloaded scene objects (prevent from reloading their managed objects, etc.)
     ObjectsRemovalService::Flush();
 
-    Physics::EndPlay();
-
     return false;
 }
 
@@ -1030,8 +1027,6 @@ bool Level::loadScene(rapidjson_flax::Value& data, int32 engineBuild, Scene** ou
     // Link scene and call init
     {
         PROFILE_CPU_NAMED("BeginPlay");
-
-        Physics::BeginPlay();
 
         ScopeLock lock(ScenesLock);
         Scenes.Add(scene);
--- a/Source/Engine/Physics/Actors/RigidBody.cpp
+++ b/Source/Engine/Physics/Actors/RigidBody.cpp
@@ -2,7 +2,6 @@
 
 #include "RigidBody.h"
 #include "PxMaterial.h"
-#include <iterator>
 
 #include "Engine/Core/Log.h"
 #include "Engine/Physics/Utilities.h"
--- a/Source/Engine/Physics/Actors/RigidBody.cpp
+++ b/Source/Engine/Physics/Actors/RigidBody.cpp
@@ -2,10 +2,14 @@
 
 #include "RigidBody.h"
 #include "PxMaterial.h"
+#include <iterator>
+
+#include "Engine/Core/Log.h"
 #include "Engine/Physics/Utilities.h"
 #include "Engine/Physics/Colliders/Collider.h"
 #include "Engine/Physics/Physics.h"
 #include "Engine/Physics/PhysicalMaterial.h"
+#include "Engine/Physics/PhysicsScene.h"
 #include "Engine/Serialization/Serialization.h"
 #include <ThirdParty/PhysX/extensions/PxRigidBodyExt.h>
 #include <ThirdParty/PhysX/PxRigidActor.h>
@@ -476,7 +480,7 @@ void RigidBody::CreateActor()
 
     // Register actor
     const bool putToSleep = !_startAwake && GetEnableSimulation() && !GetIsKinematic() && IsActiveInHierarchy();
-    Physics::AddActor(_actor, putToSleep);
+    GetPhysicsScene()->AddActor(_actor, putToSleep);
 
     // Update cached data
     UpdateBounds();
@@ -580,7 +584,7 @@ void RigidBody::EndPlay()
     if (_actor)
     {
         // Remove actor
-        Physics::RemoveActor(_actor);
+        GetPhysicsScene()->RemoveActor(_actor);
         _actor = nullptr;
     }
 }
@@ -626,4 +630,16 @@ void RigidBody::OnTransformChanged()
         // Base
         PhysicsActor::OnTransformChanged();
     }
+}
+
+void RigidBody::OnPhysicsSceneChanged(PhysicsScene* previous)
+{
+    ASSERT(previous);
+
+    PhysicsActor::OnPhysicsSceneChanged(previous);
+
+    previous->UnlinkActor(_actor);
+
+    const bool putToSleep = !_startAwake && GetEnableSimulation() && !GetIsKinematic() && IsActiveInHierarchy();
+    GetPhysicsScene()->AddActor(_actor, putToSleep);
 }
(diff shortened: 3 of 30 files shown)


mafiesto4:
Great stuff! I like this approach. I thought about introducing concept of `World` that would consist a list of loaded scenes that can be simulated together but this might be even more flexible.

iainmckay:
@mafiesto4 I've been through and added support for all the physical entities in Flax and added what I have determined to be all the required functionality to fully support this feature.

What would you like to validate this feature? I haven't spotted a test suite to add tests, perhaps I'm missing it?

iainmckay:
I see the Linux build pipeline is failing. I'll look in to it.

Source/Engine/Level/Actor.cpp
```diff
@@ -1806,3 +1808,26 @@ void Actor::FromJson(const StringAnsiView& json)
...
+    ASSERT(scene);
+
+    auto previous = GetPhysicsScene();
+    _physicsScene = scene;
+
+    if (previous != _physicsScene)
+        OnPhysicsSceneChanged(previous);
+
+    // cascade
+    for (auto child : Children)
```
mafiesto4, line 1823: Call `SetPhysicsScene` for children only if physics scene gets changed for this actor.

Source/Engine/Physics/Actors/RigidBody.cpp
```diff
@@ -2,10 +2,14 @@
 
 #include "RigidBody.h"
 #include "PxMaterial.h"
+#include <iterator>
```
mafiesto4, line 5: Unused iterator include + remove gab in includes.

Source/Engine/Physics/Colliders/CharacterController.cpp
```diff
@@ -7,9 +7,10 @@
 #include "Engine/Serialization/Serialization.h"
 #include "Engine/Engine/Time.h"
 #include "Engine/Physics/PhysicalMaterial.h"
+#include "Engine/Physics/PhysicsScene.h"
+
```
mafiesto4, line 11: Unnecessary gap between includes

Source/Engine/Physics/Colliders/Collider.h
```diff
@@ -232,4 +232,6 @@ DECLARE_SCENE_OBJECT_ABSTRACT(Collider);
     void OnParentChanged() override;
     void OnTransformChanged() override;
     void OnLayerChanged() override;
+
```
mafiesto4, line 235: Remove gap.

Source/Engine/Physics/Colliders/SplineCollider.cpp
```diff
@@ -8,12 +8,14 @@
...
 #include "Engine/Physics/Utilities.h"
 #include "Engine/Physics/Physics.h"
+#include "Engine/Physics/PhysicsScene.h"
 #include "Engine/Profiler/ProfilerCPU.h"
 #if COMPILE_WITH_PHYSICS_COOKING
 #include "Engine/Physics/CollisionCooking.h"
 #include <ThirdParty/PhysX/PxPhysics.h>
 #include <ThirdParty/PhysX/extensions/PxDefaultStreams.h>
 #endif
+
```
mafiesto4, line 18: Another gap.

Source/Engine/Physics/Physics.cpp
```diff
@@ -422,45 +324,8 @@ bool PhysicsService::Init()
...
-    }
-    else
-    {
-        LOG(Info, "Missing PVD client scene.");
-    }
-#endif
-
-    // Init characters controller
-    ControllerManager = PxCreateControllerManager(*PhysicsScene);
+    Physics::DefaultScene = new PhysicsScene(String("Default"), settings, cpuInfo);
```
mafiesto4, line 327: Use `New<PhysicsScene>(TEXT("Default"))` as a default memory allocator in Flax.

Source/Engine/Physics/Physics.h
```diff
@@ -88,41 +90,44 @@ DECLARE_SCRIPTING_TYPE_NO_SPAWN(Physics);
...
+    API_FIELD(ReadOnly)
+    static PhysicsScene* DefaultScene;
 
     /// <summary>
-    /// Gets the default query filter callback used for the scene queries.
+    /// List with all physics scenes (readonly).
     /// </summary>
-    static PxQueryFilterCallback* GetQueryFilterCallback();
+    API_FIELD(ReadOnly)
+    static Array<PhysicsScene*> Scenes;
```
mafiesto4, line 114: Can we use forward declaration `Array<PhysicsScene*, HeapAllocation>` and remove include to Array in `Physics.h`?

Source/Engine/Physics/Physics.h
```diff
@@ -514,11 +519,22 @@ DECLARE_SCRIPTING_TYPE_NO_SPAWN(Physics);
...
+
     /// <summary>
     /// Called during main engine loop to collect physic simulation results and apply them as well as fire collision events.
     /// </summary>
     API_FUNCTION() static void CollectResults();
 
+    /// <summary>
+    /// Called during main engine loop to collect physic simulation results on all registered scenes and apply them as well as fire collision events.
+    /// </summary>
+    static void CollectResultsAll();
```
mafiesto4, line 536: Do we need separate `CollectResults` and `CollectResultsAll`? For me, it could by just `CollectResults` and collect results from all scenes by default.

Source/Engine/Physics/Physics.h
```diff
@@ -573,5 +589,9 @@ DECLARE_SCRIPTING_TYPE_NO_SPAWN(Physics);
     /// Marks that joint has been removed (all collision events should be cleared to prevent leaks of using removed object).
     /// </summary>
     /// <param name="joint">The joint.</param>
-    static void RemoveJoint(Joint* joint);
+    static void RemoveJointAll(Joint* joint);
```
mafiesto4, line 592: I think we could stick to the previous name.

Source/Engine/Physics/PhysicsScene.cpp
```diff
@@ -0,0 +1,770 @@
...
+    {
+        return PxQueryHitType::eBLOCK;
+    }
+
+    return PxQueryHitType::eNONE;
+}
+
+#endif
+
+PhysicsScene::PhysicsScene(String name, PhysicsSettings settings, CPUInfo cpuInfo)
```
mafiesto4, line 95: Use `const String& name` to optimize memory allocations.
Pass structure param as reference (`const PhysicsSettings&`).
Maybe skip `CPUInfo` in  ctor param and just read it from `Platform` manually here to make the setup easier for the user?

Source/Engine/Physics/PhysicsScene.h
```diff
@@ -0,0 +1,560 @@
...
+
+#include "SimulationEventCallback.h"
+
+#include "Engine/Scripting/ScriptingObject.h"
+#include "Engine/Scripting/ScriptingType.h"
+#include "Engine/Core/Collections/Array.h"
+#include "Engine/Core/Math/Vector3.h"
+#include "Engine/Core/Math/Quaternion.h"
+#include "Types.h"
+#include <ThirdParty/PhysX/PxPhysicsAPI.h>
```
mafiesto4, line 11: I'd prefer to not include any PhysX headers in `PhysicsScene.h`

Source/Engine/Level/Level.cpp
```diff
@@ -776,6 +779,8 @@ bool LevelImpl::unloadScene(Scene* scene)
     // Force flush deleted objects so we actually delete unloaded scene objects (prevent from reloading their managed objects, etc.)
     ObjectsRemovalService::Flush();
 
+    Physics::EndPlay();
```
mafiesto4, line 782: What's the purpose of this call?

Source/Engine/Level/Level.cpp
```diff
@@ -1026,6 +1031,8 @@ bool Level::loadScene(rapidjson_flax::Value& data, int32 engineBuild, Scene** ou
     {
         PROFILE_CPU_NAMED("BeginPlay");
 
+        Physics::BeginPlay();
```
mafiesto4, line 1034: What's that?

mafiesto4 (changes requested):

Source/Engine/Physics/Actors/RigidBody.cpp
iainmckay: Done

Source/Engine/Level/Actor.cpp
iainmckay: Done

Source/Engine/Physics/Colliders/CharacterController.cpp
iainmckay: Done

Source/Engine/Physics/Colliders/Collider.h
iainmckay: Done

Source/Engine/Physics/Physics.h
iainmckay: Done

Source/Engine/Physics/PhysicsScene.cpp
iainmckay: Done, I've gone with your suggestion regarding `CPUInfo`. This was just my preference of passing in dependencies.

Source/Engine/Level/Level.cpp
iainmckay: This, paired with `BeginPlay`, is used to reset `Physics::Scenes`. I couldn't find another clean way to achieve this, without it `Physics::Scenes` continues to grow between loads and eventually crashes because the scenes hold references to things that don't exist anymore.

@mafiesto4 Do you have a more flax-ish method or suggestion on how to achieve this? If my method was to be kept I had thought about generalising it so that `Begin/EndPlay` was a member of `EngineService` then any service could hook on to these callbacks.

Source/Engine/Level/Level.cpp
iainmckay: See comments above regarding `EndPlay`.

Source/Engine/Physics/PhysicsScene.h
iainmckay: OK, I got rid of the include by implementing a pimpl instead. C++ isn't my strongest language so if you have a preferred alternate method of resolving your request let me know and I'll implement that instead.

Source/Engine/Level/Level.cpp
mafiesto4: Empty Physics Scenes should be left to stay around as they might be used when the next level gets loaded (eg. additional level streamed in or map changed). If game want, it can manage manually the Physics Scene lifetime by removing it with Destroy (you could add unregistering for PhysicsScene in destructor). We can just destroy all scenes on physics system unload.

iainmckay: Let me revisit this. I had engine crashes across subsequent play/stop cycles in the editor, IIRC some collision events were persisting across PIE sessions.

Source/Engine/Physics/Physics.cpp
iainmckay: Thanks, I replaced this with a call to `Physics::FindOrCreateScene` to centralize where `PhysicsScene`'s are created from and replaced the `new ` with the flax allocator.

Source/Engine/Physics/Physics.h
iainmckay: Yes, this is part of maintaining backwards compatibility. `CollectResults` will internally call `CollectResults` on the default scene and therefore maintains the ability of existing projects to manually collect results without needing to change their code.

 `CollectResultsAll` is called by the engine to collect results on all registered scenes. I think it's important to maintain this distinction.

Source/Engine/Physics/Physics.h
iainmckay: Done

Source/Engine/Level/Level.cpp
iainmckay: OK, so it appears that the issue I originally had has been resolved as the solution became more stable and I've been able to remove the Begin/EndPlay additions.

However, this feels leaky in the editor since now the (empty) physics scenes will persist across multiple PIE sessions. If you think it should be the responsibility of the game to clean these up even in PIE sessions then we'll leave it as-is. My feeling is that exiting a PIE session should almost behave like restarting an exported game.

What would your decision be?

iainmckay:
@mafiesto4 Can I poke you to review the questions I raised?

Source/Engine/Physics/PhysicsScene.cpp
```diff
@@ -0,0 +1,818 @@
...
+    {
+        return PxQueryHitType::eBLOCK;
+    }
+
+    return PxQueryHitType::eNONE;
+}
+
+#endif
+
+class PhysicsScenePhysX
```
mafiesto4, line 96: What about making `PhysicsScenePhysX` to inherit from `PhysicsScene` and turning its API into virtual calls so we could in the future allow for using easily custom physics engine? (like Havok)

Source/Engine/Physics/PhysicsScene.cpp
```diff
@@ -0,0 +1,818 @@
...
+    Array<PxRaycastHit> WheelHitResults;
+    Array<PxWheelQueryResult> WheelVehiclesResultsPerWheel;
+    Array<PxVehicleWheelQueryResult> WheelVehiclesResultsPerVehicle;
+    PxBatchQuery* WheelRaycastBatchQuery = nullptr;
+    PxVehicleDrivableSurfaceToTireFrictionPairs* WheelTireFrictions = nullptr;
+    Array<WheeledVehicle*> WheelVehicles;
+#endif
+};
+
+PhysicsScene::PhysicsScene(const String& name, const PhysicsSettings& settings)
```
mafiesto4, line 128: `PhysicsScene` could contain `Init` method that would take those settings and return `true` on initialization fail.

mafiesto4 (commented):
Looks nice! I think changing `PhysicsScene` into virtual base class for physics system would allow us to separate PhysX impl and allow to plug other physics engine. Do you want to do this? If not I can merge it and take over from here 😄

Source/Engine/Physics/PhysicsScene.cpp
iainmckay: It makes sense but I think it's out of scope for this PR but I understand the appeal of making these changes all at once.

I can do this but I'd suggest merging this first and implementing the abstraction in another PR. There are more things that need to be abstracted such as the CharacterController and Vehicle as they have direct dependencies on PhysX. I'm sure this will involve discussion and it makes sense to me to keep the issues separate.

It's up to you though, we can implement both features in this branch if you feel strongly that way.

mafiesto4 (approved):

mafiesto4:
Ok, merged. It took quite some time but I've managed to refactor the whole `Physics` module to have separated PhysX implementation via `PhysicsBackend` interface. Now it will be easy to plug-in another physics engine (eg. Havok).
427846f73b265e6ed9bcc1511d7e69301fb33e4e

iainmckay:
Fantastic!

Flavelius:
I'm interested in the 'world' concept mentioned above (coming from godot, using SubViewports); now that PhysicsScene seems to replace that as a more 'powerful' alternative, is there a way to skip/control the rendering of a / per scene aswell?

mafiesto4:
@Flavelius I think the easiest way would be using Layers to mask certain levels for rendering or use custom SceneRenderTask with objects to draw - we could add a way to specify list of scenes instead of list of actors. In general, engine has concept of SceneRendering (hidden from user) that has optimized cache for draws scene actors so we could use that for multiple-separate-worlds drawing in future.

mafiesto4:
Added in 0db259e300b8204130a1fe891d8e07cc6f01abde. New docs with code example for this feature: https://github.com/FlaxEngine/FlaxDocs/commit/de3ed1ca4cdcd0c5cafdcbb7906df28ba76cbea6?short_path=ab4eb9e#diff-ab4eb9ece71c312e881cb26feb95b14ca7134bf27ece1bbffa3be301a086567c

PR #674, opened 2021-12-02
Miscellaneous improvements

intolerantape:
Pushing some of my local commits that I thought would be useful for other people.

**Implemented Math::SinCos(double) in Math.cpp.** The double version of this function was declared in Mathd.h but not implemented anywhere.

**Tweaked Dictionary.h to support movable value types with no copy constructors.** Dictionary::SetCapacity() was preventing proper support here because the old data was being copied instead of moved.

**Added a constructor to Function class in Delegate.h to support constructing directly from reference-captured lambdas.** Mafiesto graciously added support for reference captured lambda's, but I was too dumb to notice because they had to be initialized via the Bind() method. So I copy-pasted (most of) the body of the appropriate Bind() method into a new constructor and now I can initialize Function objects like this:

    Function<void()> funkytown01 = [&]() {};
    Function<void()> funkytown02([&]() {});

... which makes me very happy!

### DIFF
--- a/Source/Engine/Core/Collections/Dictionary.h
+++ b/Source/Engine/Core/Collections/Dictionary.h
@@ -530,7 +530,7 @@ public:
             for (int32 i = 0; i < oldSize; i++)
             {
                 if (oldData[i].IsOccupied())
-                    Add(oldData[i].Key, oldData[i].Value);
+                    Add(oldData[i].Key, MoveTemp(oldData[i].Value));
             }
         }
         if (oldElementsCount != 0)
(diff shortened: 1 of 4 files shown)


intolerantape:
**Added overload of Task::StartNew() to support member functions that return void.** This seems to have been overlooked, so I added it. The new function is simply a copy-paste of the corresponding bool-returning overload, but with the action's return value changed to bool.

mafiesto4:
Thanks! I did a little change top reuse `Bind<T>` code for lambda: 2daec98fd7b483af73fc100b024ed264954a1713

PR #675, opened 2021-12-06
Various Xaudio2 backend fixes

GoaLitiuM:
Fixes a crash in audio clip editor when Play-button is pressed. This probably should be fixed elsewhere later, but for some reason the XAudio2 backend creates the audio source when the parent is assigned and not when the clip is assigned. The initializer assigns the parent before clip which skips the audio source creation and leads to crash when the audio source is expected to be loaded when it's being played.

Also fixes the audio clips volume levels to correct levels, before the fix the volume of the audio clips were more quieter than expected.

### DIFF
--- a/Source/Editor/Windows/Assets/AudioClipWindow.cs
+++ b/Source/Editor/Windows/Assets/AudioClipWindow.cs
@@ -168,11 +168,10 @@ namespace FlaxEditor.Windows.Assets
             }
             if (!_previewSource)
             {
-                _previewSource = new AudioSource
-                {
-                    Parent = _previewScene,
-                    Clip = _asset,
-                };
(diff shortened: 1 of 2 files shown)


mafiesto4:
Thanks for pointing this out! 😸 
I did the fix for XAudio here: 1b34b3b63ae57b80a28d0983640eb0583648cf7a without modifying the `AudioClipWindow`. Then I tweaked those coeffs for the audio matrix here: 660c8ef95c70abf2ad6bea9efc49a5d9d1381adb based on the docs from here: https://docs.microsoft.com/en-us/windows/win32/api/xaudio2/nf-xaudio2-ixaudio2voice-setoutputmatrix

PR #681, opened 2021-12-24
Fix blend shape offsets

xKamuna:
Blend shapes weren't taking into account the automatic splits that FlaxEngine does when you have multiple materials.  This pull request fixes missing blend shapes due to this issue, using existing code as formatting inspiration.


mafiesto4:
Thanks!

PR #682, opened 2021-12-31
Fix issue with filters being ignored for two controllers colliding

xKamuna:
Adds CCT Callbacks to physics queries so that layer collision settings are respected.  Tried to follow the existing code as much as possible, let me know if anything should be removed, added, or edited.


mafiesto4:
Thanks!

PR #687, opened 2022-02-04
Update CSProjectGenerator to support .NET Framework 4.8 Targeting Pack

PrecisionRender:
Right now, every time we create a new C# script, Flax Engine defaults to setting up the C# project file to use the .NET Framework 4.5 Targeting Pack, meaning we need to retarget the project file each time to work with the 4.8 Framework Targeting Pack, which is the newest version. In Visual Studio, this is rather simple, as we just need to press 2 buttons, but in Visual Studio Code, it requires going into the `.csproj` file and changing it ourselves, which is less than ideal.

This PR just changes the generated `.csproj` file to use the 4.8 Framework Targeting Pack by default.

(This is my very first PR on any source-available project. I hope that I'm doing it properly. 🙂)


mafiesto4:
Actually, that's because Flax uses .Net 4.5 only right now. I guess it would be great to have a detection system for framework version that is installed for Visual Studio and use a newer one if possible.

PrecisionRender:
Oh, okay. I think that would be a nice feature to have, but not super important.

mafiesto4:
Merged to `mono-6` branch (https://github.com/FlaxEngine/FlaxEngine/commit/a4be261d464b97d60e279309c6a57b4a342c87db). Thanks!

PR #691, opened 2022-02-18
Add cursor clipping

ScottLongley:
Cursor clipping confines the cursor within a specified rectangle, which is necessary for games that use edge-scrolling (RTS, MOBA, etc.)

I'm submitting this as a draft as the code is for Windows only. If you'd like to make use of this, feel free to compile this branch and set CursorLock to CursorLockMode.Clipped.


GoaLitiuM:
I would personally extend the `CursorLockMode` to have another mode for clipping the cursor within the game window. I had this version of clipping implemented in my personal fork but I've probably lost the code already.

mafiesto4:
Nice. Good idea with using enum for this.

PR #698, opened 2022-03-17
Enable DXGI flip presentation mode and tearing in D3D11 mode

GoaLitiuM:
Flip presentation is already used in D3D12 mode, so why not in D3D11 as well? This is only supported in Windows 10 or later systems, for older systems the non-flip modes are used instead.


GoaLitiuM:
Fullscreen mode wasn't working in this new flip presentation mode which is also used in D3D12 mode, so that should work in both modes now.

PR #699, opened 2022-03-19
Always run fixed update ticks at fixed deltatime

GoaLitiuM:
The random variance in fixed updates makes it impossible to do anything deterministic during physics ticks.

### DIFF
--- a/Source/Engine/Engine/Time.cpp
+++ b/Source/Engine/Engine/Time.cpp
@@ -104,7 +104,7 @@ bool Time::TickData::OnTickBegin(float targetFps, float maxDeltaTime)
 
         if (targetFps > ZeroTolerance)
         {
-            int skip = (int)(1 + (time - NextBegin) / (1.0 / targetFps));
+            const int skip = (int)(1 + (time - NextBegin) / (1.0 / targetFps));
             NextBegin += (1.0 / targetFps) * skip;
         }
     }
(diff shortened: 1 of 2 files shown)


Erdroy:
I'd suggest to test this with my PR #591 
Maybe it is the time, to get it merged too.

GoaLitiuM:
There is no catchup/frameskip mechanism in FixedStepTicks anymore with these changes, so the physics will run slowly when the target FPS is not met, and runs really fast after catching up. The simple fix for this is to just scrap the whole FixedStepTickData for physics and use the same TickData in all three ticking systems.

I'm not entirely sure why the physics ticking worked differently previously compared to update and render ticking. The old framerate smoothing technique definitely does not fit there as it causes slight variances in the deltatime.

GoaLitiuM:
Closing in favor of #703

GoaLitiuM:
I was a bit too hasty with this, I'm changing this PR to Draft while working on another solution to this. Even if the frameskipping is implemented to this method, the physics will still run slowly at irregular intervals (but with correct deltaTime).

Should the fixed updates also use sub-stepping similarly how the physics engine has these options? So for example if PhysicsFPS is set to 60 and game is running so slowly updates and draws happen at 30fps, should fixed updates still occur 60 times a second (or two times per drawn/updated frame in this case)?

PR #701, opened 2022-03-22
Implement RawInput

GoaLitiuM:
This is still very much work in progress. More work is required for example to properly store the mouse deltas in the engine instead of calculating the deltas from previous input state. Also runtime detection for device changes is still not implemented at this stage.


GoaLitiuM:
Closing this in favor of #2800 which includes support for raw input.

PR #702, opened 2022-03-26
Improve font rendering and alignment with custom DPI

GoaLitiuM:
Some screen DPI scaling settings (notably the 125% DPI in Windows) caused some characters to not align properly with rest of the characters in the same line:
before

After the fix:
after


mafiesto4:
Thanks!

PR #714, opened 2022-05-24
Free p variable after usage.

AlexandruApostol33:


mafiesto4:
What do you mean here? Any commit from `gi` branch fixes something on `master`?

PR #716, opened 2022-05-29
Add "View Size" node for GUI materials

GoaLitiuM:
This works the same way as the "Screen Size" node, but it returns the size of the rectangle where material is projected in the screen. For example in Material editor, the "Screen Size" will return the size of the window, but this node will return the size of the preview area where the material is rendered. When rendering full screen GUI materials in Editor / Game windows, the actual size of the preview area is returned in this node.


GoaLitiuM:
Is this good otherwise? I'm still unsure about the name of the node itself.

mafiesto4:
Yeah it might be useful, especially when doing some per-pixel masking or gradients that should know about control dimensions.

mafiesto4:
Ah also you can bump the `MATERIAL_GRAPH_VERSION` number up so it will update any existing GUI material shaders (you've added a new constant to the shader so the shader has to match the runtime code). (https://github.com/FlaxEngine/FlaxEngine/blob/master/Source/Engine/Graphics/Materials/MaterialShader.h#L13)

GoaLitiuM:
Bumped up the graph version

PR #721, opened 2022-06-05
Support dispatching jobs within jobs in JobSystem

GoaLitiuM:
The job system does not properly keep track of multiple batched jobs if other jobs were pushed to the job queue before the previously submitted batch was finished. The previous implementation waits until the dispatched amount of jobs were finished, which might include unrelated jobs queued around the same time to the system. This change ensures all the individual jobs in the submitted batch of jobs is completely executed when waited upon.

### DIFF
--- a/Source/Engine/Threading/JobSystem.cpp
+++ b/Source/Engine/Threading/JobSystem.cpp
@@ -53,6 +53,7 @@ struct JobData
 {
     Function<void(int32)> Job;
     int32 Index;
+    int64 JobKey;
 };
 
 template<>
@@ -82,6 +83,12 @@ public:
     }
 };
 
+struct JobContext
+{
+    volatile int64 RefCount = 0;
+    volatile int64 JobsLeft = 0;
+};
+
 namespace
 {
     JobSystemService JobSystemInstance;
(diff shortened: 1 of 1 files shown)


Source/Engine/Threading/JobSystem.cpp
```diff
@@ -256,29 +282,37 @@ int64 JobSystem::Dispatch(const Function<void(int32)>& job, int32 jobCount)
 #endif
 }
 
-void JobSystem::Wait()
-{
-#if JOB_SYSTEM_ENABLED
-    Wait(Platform::AtomicRead(&NextLabel));
```
mafiesto4, line 262: I guess, this one should wait for all active contexts?

Source/Engine/Threading/JobSystem.cpp
```diff
@@ -221,9 +239,18 @@ int64 JobSystem::Dispatch(const Function<void(int32)>& job, int32 jobCount)
 #if JOB_SYSTEM_USE_STATS
     const auto start = Platform::GetTimeCycles();
 #endif
+    const auto label = Platform::InterlockedAdd(&JobLabel, (int64)jobCount) + jobCount;
 
     JobData data;
     data.Job = job;
+    data.JobKey = label;
+
+    auto context = New<JobContext>();
```
mafiesto4, line 248: Could we skip memory dynamic allocations or just reuse them from jobs contexts pool (eg. 8?)

GoaLitiuM: Actually I have to revert this change, the dictionary holding the context might get reallocated when it grows so it would invalidate the existing references when waiting for the job to finish.

mafiesto4:
Merged. This fixes the problem of enqueueing and waiting for jobs from 2 threads at once. Improved a little in ec7aece1353582992629ccf53730926cafc82229

PR #728, opened 2022-06-16
EnvironmentProbe selectable resolution

whocares77:
Added the ability to select cubemap 
resolution for Environment Probe actors (per-actor).
Also you can set the default resolution in the Graphics Settings.

But as i said to @mafiesto4 dm, assertion `EnvironmentProbe.cpp:89` fails when you press Bake button and Resolution is not equal 128. I have no idea why it happens and how to fix this.


whocares77:
btw you must restart the editor after change any setting in Graphics Settings, is this a bug? @mafiesto4

mafiesto4:
@whocares77 Thanks for the contribution! I did some minor cleanup here: a8579cadcc2a95d4a43533def97ad49f0a13dace. Works pretty good 😄

PR #735, opened 2022-08-24
Add World Triplanar Texture node to shader editor

PrecisionRender:
Adds the World Triplanar Texture node to the Material editor. The node projects a texture using world-space coordinates instead of using object UVs. This is very useful in cases where a texture should not move with an object, such as with prototype materials.
image

Before:

https://user-images.githubusercontent.com/89754713/186447931-0aff1e49-d2c5-425f-ab01-7d460a8685ac.mp4

After:

https://user-images.githubusercontent.com/89754713/186447963-836f1935-ca2b-4652-a04c-be2074b44a61.mp4

Test project: [MaterialTest.zip](https://github.com/FlaxEngine/FlaxEngine/files/9417120/MaterialTest.zip)

### DIFF
--- a/Source/Engine/Tools/MaterialGenerator/MaterialGenerator.Textures.cpp
+++ b/Source/Engine/Tools/MaterialGenerator/MaterialGenerator.Textures.cpp
@@ -444,6 +444,48 @@ void MaterialGenerator::ProcessGroupTextures(Box* box, Node* node, Value& value)
         value = box == gradientBox ? gradient : distance;
         break;
     }
+    // World Triplanar Texture
+    case 16:
+    {
+        // Get input boxes
+        auto textureBox = node->GetBox(0);
+        auto scaleBox = node->GetBox(1);
+        
+        if (!textureBox->HasConnection())
+        {
+            // No texture to sample
+            value = Value::Zero;
+            break;
+        }
+
+        if (!CanUseSample(_treeType))
+        {
+            // Must sample texture in pixel shader
+            value = Value::Zero;
+            break;
+        }
+
+        const auto texture = eatBox(textureBox->GetParent<Node>(), textureBox->FirstConnection());
+        const auto scale = tryGetValue(scaleBox, node->Values[0]);
+
+        auto result = writeLocal(Value::InitForZero(ValueType::Float4), node);
+
+        const String triplanarTexture = String::Format(TEXT(
+            "   float3 worldPos = input.WorldPosition.xyz * {1} / 1000;\n"
+            "   float3 normal = input.TBN[2];\n"
+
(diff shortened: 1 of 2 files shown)


Source/Engine/Tools/MaterialGenerator/MaterialGenerator.Textures.cpp
```diff
@@ -444,6 +444,48 @@ void MaterialGenerator::ProcessGroupTextures(Box* box, Node* node, Value& value)
...
+        const auto texture = eatBox(textureBox->GetParent<Node>(), textureBox->FirstConnection());
+        const auto scale = tryGetValue(scaleBox, node->Values[0]);
+
+        auto result = writeLocal(Value::InitForZero(ValueType::Float4), node);
+
+        const String triplanarTexture = String::Format(TEXT(
+            "   float3 worldPos = input.WorldPosition.xyz * {1} / 1000;\n"
+            "   float3 normal = input.TBN[2];\n"
+
+            "   {2} += {0}.Sample(SamplerLinearWrap, worldPos.yz) * abs(dot(normal, float3(1,0,0)));\n"
```
mafiesto4, line 477: Optionally: provide `float Blend = 1.0f` input to the node and apply it here to control the blending threshold:
```hlsl
{2} += {0}.Sample(SamplerLinearWrap, worldPos.yz) * pow(abs(dot(normal, float3(1,0,0)), {3}));
// {3} = Blend
```
(would be nice to have so the transition on the object edges could be bigger or smaller)

Source/Engine/Tools/MaterialGenerator/MaterialGenerator.Textures.cpp
```diff
@@ -444,6 +444,48 @@ void MaterialGenerator::ProcessGroupTextures(Box* box, Node* node, Value& value)
...
+            break;
+        }
+
+        const auto texture = eatBox(textureBox->GetParent<Node>(), textureBox->FirstConnection());
+        const auto scale = tryGetValue(scaleBox, node->Values[0]);
+
+        auto result = writeLocal(Value::InitForZero(ValueType::Float4), node);
+
+        const String triplanarTexture = String::Format(TEXT(
+            "   float3 worldPos = input.WorldPosition.xyz * {1} / 1000;\n"
```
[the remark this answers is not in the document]
PrecisionRender: Fixed.

Source/Engine/Tools/MaterialGenerator/MaterialGenerator.Textures.cpp
PrecisionRender: Fixed.

mafiesto4:
Thanks! In a1dea6a756bbcb365c7620dacfbc0778c3f31d05 I added `{}` so the variable names won't conflict when using this node multiple times and adjusted the blend weight usage so it works correctly (my code I gave you before was invalid, now it works good).
image

PR #736, opened 2022-09-04
Prefer LLVM linker over other linkers on Linux

GoaLitiuM:
LLVM's linker performed much faster than the default linker clang is using (LD?), so why not use it instead when it's detected in the environment.


mafiesto4:
Yeah we could have better usage of linker on Clang. But the build on Github Actions is not failing.

GoaLitiuM:
Some quick benchmarks of different linkers (includes the compilation time of BinaryModule.cpp in this case, average time from 4 different runs):

- ld.bfd: 5.36s (the current default linker)
- ld.gold: 3.89s
- ld.lld: 1.93s

mafiesto4:
On old Ubuntu laptop with HDD the diff is even bigger:
```
Before: 17s
After: 5s
```

PR #742, opened 2022-09-07
Faster results with less allocate

GroophyLifefor:


GroophyLifefor:
Linq has special allocate optimizations and you can get results faster with exceptions.

GroophyLifefor:
This is a very small change and the effect will be very low, but there is a proverb in turkey, "drop by drop, it becomes a lake."

stefnotch:
@GroophyLifefor 
Linq has special optimizations, because it's usually slower and much harder to optimise than a plain for loop.
Btw, check out Benchmark.NET if you want to profile minor code tweaks like that. Do keep in mind that Flax uses a different .NET runtime though.

GroophyLifefor:
@stefnotch 
You're right sir, this pr useless, sorry mafiesto.

PR #749, opened 2022-09-15
Update SplashScreen.cpp

kerelape:


mafiesto4:
Is it some pop culture or gaming-reference?

kerelape:
> Is it some pop culture or gaming-reference?

reference to red engine

PR #759, opened 2022-10-01
Adding visual cursor changes when sliding value box or resizing docked window

Tryibion:
This change is a UX change. Having the cursor change when hovering over the "begin sliding" icon and while sliding the value allows the user to better understand what action they should take and allows for a better visual experience. Having the cursor change when hovering over or when sliding to resize a docked window allows for a much better experience for the user because they will have a visual cue of what can happen when being over the "re-size" rectangle. Other game engine use a similar functionality and it would be nice for Flax to follow suit.


mafiesto4:
Cool! I added small fix ebd88a407d50bcc90503256eba2361daf95fecc0 to properly restore cursor after mouse leaves the control area.

Issue #680, opened 2021-12-21
Incorrect transform gizmo depth order


PrecisionRender:
**Issue description:**
The transform gizmo is arguably one of the most important tools in a 3D workspace. Unfortunately, the gizmos in Flax Engine tend to be visually confusing. Here are a few examples:

https://user-images.githubusercontent.com/89754713/146993207-d4744914-02b4-4b93-a8a2-86168f383482.mp4

Is the Z axis closer to me? Or is the X axis? I can't tell until I move up.

https://user-images.githubusercontent.com/89754713/146992151-bdcceee8-d5ab-4571-8b68-b780882b8cf1.mp4

Which direction is Z, and which is X? I'm not too sure.

https://user-images.githubusercontent.com/89754713/146992235-b6b974f4-1c9d-468b-a855-85a0cf860a4d.mp4

Same problem as translate gizmo.

Not only is this kind of annoying to work with, but it can be disorienting since we don't have a dedicated tool for aligning ourselves with the world like in UE or Unity.

https://user-images.githubusercontent.com/89754713/146992851-de724603-95cd-448a-b7c3-65ab5041027d.mp4

Maybe we could have something like in UE4, where the gizmos have proper sorting, perspective distortion, and mesh lighting around them. In UE4, you can rely on the gizmos to properly represent their axis' direction. Or maybe we can at least have an XYZ gizmo somewhere in the editor like in UE4's editor.

**Steps to reproduce:**
Select an actor and look around the editor.

**Minimal reproduction project:**
None needed.

**Flax version:**
1.2.6226

plemsoft:
@mafiesto4 Fixed this so draw order is working, To fix the rotate, you can add new "wider" circleMesh (like UE), so direction is visible. Will make a pull later just so others people do no start working on this :)

mafiesto4:
@plemsoft Cool! Feel free to open PR once you're ready to merge the fix

PR #765, opened 2022-10-11
Changes/fixes.

plemsoft:
Added - ScriptingObject* TryFindObject(MClass* mclass), Allow you from c++ to use MAssembly GetBinaryModuleGame, and get a monoobject from a class, monoobject is needed later to Invoke a method from the class.

Fixed - Gizmo draw order ( https://github.com/FlaxEngine/FlaxEngine/issues/680 )

Added - aiProcess_SortByPType to importer, i seen many object that have a small mesh with point faces that are not part of the object, so just ignoring these make the objects import.

Added - Support for LOD meshes using _ , like LOD_0...

Fixed - Windows, launcher use a uppercase D:\ , if you try to start your project from VS/Shortcut and use a lowercase d:\ it would ruin your project as it assign new asset IDs without remapping.

### DIFF
--- a/Source/Editor/Gizmo/TransformGizmoBase.Draw.cs
+++ b/Source/Editor/Gizmo/TransformGizmoBase.Draw.cs
@@ -54,7 +54,12 @@ namespace FlaxEditor.Gizmo
             if (!_isActive || !IsActive)
                 return;
 
-            Matrix m1, m2, m3;
+            //PE: As all axisMesh have the same pivot, add a little offset to the x axisMesh,
+            //PE: This way SortDrawCalls is able to sort the draw order.
+            //PE: https://github.com/FlaxEngine/FlaxEngine/issues/680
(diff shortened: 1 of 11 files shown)


mafiesto4:
Thanks for lots of improvements and fixes!

I did some further improvements:
- Cleanup to match engine codebase style - 4ba93d9d861b7779a1fd616388a2185e62740784
- Add unit test for `ModelTool::DetectLodIndex` to verify all cases - 5cfe3b88c744a5adc54162158b9b8066006bf68b 
- Optimize asset registry to use string comparison that ignores case (instead of string allocation) - c41248f4024873ccb31a6a68cb6b83d8fdddfa81
- Add undo/redo and navmesh dirtying when using selection rotate shortcut (R key) - f67d1412dee1eff6ef3ce8085c8a45349d5ad79f

PR #773, opened 2022-10-16
Added performing layout when deleting actors and scrolling to duplicated/ pasted actor when action is performed

Tryibion:
Hello, this PR is to add a nice QoL feature to the scene view. 

This adds the ability for the scene window to perform its layout when an actor is deleted to make sure it gets rid of unwanted empty space (not sure why I had to run the function 2 times, but it doesn't work with just 1). 

This also adds the ability for the scene window to move to the new selected actor after a duplicate or paste is performed, this is nice especially when needing to duplicate an actor many times.

### DIFF
--- a/Source/Editor/Modules/SceneEditingModule.cs
+++ b/Source/Editor/Modules/SceneEditingModule.cs
@@ -452,9 +452,11 @@ namespace FlaxEditor.Modules
             if (isSceneTreeFocus)
             {
                 Editor.Windows.SceneWin.Focus();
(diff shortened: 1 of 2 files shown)


Tryibion:
Before:

https://user-images.githubusercontent.com/71274967/197399805-012a4d58-beaa-4aeb-bc69-81db171d9ab6.mp4

After:

https://user-images.githubusercontent.com/71274967/197399818-ba3fe46e-1165-47ed-a656-e4c405f48957.mp4

PR #776, opened 2022-10-17
All Actor and Custom Actor Tabs in ToolBox

Tryibion:
This adds an all actor tab that shows all of the actors. This is useful because it lets the user browse all of the actors if they are searching for one but don't remember its name. I also added an exclude list, I think I got the ones the crash the editor or cause problems with workflows, but I wasn't sure if there were any others that should be excluded.

This also adds an attribute that can be put ahead of an actor class and will add it to the custom actors tab in the toolbox, this is useful because it lets users add their own actors to the toolbox if they often use them and don't want to search for them.

### DIFF
--- a/Source/Editor/Windows/ToolboxWindow.cs
+++ b/Source/Editor/Windows/ToolboxWindow.cs
@@ -2,10 +2,6 @@
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


Tryibion:
Closing in favor of #823

PR #781, opened 2022-10-20
added cursor change for moving the property splitter

Tryibion:
This is to follow along with the other cursor changes on resizing things. This adds a cursor change when mouse is over the property splitter as well as moving it.


Tryibion:
https://user-images.githubusercontent.com/71274967/197400319-4db5ba92-75ee-4b0f-9ccf-d2ef959d5ddf.mp4

PR #783, opened 2022-10-20
Added value boxes hiding cursor and returning it to its position when done

Tryibion:
Hello, this is a QoL addition. I always found it cumbersome to have to move my mouse back to the property panel after I had slid a value in a value box. This change makes the value box more like unreal engines where it will hide the cursor and then move the cursor back to its original position when done sliding.


Tryibion:
https://user-images.githubusercontent.com/71274967/197400010-032b5a56-64aa-4db4-9640-8f78d4ddb26e.mp4

PR #792, opened 2022-10-21
changed the default for script members order.

Tryibion:
Hello, this is often something asked in the Discord and often the first thing people change when they start with the engine. It may as well be set to Declaration by default as that is what people are used to coming from different engines

### DIFF
--- a/Source/Editor/Options/GeneralOptions.cs
+++ b/Source/Editor/Options/GeneralOptions.cs
@@ -137,9 +137,9 @@ namespace FlaxEditor.Options
         /// <summary>
         /// Gets or sets an order of script properties/fields in properties panel.
         /// </summary>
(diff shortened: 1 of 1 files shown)


stefnotch:
For posterity/documentation purposes: 

`MembersOrder.Declaration` relies on compiler internals behaving in a certain way. It's not very elegant, but it's one of the few ways that at least work at the moment.

https://github.com/FlaxEngine/FlaxEngine/blob/3c94814e9f6bd766663915e90633d4455328fa6c/Source/Editor/CustomEditors/Editors/GenericEditor.cs#L176-L179

honzapatCZ:
Man you've got some balls, when I made the original PR mafi accepted the sorting only optionally. 
Good Job pushing it thru.

PR #794, opened 2022-10-22
Small change to spacing in content view

Tryibion:
Hello, this adds a small amount of space between content items. It isn't a huge change, but gets rid of the issue of missing a column of items on the right side of the content view when at specific sizes and just feels a little better.


Tryibion:
Before:


https://user-images.githubusercontent.com/71274967/197400154-202fdb3b-e131-4f93-9f57-cd286d66c54f.mp4


After:


https://user-images.githubusercontent.com/71274967/197400188-409a8f50-bcf3-4ce4-9c73-975a15db4728.mp4

PR #809, opened 2022-10-28
Moved Material Instance override options into a settings icon with context menu

Tryibion:
Hello, I felt that the place of the parameter override all and remove override buttons are in a weird spot. This PR moves those buttons into a context menu that is accessible through a new settings icon on the toolstrip. 

New Button:

new toolstrip button

New CM:

newcm

### DIFF
--- a/Content/Editor/IconsAtlas.flax
+++ b/Content/Editor/IconsAtlas.flax
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:ff5774c1073e5d4e5de362f7f8a0a74822cacb0b074272d4bba341be0c92d4ed
-size 5611572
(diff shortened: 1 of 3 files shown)


Tryibion:
Hmm failed because of Git LFS permissions?

mafiesto4:
Yeah you can upload new icons atlas texture within this issue so I can manually update it in repo.

Also, we could add `Revert All Parameters` button to this context menu as well.

Tryibion:
IconsAtlas

Here is the atlas texture. I will add the revert all parameters to that as well.

Tryibion:
Another option is to put the setting icon on the properties header similar to the general setting for actors.

Tryibion:
Closing in favor of #2878

PR #810, opened 2022-10-28
organized the "new" child context menu in the content window into categories

Tryibion:
Hello, this PR organizes the "New" context menu in in the content window into categories as seen in the image below. See below comments for other improvements. 

new content cm

### DIFF
--- a/Source/Editor/Windows/ContentWindow.ContextMenu.cs
+++ b/Source/Editor/Windows/ContentWindow.ContextMenu.cs
@@ -157,9 +157,9 @@ namespace FlaxEditor.Windows
             // loop through each proxy and user defined json type and add them to the context menu
             foreach (var type in Editor.CodeEditing.All.Get())
             {
-                if (type.IsAbstract || !type.HasAttribute(typeof(ContentContextMenuAttribute), true))
+                if (type.IsAbstract || !type.HasAttribute(typeof(ContentContextMenuAttribute), true) || Editor.CodeEditing.Actors.Get().Contains(type) || Editor.CodeEditing.Scripts.Get().Contains(type))
                     continue;
-                
+
                 ContentContextMenuAttribute attribute = null;
                 foreach (var typeAttribute in type.GetAttributes(true))
                 {
--- a/Source/Editor/Windows/ContentWindow.ContextMenu.cs
+++ b/Source/Editor/Windows/ContentWindow.ContextMenu.cs
@@ -1,6 +1,7 @@
 // Copyright (c) 2012-2022 Wojciech Figat. All rights reserved.
 
 using System;
+using System.Collections.Generic;
 using FlaxEditor.Content;
 using FlaxEditor.GUI.ContextMenu;
 using FlaxEngine;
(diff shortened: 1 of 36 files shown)


brmassa:
As a bonus, a public API/function/Attribute to let users add new entries by code, like Unity's 
[CreateContectMenu("myGame/myCustomAsset")]
class myCustomAsset {}

Source/Editor/Windows/ContentWindow.ContextMenu.cs
```diff
@@ -159,7 +189,17 @@ private void ShowContextMenuForItem(ContentItem item, ref Float2 location, bool
                 var p = Editor.ContentDatabase.Proxy[i];
                 if (p.CanCreate(folder))
                 {
-                    c.ContextMenu.AddButton(p.Name, () => NewItem(p));
+                    if (_animationGroupNames.Contains(p.Name))
```
mafiesto4, line 192: I think it would be better to let `AssetProxy` return the name of the category and use it here to build nested menu - easier to extend and maintain in the future.

mafiesto4 (changes requested):

Source/Editor/Windows/ContentWindow.ContextMenu.cs
Tryibion: I am thinking of doing it this way. Each asset proxy will have its category name that is set. Then use something like this:

```csharp
 // categorize the asset proxies according to their category name
            Dictionary<string, List<AssetProxy>> categorizedProxyList = new Dictionary<string, List<AssetProxy>>();
            for (int i = 0; i < Editor.ContentDatabase.Proxy.Count; i++)
            {
                var p = Editor.ContentDatabase.Proxy[i];
                if (p.CanCreate(folder))
                {
                    if (p is AssetProxy ap)
                    {
                        if (categorizedProxyList.ContainsKey(ap.CategoryName))
                        {
                            categorizedProxyList.TryGetValue(ap.CategoryName, out var apList);
                            apList?.Add(ap);
                        }
                        else
                        {
                            categorizedProxyList.Add(ap.CategoryName, new List<AssetProxy>() {ap});
                        }
                    }
                }
            }
            
            c = cm.AddChildMenu("New");
            c.ContextMenu.Tag = item;
            c.ContextMenu.AutoSort = true;
            int newItems = 0;
            for (int i = 0; i < Editor.ContentDatabase.Proxy.Count; i++)
            {
                var p = Editor.ContentDatabase.Proxy[i];
                if (p.CanCreate(folder))
                {
                    if (p is AssetProxy ap && categorizedProxyList.TryGetValue(ap.CategoryName, out var apList))
                    {
                        if (apList.Contains(ap))
                        {
                            if (apList.Count > 1)
                            {
                                var childMenu = c.ContextMenu.GetOrAddChildMenu(ap.CategoryName);
                                childMenu.ContextMenu.AutoSort = true;
                                childMenu.ContextMenu.AddButton(p.Name, () => NewItem(p));
                            }
                            else
                            {
                                c.ContextMenu.AddButton(p.Name, () => NewItem(p));
                            }
                        }
                    }
                    else
                    {
                        c.ContextMenu.AddButton(p.Name, () => NewItem(p));
                    }

                    newItems++;
                }
            }
    ```

Can you think of an easier way besides using 2 loops?

Tryibion:
I think I came up with a better way to do this similar to how I did the actors context menu PR. Am working on it.

Tryibion:
Alright, did some work on this. Now this uses an attribute to put proxies and user defined json assets into the content context menu. anything that should be added can now put [ContentContextMenu(path)] before the proxy or class and it will add to the content context menu with what ever path is defined.

Proxy:
image
new content cm

User Defined:
image
new content custom cm

Tryibion:
> As a bonus, a public API/function/Attribute to let users add new entries by code, like Unity's [CreateContectMenu("myGame/myCustomAsset")] class myCustomAsset {}

made this... we will see if it gets approved though

brmassa:
amazing. 
This is not only add a great functionality for the user but unifies the api for internal use too! Congrats @Tryibion!

mafiesto4 (approved):

mafiesto4:
Added some docs: https://github.com/FlaxEngine/FlaxDocs/commit/b9901ba8c9b874dfe0e4bd214dbf8c082f6bab73

Issue #789, opened 2022-10-21
Crtl+S and/or auto-save on Prefab/JsonAsset dialog window


brmassa:
Would be great to auto-save Prefabs and JsonAsstes changes while in the dialog window, just like Actors edited in the inspector.
At least a Crtl+s to save via keyboard.

As a bonus, we have to press exit the field (pressing enter) in order to register the field change. If one opens an asset, changes a value and closes the window immediately, without "registering" the change, all is lost. Consider implementing input registration for each key pressed, just like the autocomplete field.

mafiesto4:
You can use Editor Options to configure auto-save: https://docs.flaxengine.com/manual/editor/options/index.html

brmassa:
Auto save content respects the Frequency value, which does not make sense when editing an asset. It should be instantaneous on each value changed.
The Ctrl+S shortcut also do not work on the dialog.
Should I open another issue?

mafiesto4:
Ah right, there is an issue with keyboard focus on Json Asset and input fields so Ctrl+S shortcut doesn't work.

mafiesto4:
Fixed in e713273942c43faf8c8e25e16701a0b1ba8d8a46

PR #812, opened 2022-10-31
fixes focus issues in json asset editor and prefab editor

Tryibion:
Hello, adding  a focus call once the scene is marked as edited fixes the issue with the text box returning focus to the window for different input actions like saving. Fix for #789.


Tryibion:
hmm actually this needs a bit more work because it effects sliders not being able to slide

PR #822, opened 2022-11-11
Rename on left click

Tryibion:
Hello, this PR allows for the item to be renamed when the user left clicks outside of the rename popup instead of having to only press enter to rename.


Tryibion:
looks like it timed out on the Mac check

mafiesto4:
It works fine but I'm not sure if it's not too big a change. It would be good to know what others think about it.

PR #827, opened 2022-11-15
PE: Fixes/changes and additions.

plemsoft:
PE: csharp allow setting material minimum and max depth, for on top rendering.
PE: allow setting material minimum and max depth, for on top rendering.
PE: Also support on top SpriteRender for crosshair...
PE: AudioSource - Stop() followed by Play() now act like a restart.
PE: Support importing WAV files that contain "extra format bytes" in "fmt" header.
PE: Fixed - Custom animation import, frame index start and end was not the actual index but depended on frames per second. Made it impossible to use, now use the real frame index.
PE: FBX Import - Improved normal map detection using diffuse name, if normap map was not setup inside object.
PE: DDS - Improve import time if source already has mipmaps and are compressed.

### DIFF
--- a/Source/Editor/Windows/Assets/MaterialWindow.cs
+++ b/Source/Editor/Windows/Assets/MaterialWindow.cs
@@ -128,6 +128,12 @@ namespace FlaxEditor.Windows.Assets
             [EditorOrder(440), DefaultValue(MaterialPostFxLocation.AfterPostProcessingPass), EditorDisplay("Misc"), Tooltip("The post fx material rendering location.")]
             public MaterialPostFxLocation PostFxLocation;
 
+            [EditorOrder(450), DefaultValue(0.0f), EditorDisplay("Misc"), Tooltip("Minimum depth used for rendering."), Limit(0.0f, 1.0f, 0.001f)]
+            public float MinDepth;
+
+            [EditorOrder(460), DefaultValue(1.0f), EditorDisplay("Misc"), Tooltip("Maximum depth used for rendering."), Limit(0.0f, 1.0f, 0.001f)]
+            public float MaxDepth;
+
             // Parameters
 
             [EditorOrder(1000), EditorDisplay("Parameters"), CustomEditor(typeof(ParametersEditor)), NoSerialize]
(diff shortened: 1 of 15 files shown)


mafiesto4:
Great improvements! Would be better to have them as separate PRs :)

One thing though, what would be min/max values for this material depth override to have it working for gun clipping? If I set `Max Depth` to `0` (or similar) then the weapon is properly rendered over nearby objects but it has the wrong linking applied.

Have you tried different technique to achieve it? Eg. https://www.youtube.com/watch?v=zqfzvHCcvZs&ab_channel=JackFrench

plemsoft:
If you look at the video fix, you see that it do the same "Scale Depth", so its the same. And yes I done it in other engines using other technique, but they are all more slow and require separate render pass (after or inside transparent).

Its still linear depth so what you do is moving the depth closer to the camera, so if you set it to min 0 and max 0 there would not be any depth to sort the drawing of the objects itself, so give it a little depth to work with like Min: 0.0 Max: 0.1,
this will also work for sprite rendering like if you want a place a crosshair and always rendering it on top.

It also work in reverse like if you want to place some skybox objects behind everything else (moving planets/stars...) you can set it to min: 0.9 max: 1.0 (or adjust it to go just above your current skybox depth).

It already change the materialinfo to version 11 and the header, so just after you merge perhaps add some more floats to the materialinfo like ObjectFOV or float filler_for_later[20] (without API_FIELD()), so these can be used later without changing material version/headers/upgrading code... :)

Then I can look at the ObjectFOV later, its kind of the same, we just need to change the projection matrix before rendering.

I have more commits, will make them into separate PR's in the future, but it would not be easy for me to separate these into separate PR's without remaking everything, so perhaps you could just merge this and reverse the commits you don't want ?

mafiesto4:
Okay, that explains more. Can you make a separate PR with materials depth min/max and other improvements? (I would like to merge the second one immediately)

plemsoft:
@mafiesto4 Not sure I understand , its all in this PR ? , are there any of the above commits you don't want ? , I can reverse them in this PR then ?

mafiesto4:
The material-related upgrade with depth min/max feature is something I'd like to separate from this PR - it's a little more complicated feature. The rest can stay so I would merge those changes without materials-stuff.

plemsoft:
@mafiesto4 First one is now here: https://github.com/FlaxEngine/FlaxEngine/pull/837

PR #836, opened 2022-11-26
Implement .NET 7 runtime support and bindings generation

GoaLitiuM:
Trello checklist in roadmap: https://trello.com/c/xgS3TphU/351-net-7

This is a mostly functional prototype of the engine running in .NET 7 runtime (Windows-only for now). The editor should be able to run the Flax Samples without any issues, and cooked builds should also be working. Not all features have been tested fully, so some errors can be expected.

The engine is still expected to be built with Mono header files included, as some of the Mono-related functions are called directly throughout the engine. The prototype wraps the Mono-functions and calls functionally equivalent methods in managed side using `nethost` library for hosting the .NET runtime in the same process. In the future, the engine should avoid calling Mono-functions directly and instead use the already established wrapper classes in `Engine/Scripting/ManagedCLR` to communicate with managed runtime.


GoaLitiuM:
Disabled the other Github actions for now and only running the Windows-related workflows.

Source/Tools/Flax.Build/Projects/VisualStudio/VCProjectGenerator.cs
```diff
@@ -175,9 +176,9 @@ public override void GenerateProject(Project project)
                 vcProjectFileContent.AppendLine("    <LibraryWPath />");
                 vcProjectFileContent.AppendLine("    <SourcePath />");
                 vcProjectFileContent.AppendLine("    <ExcludePath />");
-                vcProjectFileContent.AppendLine(string.Format("    <NMakeBuildCommandLine>{0} -build</NMakeBuildCommandLine>", cmdLine));
-                vcProjectFileContent.AppendLine(string.Format("    <NMakeReBuildCommandLine>{0} -rebuild</NMakeReBuildCommandLine>", cmdLine));
-                vcProjectFileContent.AppendLine(string.Format("    <NMakeCleanCommandLine>{0} -clean</NMakeCleanCommandLine>", cmdLine));
+                vcProjectFileContent.AppendLine(string.Format("    <NMakeBuildCommandLine>dotnet {0} -build</NMakeBuildCommandLine>", cmdLine));
```
mafiesto4, line 179: Why not run the build tool directly here?

Source/Engine/Core/Types/Variant.cpp
```diff
@@ -1578,8 +1590,10 @@ Variant::operator void*() const
     case VariantType::Blob:
         return AsBlob.Data;
     case VariantType::ManagedObject:
-#if USE_MONO
-        return AsUint ? mono_gchandle_get_target(AsUint) : nullptr;
+#if USE_NETCORE
+        return AsUint64 ? MUtils::GetGCHandleTarget(AsUint64) : nullptr;
```
mafiesto4, line 1594: Maybe we could just use `AsUint64` for GCHandleTarget storage for both Mono and NetCore versions? (simplify code)

Source/Engine/Main/Windows/main.cpp
```diff
@@ -50,14 +50,14 @@ int APIENTRY wWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPTSTR lpCmd
 #endif
 
     Platform::PreInit(hInstance);
-    __try
+    //__try
```
mafiesto4, line 53: NetCore catches C++ exceptions?

Source/Engine/Scripting/Types.h
```diff
@@ -49,6 +49,17 @@ typedef void MObject;
...
 
+#if USE_NETCORE
+struct _MonoDomain {};
+struct _MonoThread {};
+#endif
+
+#if USE_NETCORE
+typedef unsigned long long gchandle;
+#else
+typedef uint32 gchandle;
```
mafiesto4, line 60: Maybe we will simplfy gchandle to uint64 so Mono runtime will do casting but we will get simpler code in engine?

mafiesto4 (commented):
Great stuff!

Source/Tools/Flax.Build/Projects/VisualStudio/VCProjectGenerator.cs
GoaLitiuM: I think this was leftover code from the early versions where the build tool got built as a DLL-only, and you'd need to use `dotnet` to call DLL-binaries.

Source/Engine/Main/Windows/main.cpp
GoaLitiuM: More leftover code from debugging

mafiesto4:
Merged!  Works great! Thanks :) Now, time to work on quality and matching all the features. I'll take platforms support on myself.

Progress: https://trello.com/b/NQjLXRCP/flax-roadmap

PR #837, opened 2022-11-28
Fixes/changes and additions.

plemsoft:
PE: AudioSource - Stop() followed by Play() now act like a restart.
PE: Support importing WAV files that contain "extra format bytes" in "fmt" header.
PE: Fixed - Custom animation import, frame index start and end was not the actual index but depended on frames per second. Made it impossible to use, now use the real frame index.
PE: FBX Import - Improved normal map detection using diffuse name, if normap map was not setup inside object.
PE: DDS - Improve import time if source already has mipmaps and are compressed.

### DIFF
--- a/Source/Engine/Audio/AudioSource.cpp
+++ b/Source/Engine/Audio/AudioSource.cpp
@@ -132,7 +132,7 @@ void AudioSource::Play()
             Clip->RequestStreamingUpdate();
 
             // If we are looping and streaming also update streaming buffers
-            if (_loop)
+            if (_loop || state == States::Stopped)
                 RequestStreamingBuffersUpdate();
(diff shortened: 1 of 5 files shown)


plemsoft:
@mafiesto4 Not sure why the linux test fails:
/home/runner/work/FlaxEngine/FlaxEngine/Source/Engine/Tests/TestCollections.cpp:79: FAILED:
  CHECK( a1.Get(i) == v )
with expansion:
  false == true

Don't look related to any of the code changes ?

mafiesto4:
Thanks! That failing test was my bug and should be fixed by db6aab1cf6142be4bcfc872a99f381af9986508a

PR #839, opened 2022-11-29
On top rendering using MinDepth and MaxDepth.

plemsoft:
PE: csharp allow setting material minimum and max depth, for on top rendering.
PE: allow setting material minimum and max depth, for on top rendering.
PE: Also support on top SpriteRender for crosshair...
PE: Reserve some floats in materialinfo11 to allow new features in the future without changing header version and upgrading shaders.
PE: Static FOV, material will always have this static FOV even if camera change FOV.

Usage: Min Depth: 0.0 , Max Depth: 0.1
Usage: Static FOV: 60

@mafiesto4 Added Static FOV to this pull as it fit with the material/header upgrade, so it now fix both weapon clipping and weapon FOV issues.

@mafiesto4 Added distance bias to control draw order, can be used to set transparency sort priority. Added to this pull as this also require a material update :)

distancebias


mafiesto4:
Here is some work for customizable FPS weapon rendering: https://forum.flaxengine.com/t/first-person-view-model/892/2?u=mafiesto4

plemsoft:
@mafiesto4 Thanks :) , will check if I'm able to use it , I will close this pull as its not needed anymore.

PR #843, opened 2022-12-01
Core classes Random & RandomStream port to C#

Withaust:


Withaust:
@mafiesto4 I changed the base branch to `1.5`, since it seemed more fitting.

PR #850, opened 2022-12-09
Project independent saved windows layouts

Tryibion:
Hello, this PR allows the saved windows layouts to be accessed from any project. This also adds an apply and delete button so the user can remove them if they wish.


Tryibion:
Would it be a good idea to archive the layouts instead of deleting them if the user clicks delete? That way there could be a path to get it back.

PR #859, opened 2022-12-16
Fix issue of game window not returning focus to parent on mouse leave.

Tryibion:
Hello, this PR returns the focus to the parent of the game window when the mouse leaves it. This issue was causing the cursor not to change on the splitters if the game window was focused.


RuanLucasGD:
@Tryibion this PR fix this #1370 ?

PR #870, opened 2022-12-24
.NET7: Minor fixes

GoaLitiuM:
A fix for earlier commit regarding existing parameter names

### DIFF
--- a/Source/Engine/Scripting/DotNet/CoreCLR.cpp
+++ b/Source/Engine/Scripting/DotNet/CoreCLR.cpp
@@ -129,12 +129,7 @@ void CoreCLR::RegisterNativeLibrary(const char* moduleName, const char* modulePa
 
 void* CoreCLR::Allocate(int size)
 {
-#if PLATFORM_WINDOWS
-    void* ptr = CoTaskMemAlloc(size);
-#else
-    void* ptr = malloc(size);
-#endif
-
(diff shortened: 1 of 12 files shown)


Source/Engine/Scripting/DotNet/CoreCLR.cpp
```diff
@@ -129,12 +129,7 @@ void CoreCLR::RegisterNativeLibrary(const char* moduleName, const char* modulePa
...
 {
-#if PLATFORM_WINDOWS
-    void* ptr = CoTaskMemAlloc(size);
-#else
-    void* ptr = malloc(size);
-#endif
-
+    void* ptr = Platform::Allocate(size, 16);
 #if COMPILE_WITH_PROFILER
     Platform::OnMemoryAlloc(ptr, size);
```
mafiesto4, line 134: You can remove `OnMemoryAlloc`/`OnMemoryFree` calls when using `Platform::Allocate`/`Free` (it already calls profiler info)

mafiesto4 (changes requested):

GoaLitiuM:
Fixed. Would be probably a good idea to use `nameof(parameterName)` instead of `"parameterName"` in `MarshalUsing` attributes to catch errors earlier when invalid parameter names are generated.

PR #873, opened 2022-12-27
Serialize C++ includes cache results

GoaLitiuM:
Almost half the build tool runtime is spent scanning includes in C++-files, now the results are cached and invalidated when files last write timestamp changes.


mafiesto4:
Very good improvement!

I tested the performance of `IncludesCache.FindAllIncludedFiles` when building Flax with no file changed and with valid cache:
Before `470.653 ms`
After: `5.6776 ms`

This results in a significant boost for incremental builds.

PR #880, opened 2022-12-29
First property name letter to upper case fix

Tryibion:
Hello, this makes the first letter to always be capitalized in the UI for properties.


Tryibion:
Not sure why it failed on Linux...

PR #882, opened 2022-12-30
.NET7: Implement ManagedHandle replacement for GCHandles

GoaLitiuM:
Allows for more controlled management of weak handles, increasing their lifetime for slightly longer than with GCHandles.


GoaLitiuM:
Also fixed some issue with creating delegates to scripting assemblies after script reload. Some runtime methods stores the internally created delegate types into one dynamic assembly that persists throughout the script reloads, and causes some issues when the delegate for same method is created again.

Issue #832, opened 2022-11-24
Casting result of binary AND operator for enums.


mtszkarbowiak:
https://github.com/FlaxEngine/FlaxEngine/blob/76c1045444135c841d097fb69e9effa3f38c0d50/Source/Engine/Core/Types/BaseTypes.h#L146

Binary AND operator is the only operation, which has no casting to type T.

mafiesto4:
Fixed in f127bbebe1638302d5bd0761d4922966dc12e94f

PR #886, opened 2023-01-05
Fix for issue #832

Tryibion:
I don't really know if this helps anything besides making it consistent... I think I got them all...


Tryibion:
I did see that you were wanting to use `__underlyijg_type` if you can get that to work it may be a better solution then this brute force method.

mafiesto4:
@Tryibion yeah I think I'm gonna do some major cleanup to those enums and use utilities like `EnumHasAllFlags`/`EnumHasAnyFlags` to simplify code

mafiesto4:
Done in f127bbebe1638302d5bd0761d4922966dc12e94f

Issue #825, opened 2022-11-12
Visual Studio / Visual Studio Code Auto Complete not working with Flax Master build


RuanLucasGD:
Hello, I'm having problems trying to use any code editor when I click **"Generate script project files"** in the editor.
IDEs autocomplete stops working. If I try to go back to version 1.4 and generate the project scripts again, the auto complete works again

![image](https://user-images.githubusercontent.com/79365912/201454520-c82ef254-5fe5-4cc2-bb13-a39ac71422fb.png)

Note: The code editors don't work properly but Flax compiles the scripts normally

mafiesto4:
I wonder if 84bd7da620c0a840402332c5c5e38f5885ef3f00 fixed it? Does it work fine in the latest `master` build?

RuanLucasGD:
@mafiesto4 still does not work

mafiesto4:
Fxed by #888

Issue #649, opened 2021-10-07
project/solution files unusable for VS code


nothingTVatYT:
In addition to https://github.com/FlaxEngine/FlaxEngine/issues/115 the sln as well as the the csproj files are set up in a way that breaks omnisharp.

**Issue description:**
Omnisharp in VS code complains about duplicate keys (s. duplicates.txt). Additionally a FlaxEngine.csproj referenced in FlaxEngine.Test.csproj is missing.
After fixing all the above VS code marks several errors like

'Script' is inaccessible due to its protection level [Game] csharp(CS0122)

**Flax version:**
1.2.6224

[duplicates.txt](https://github.com/FlaxEngine/FlaxEngine/files/7301853/duplicates.txt)

Issue #115, opened 2021-01-17
Visual Studio Code script projects lack code completion


GoaLitiuM:
Generated VSCode projects does not have the required information for C# Omnisharp extension to enable Intellisense/code completion support. Omnisharp uses VS solution and project files for Intellisense, so these files needs to be generated for VSCode projects as well.

Workaround for this issue is to generate VS2019 project files first followed by generating VSCode project files, then Omnisharp active project (`Omnisharp: Select Project` in Command Palette) must be changed to the generated solution file, after that Intellisense starts working properly for the game project.

GoaLitiuM:
Intellisense is only partially working, the correct `Script` class for example can't be found, showing the following errors:
![image](https://user-images.githubusercontent.com/4064397/109684135-a6415400-7b88-11eb-8c2e-4b41d162b1d0.png)

mafiesto4:
In `1.1` it works fine in pre-built engine. The problem persist when using source-build engine due to using generated bindings code (`*.Gen.cs` files eg. from `Cache/Intermediate/FlaxEditor/Windows/x64/Debug/Scripting/Scripting.Gen.cs`).

PR #888, opened 2023-01-08
Fix Intellisense issues with Visual Studio Code

GoaLitiuM:
The C#-plugin in VSCode does not use any of the intended build configurations by default, which causes errors about missing generated bindings code. The solution to this is to generated `omnisharp.json`-file with default build configuration set to `Editor.Debug` so the Intellisense can find the correct definitions in the generated bindings files.

Includes a minor fix to not overwrite generated VSCode files, as these can contain project specific configuration settings or other custom tasks user can define for the project.


GoaLitiuM:
Possible fix for #825, #649, #115

GoaLitiuM:
I dropped the commit which prevents overwriting the VSCode project files. The project files would need to be rewritten whenever something changes in the environment (compiler version, FlaxEditor location), while preserving any user made changes.

RuanLucasGD:
@GoaLitiuM Compiling the editor as Editor.Debug and adding the config file to the project folder worked for me.

@mafiesto4 
Fix confirmed for #825

PR #892, opened 2023-01-14
Allow unsubscribing from managed events multiple times

GoaLitiuM:


GoaLitiuM:
This also fixes removal of already unsubscribed events decreasing the internal counter under zero

PR #899, opened 2023-01-19
Adding Some Methods

RuanLucasGD:
Hello, in this PR i added some methods:

- Vector2.Normalized
- Float2.Normalized
- Gamepad.IsAnyKeyDown
- Mouse.IsAnyKeyDown

### DIFF
--- a/Source/Engine/Core/Math/Vector2.cs
+++ b/Source/Engine/Core/Math/Vector2.cs
@@ -302,6 +302,19 @@ namespace FlaxEngine
         }
 
         /// <summary>
+        /// Gets the normalized vector. Returned vector has length equal 1.
+        /// </summary>
+        public Vector2 Normalized
+        {
+            get
+            {
+                Vector2 result = this;
+                result.Normalize();
+                return result;
+            }
+        }
+
+        /// <summary>
         /// Creates an array containing the elements of the vector.
         /// </summary>
         public Real[] ToArray()
--- a/Source/Engine/Core/Math/Float2.cs
+++ b/Source/Engine/Core/Math/Float2.cs
@@ -268,6 +268,19 @@ namespace FlaxEngine
         public float LengthSquared => X * X + Y * Y;
 
         /// <summary>
+        /// Gets the normalized vector. Returned vector has length equal 1.
+        /// </summary>
+        public Float2 Normalized
+        {
+            get
+            {
+                Float2 result = this;
+                result.Normalize();
+                return result;
+            }
+        }
+
+        /// <summary>
         /// Converts the vector into a unit vector.
         /// </summary>
         public void Normalize()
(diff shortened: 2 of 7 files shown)


Source/Engine/Core/Math/Vector2.cs
```diff
@@ -301,6 +301,19 @@ public void Normalize()
...
+        /// <summary>
+        /// Gets the normalized vector. Returned vector has length equal 1.
+        /// </summary>
+        public Vector2 Normalized
+        {
+            get
+            {
+                Vector2 result = this;
+                result.Normalize();
+                return result;
```
Tryibion, line 313: could this just be simplified to `return this.Normalize()`?

Tryibion:
Is there already a function for Keyboard any key down? could that be easy to add as well if it doesnt exist?

RuanLucasGD:
@Tryibion This function already exists, but the ones I added didn't exist

Source/Engine/Core/Math/Vector2.cs
RuanLucasGD: That would simplify things, thanks I'll adjust

RuanLucasGD: No,` Normalize()` returns a void

Tryibion: Ah, okay didn't know that. Makes sense then. Could something like this also be useful for vector3?

RuanLucasGD: It already exists, but I found it strange that the same function did not exist in Vector2 so I created it. I just copied the same function from Vector3 to Vector2

mafiesto4:
I think it would be better to name new methods `IsAnyButtonDown` for Gamepad and Mouse.

RuanLucasGD:
@mafiesto4 IsAnyKeyDown exists in the Keyboard class but does not exist in the Mouse and Gamepad class. What I did was create a copy of what was on the keyboard for the gamepad/mouse.

So I think that if you want to change the mouse/gamepad IsAnyKeyDown, you should also change the keyboard

https://github.com/FlaxEngine/FlaxEngine/blob/997baa3b774ac4dea739e15a55070519b0863afc/Source/Engine/Input/Keyboard.h#L70

RuanLucasGD:
@mafiesto4 What do you think of IsAnyKeyDown for keyboard and IsAnyButtonDown for gamepad/mouse

PrecisionRender:
> @mafiesto4 What do you think of IsAnyKeyDown for keyboard and IsAnyButtonDown for gamepad/mouse?

Personally, I think this is the best way to go.

RuanLucasGD:
@mafiesto4 done

Issue #805, opened 2022-10-25
Editor: Add arrow to indicate Actor components collapsed/opened state


brmassa:
In the inspector, there is no visual cue to indicate that a certain category (General, Transform, Script, Control, etc) is collapsed or just empty. Suggest to add the same arrow used on drop combobox.

#QoL

![image](https://user-images.githubusercontent.com/13696218/197891554-956e1fcc-6c4b-4a92-b329-4389da2e702f.png)

mafiesto4:
Implemented in #903

PR #903, opened 2023-01-21
Various UI improvements around drop panels in the editor

Tryibion:
Hello this adds some improvements to drop panels in the editor to make them for user friendly. Increase header size, add containment area, and adding arrows. This allows for the user to more clearly see which drop panel they are working under. This stemmed from #805.

properties window

graphics window


brmassa:
Great! I've pointed this out in this issue: https://github.com/FlaxEngine/FlaxEngine/issues/805

PR #911, opened 2023-01-28
Fix process creation on Linux

GoaLitiuM:
Forked child process did not exit properly, sometimes causing scripting compilation to get stuck and never finish.


GoaLitiuM:
Didn't seem to fix the issue, the parent process gets stuck reading the pipe in line with `while (fgets(lineBuffer, sizeof(lineBuffer), stdPipe) != NULL)`.

mafiesto4:
Ah I noticed this problem too!

GoaLitiuM:
Not sure if the latest changes fixes this, it's hard to reproduce.

PR #918, opened 2023-01-29
Fix compiler warnings with Clang 15 or later

GoaLitiuM:


GoaLitiuM:
I'm getting a lot of deprecation warnings when compiling the engine with Clang 16, though these deprecation warnings started in version 15.

PR #925, opened 2023-02-03
Changed `GetGameWindowSize` to return viewport size.

Tryibion:
This allows for certain functions to behave correctly. There was an issue that i noticed with the `Camera.ConvertMouseToRay()` function because it was going off of the game window size and not the viewport size. It was behaving incorrectly if the user has a custom viewport size.


mafiesto4:
Maybe `GetGameWindowSize` should simply return the actual viewport size instead of the game window size? So it should handle the custom aspect ratio or resolution usage too?

Tryibion:
That is a possibility. I just created a whole new function for it in case anywhere it was being used needed the actually game window. It would be a pretty easy switch.

Tryibion:
@mafiesto4 changed it.

Issue #50, opened 2020-12-24
Potential Visject Improvements


stefnotch:
### PR https://github.com/FlaxEngine/FlaxAPI/pull/21

- Arrow Keys & Enter to choose an autocomplete suggestion
- Alternative names for nodes (`*` for multiply, `+` for add, ...)
- Visject Surface keeps it's focus on node creation
- ~~Smart:tm: node connecting~~
- ~~Newly created nodes get added to the selection~~
- ~~Node gets removed from selection when you add some connections~~

### PR https://github.com/FlaxEngine/FlaxAPI/pull/25

- When creating a connection, you can right click to create a new node and it will get connected to that one.
- Removed some features added in FlaxEngine/FlaxAPI#21 because [they weren't intuitive](https://github.com/FlaxEngine/FlaxAPI/issues/22#issuecomment-405356097)

### PR https://github.com/FlaxEngine/FlaxAPI/pull/29

- `3` → should suggest `float` and `int`
- `true` → should suggest `bool`

### PR https://github.com/FlaxEngine/FlaxAPI/pull/39

- Visject Context Menu sorting (e.g. currently selected box is a blue texture box --> type `tex` --> first suggestion is `Sample Texture`
- Paste() now pastes at the position of the cursor
- Visject CM pops up on mouse release when creating a new connection
- Visject CM default position is now offset by 5 pixels. This offset was added in case the user doesn't like the CM and wants to quickly get rid of it by clicking.

### PR https://github.com/FlaxEngine/FlaxAPI/pull/78

- Type to create node at cursor (no right clicking required)
- Boxes can be selected
- Typing after selecting a box or output-only node results in the primary menu popping up
- Newly created nodes are selected by default
- Current box advances after creating a new node
- Type `//` to create a comment above a node    
- Brackets!

### PR https://github.com/FlaxEngine/FlaxAPI/pull/83

- `if` as alternative title for the branch node

### PR https://github.com/FlaxEngine/FlaxAPI/pull/91

- Highlight connections of currently selected box
- Prioritize context menu matches that start with the search text

### TODO:

- `.xyz` → `Mask XYZ`
- `.zyy` → `Unpack Vector3` + `Pack Vector3`
- Select *Unpack Material Layer* Node &rarr; Type `.color` &rarr; selects the color output
  - Add check to make sure that you can't create outputs that start with a number
  - Or maybe use the Visject context menu for this?
- Node creation dodging (no creating nodes on top of another one!)
[...]

PR #934, opened 2023-02-09
VisjectSurface Shortcuts

Swiggies:
Adds the ability to have shortcuts for any VisjectSurfaces, with special attention to the Material surface. Functions very similarly to other material graphs in similar software, you can hold down a keyboard key and on mouse up, it will spawn the relevant node. To get this functionality it unfortunately meant removing the functionality of simply typing to bring up the primary menu, but instead you can simply press the spacebar to bring the menu up. With a node selected and pressing spacebar you can also automatically spawn a node that will try and connect to the selected node.

While this does remove functionality I do believe it will bring a more comfortable experience for users while also providing them with a familiar one if they are used to other similar graph systems.

1-5 will spawn a float1-4 and then color node
A = Add
S = Subtract
D = Divide
L = Lerp
E = Power
G = Get Parameter

On the material graph you can use
T for a Texture Sampler
U for a TexCoord node

To add similar functionality to other VisjectSurfaces you can simply override the GetShortcutNode() function and add the shortcuts as demonstrated in MaterialSurface.cs

### DIFF
--- a/Source/Editor/Surface/MaterialSurface.cs
+++ b/Source/Editor/Surface/MaterialSurface.cs
@@ -110,54 +110,11 @@ namespace FlaxEditor.Surface
         }
 
         /// <inheritdoc/>
-        public override bool OnMouseDown(Float2 location, MouseButton button)
+        protected override void GetShortcutNode(KeyboardKeys key, out ushort archetype, out ushort node)
         {
-            if (button != MouseButton.Left) return base.OnMouseDown(location, button);
-
-            GetNode(_heldKey, out var archetype, out var node);
-
-            if (archetype == 0 || node == 0) return base.OnMouseDown(location, button);
-
-            var spawnedNode = Context.SpawnNode(archetype, node, _rootControl.PointFromParent(location));
-            return base.OnMouseDown(location, button);
-        }
-
-        private void GetNode(KeyboardKeys key, out ushort archetype, out ushort node)
-        {
-            archetype = 0;
-            node = 0;
+            base.GetShortcutNode(key, out archetype, out node);
             switch (key)
             {
-                // Add node
-                case KeyboardKeys.A:
-                    archetype = 3;
-                    node = 1;
-                    break;
-                // Subtract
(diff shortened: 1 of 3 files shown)


stefnotch:
As the one who added the "simply typing to bring up the menu" feature, I'm not sure if shortcuts that don't use a modifier key like `Ctrl` or `Alt` are a good idea.

Ideally, shortcuts should be
- easy to remember
- useful, Visject has material graphs and normal scripts https://docs.flaxengine.com/manual/scripting/visual/index.html

I think improving the "typing to bring up the menu" feature could give you something even more powerful than a simplistic shortcut system. Like if certain single-character inputs suggested extremely common nodes, then that might make the "typing" approach closer to a shortcut system.

(See also #50 )

Swiggies:
I agree with all your sentiments and that's why all the functionality is still there. It just requires the press of `Spacebar` instead of pressing any key to bring up the primary menu. This PR just brings it more inline with other similar systems like Blender or Unreal, which both use `Spacebar` to bring up the primary menu, and allow the use of shortcuts by holding down a key and clicking to spawn a node.

I've also made sure the keybinds are similar to those programs as well to make it easier for users who might be comfortable with those shortcuts to migrate over to Flax. Personally, it also speeds things up quite a lot for me when I don't have to take my hand off the mouse as much to type out the names.

stefnotch:
Thanks for explaining your thought process! That does make sense to me.

One thing that would make this shortcut system even more intuitive (IMO) would be having a little icon or similar 
- that when you hover over it, shows the available shortcuts (discoverability)
- shows the currently pressed shortcut (thus also making it obvious when you're about to do a certain thing)

That can, of course, happen in a future pull request. If you do think that's a good idea, and can't do it in this PR, then please do open an issue for it, so that nobody forgets about it. :)

stefnotch:
Since this is a pretty big change from how the Visject shortcuts system works, it might be worthwhile to get other people's feedback or ideas.

PR #936, opened 2023-02-11
Context menu and tooltip fixes

### LINKED ISSUE #668: Menu closes when a submenu closes
**Issue description:**
Every time you hover over a menu item that brings up a submenu and leave that item not only does the submenu close but the parent as well. This happens in the Linux editor and makes it almost impossible to get anything done.

**Video showing the issue:**
https://discord.com/channels/437989205315158016/538664104856911883/901217501948555416

**Flax version:**
1.2.6224 up to at least current master Nov. 11th 2021

GoaLitiuM:
Tracking the window focus with `IsForegroundWindow` is somewhat inconsistent on Linux, so instead of relying on it, use the input focus to track when mouse is clicked outside the context menus. 

Fixes #668

### DIFF
--- a/Source/Editor/GUI/ContextMenu/ContextMenuBase.cs
+++ b/Source/Editor/GUI/ContextMenu/ContextMenuBase.cs
@@ -196,7 +196,6 @@ namespace FlaxEditor.GUI.ContextMenu
             desc.HasSizingFrame = false;
             OnWindowCreating(ref desc);
             _window = Platform.CreateWindow(ref desc);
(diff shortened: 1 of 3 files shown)


GoaLitiuM:
Added another semi-related fix to an issue that occurs on Linux, the tooltips wont close properly when interacting with the scene tree.

mafiesto4:
Great change! Now, context menus and tooltips work far better on Linux. I preserved old logic on Windows for now 05cb02aa0987e901b4bf5b1758bf6a051839b6d7.

PR #943, opened 2023-02-17
.NET7: Improve Linux dotnet runtime identifier detection

GoaLitiuM:
The RID detection should work in all possible distros now (including Ubuntu 22.x using the Microsoft package feed)

### DIFF
--- a/Source/Tools/Flax.Build/Build/DotNet/DotNetSdk.cs
+++ b/Source/Tools/Flax.Build/Build/DotNet/DotNetSdk.cs
@@ -103,9 +103,7 @@ namespace Flax.Build
                 // TODO: Support /etc/dotnet/install_location
 
                 // Detect custom RID in some distros
-                string osId = File.ReadAllLines("/etc/os-release").FirstOrDefault(x => x.StartsWith("ID="), "ID=linux").Substring("ID=".Length);
-
-                rid = $"{osId}-{arch}";
+                rid = Utilities.ReadProcessOutput("dotnet", "--info").Split('\n').FirstOrDefault(x => x.StartsWith(" RID:"), "").Replace("RID:", "").Trim();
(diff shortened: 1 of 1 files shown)


brmassa:
@GoaLitiuM, I forgot about this patch and I've created the issue #956 .

I'm suggesting using `System.Runtime.InteropServices.RuntimeInformation` and `System.Runtime.InteropServices.RuntimeEnvironment` classes because they already provide the necessary information (path, runtime path, arch, os, etc) regardless of the OS/architecture.

It would be interesting to rewrite the whole function using only these classes.

for the issue mentioned (not compiling in Ubuntu), I used:
```csharp
rid = System.Runtime.InteropServices.RuntimeInformation.RuntimeIdentifier.ToString();
```
but as I mentioned, I believe it can be used for all OSes.

GoaLitiuM:
> I'm suggesting using `System.Runtime.InteropServices.RuntimeInformation` and `System.Runtime.InteropServices.RuntimeEnvironment` classes because they already provide the necessary information (path, runtime path, arch, os, etc) regardless of the OS/architecture.

I never though about using the information available in runtime. This should simplify the detection a bit, though on Windows `RuntimeIdentifier` returns "win10" for me instead of the expected "win". I need to double check how this behaves in Arch.

brmassa:
Oh, yes, we have to check in all OSes. Mac

Also, due to different internal organizations on each distro, I would suggest for we have a small list of distros to test.

* Debian/Ubuntu/Mint
* Arch/Majaro
* Fedora/RedHat/SUSE
* Mac

It will cover 99% distro scenarios, IFAIK

I don't know if there is a minimum version of Windows required.

Finally, since we are moving to .NET directly, we could clean the code a lot by simply ignoring NetFramework-only stuff.

mafiesto4:
Merged.

PR #944, opened 2023-02-17
Improve C++ standard library version detection

GoaLitiuM:
Clang seems to default to the same version of the stdlibc++ which is shipped with the installed GCC. The verbose output of Clang shows the detected version of the library, so we can parse the version from there.


GoaLitiuM:
Ubuntu 20.04, no GCC installed: `LibStdCppVersion: 9, Clang: 10.0.0`
Ubuntu 22.10, GCC 11.3.0: `LibStdCppVersion: 11, Clang: 14.0.0`
Arch, GCC 12.2.1: `LibStdCppVersion: 12.2.1, Clang: 15.0.7`

PR #955, opened 2023-02-23
Added Main Menu short cuts to update with input option changes

Tryibion:
Hello, This PR adds the ability for the main menu actions short cut text to update when the user saves the short cut in the input options.


Tryibion:
I couldn't think of another way to make sure that stuff added in plugin's gets re-added besides reinitializing the editor plugins. If anyone can think of a better way, I will try it out.

mafiesto4:
What if we just use cached button objects such as `_menuEditUndo` and update their tooltip only without whole recreation?

Tryibion:
Will redo and re-submit

PR #963, opened 2023-03-05
implement openal hrtf audio support

envision3d:
- upgrade openal version
- add support to create openal context with hrtf enabled
- add option in audio settings for hrtf enablement
- add static property on Audio class for runtime toggling of hrtf
- fix bug with left/right and front/back being reversed


envision3d:
A demo project showcasing HRTF audio can be found here:

https://1drv.ms/u/s!AuZzVIiMzVLWlNNRds_Ua1IeAdvWGw?e=yn186H

mafiesto4:
Thanks for this addition!

PR #969, opened 2023-03-14
Various Flax.Build performance improvements

GoaLitiuM:
Few notable performance improvements to Flax.Build build process:

- Cache `File.Exists` and `File.GetLastWriteTime` queries in `BindingsGenerator` and `TaskGraph`, these are often called one after other so caching the file handle saves a lot here. Also same files are queried when loading caches and saving caches so it makes sense to cache the results during cache load as well.
- Avoid running the network code generation when the code is already generated.
- Use dictionary for file index lookups in ``TaskGraph.SaveCache``.
- Use new `System.Text.Json` API for Flax.Build serialization/deserialization, this seems to be over over two times faster than `Newtonsoft.Json` when loading the project files. Also utilizes source generator for slightly improved serialization times.

All these changes gave me over 30% reduction in Flax.Build calling times.

### DIFF
--- a/Source/Tools/Flax.Build/Bindings/BindingsGenerator.Cache.cs
+++ b/Source/Tools/Flax.Build/Bindings/BindingsGenerator.Cache.cs
@@ -253,7 +253,7 @@ namespace Flax.Build.Bindings
             {
                 // Version
                 writer.Write(CacheVersion);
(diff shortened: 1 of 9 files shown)


mafiesto4:
I've got 30% perf boost too on noop build with SSD drive (0.78s -> 0.54s). On HDD drive it's even higher perf boost :)

PR #978, opened 2023-04-01
Remove extra sleep when Editor is not focused

GoaLitiuM:
The engine is already sleeping the extra time between frames


mafiesto4:
Yeah Engine already has sleeping functionality when not focused. I tested and it looks like CPU % usage goes down when unfocused just like before so that additional sleep in Editor code is not needed anymore.

PR #983, opened 2023-04-08
Visject cm Visual changes

Tryibion:
Hello, this PR makes a few visual changes to the Visject CM to make it (in my opinion) look a little cleaner.

image


mafiesto4:
I did minor improvement here 01ced1cacdcd508882179eece134eaa027392191

PR #987, opened 2023-04-09
Add LateFixedUpdate event for scripts

GoaLitiuM:
Since we don't have configurable script event system or Unity's customizable script execution order system to achieve something similar, I'll propose adding `LateFixedUpdate` event for scripts which is called after all `FixedUpdate` events and physics simulation is finished.

I need to run fixed updates for game objects first and at the end of the fixed update frame, gather the results of the physics simulation and send them over network to clients. I can't do this after individual `FixedUpdate` events because objects might interact with each other in unexpected order and change the state after scripts respective `FixedUpdate` event.


mafiesto4:
Cool! We need an additional `_tickLateFixedUpdate` flag in `Script` that would be auto-set for  C#/VS scripts in `Script::SetupType` (depending on the method override existence).

PR #1025, opened 2023-04-21
Add saving and re-opening all active scenes between editor sessions.

Tryibion:
This allows the editor to save all of the open scenes on editor close and allows the editor to re-open all of the scenes on init.


mafiesto4:
Could you change that `List<Guid>`  to `Guid[]` so it's a little more optimized?

Tryibion:
> Could you change that `List<Guid>` to `Guid[]` so it's a little bit more optimized?

I have changed the `List<Guid>` to `Guid[]` when deserializing from the project cache.

PR #1027, opened 2023-04-21
Add batch creating prefabs from multiple selected actors in scene tree.

Tryibion:
Useful for adding many prefab actors at once when dragging from the scene. This does not allow renaming of each prefab on creating.


Source/Editor/Modules/PrefabsModule.cs
```diff
@@ -37,6 +37,8 @@ public sealed class PrefabsModule : EditorModule
         /// </summary>
         public event Action<Prefab, Actor> PrefabApplied;
 
+        private Actor _actor;
```
mafiesto4, line 40: Could you add a comment that it's locally cached actor ref for prefab creation? And maybe rename it to something more self-explanatory.

Source/Editor/Content/GUI/ContentView.DragDrop.cs
```diff
@@ -46,7 +46,14 @@ private bool ValidateDragActors(ActorNode actor)
         private void ImportActors(DragActors actors, ContentFolder location)
         {
             // Use only the first actor
```
mafiesto4, line 48: This comment can be removed as it's obsolete after this change

Source/Editor/Modules/PrefabsModule.cs
```diff
@@ -107,29 +120,35 @@ private void OnPrefabCreated(ContentItem contentItem)
...
                 }
+                Undo.AddAction(new MultiUndoAction(actions));
+
+                _actor = null;
             }
 
             Editor.Instance.Windows.PropertiesWin.Presenter.BuildLayout();
         }
+        
+        private void GetActorsTree(List<Actor> list, Actor a)
```
mafiesto4, line 143: Could you move this function to `FlaxEditor.Utilities.Utils` as static. Then use it here, in `ActorStaticFlagsEditor` and `ActorLayerEditor` so we reuse more code.

mafiesto4 (changes requested):

Tryibion:
Updated to #1261

PR #1034, opened 2023-04-22
Add "GetSubTags" to "Tags"

RuanLucasGD:


Source/Engine/Level/Tags.cpp
```diff
@@ -55,6 +55,24 @@ Tag Tags::Get(const StringView& tagName)
     return tag;
 }
 
+Array<Tag> Tags::GetSubTags(Tag parentTag)
+{
+    Array<Tag> subTags = Array<Tag>();
```
Withaust, line 60: Why not just use `Array<Tag> subTags;` syntax?

RuanLucasGD: @Withaust fixed. Thanks

Withaust (approved):

Source/Engine/Level/Tags.cpp
```diff
@@ -55,6 +55,23 @@ Tag Tags::Get(const StringView& tagName)
     return tag;
 }
 
+Array<Tag> Tags::GetSubTags(Tag parentTag)
+{
+    Array<Tag> subTags;
+    auto _parentTagName = parentTag.ToString();
```
mafiesto4, line 61: `const String&` instead of `auto` to reduce memory allocations

Source/Engine/Level/Tags.cpp
```diff
@@ -55,6 +55,23 @@ Tag Tags::Get(const StringView& tagName)
...
+
+    for (int i = 0; i < Tags::List.Count(); i++)
+    {
+        auto& subTagName = Tags::List[i];
+        if (subTagName.Contains(_parentTagName) && subTagName != _parentTagName)
+        {
+            subTags.Add(Tags::Get(subTagName));
```
mafiesto4, line 68: `Tag(i + 1)`instead of `Tags::Get(subTagName)` as it's already in that tags list

Source/Engine/Level/Tags.cpp
```diff
@@ -55,6 +55,23 @@ Tag Tags::Get(const StringView& tagName)
...
 
+Array<Tag> Tags::GetSubTags(Tag parentTag)
+{
+    Array<Tag> subTags;
+    auto _parentTagName = parentTag.ToString();
+
+    for (int i = 0; i < Tags::List.Count(); i++)
+    {
+        auto& subTagName = Tags::List[i];
+        if (subTagName.Contains(_parentTagName) && subTagName != _parentTagName)
```
mafiesto4, line 66: It should be `subTagName.StartsWith()` as tag can contain other tags inside and not be a parent. Also, instead of comparing strings you can compare current tag index (`i+1`) with `parentTag.Index` for faster perf.

mafiesto4 (changes requested):
Nice feature but I left some comments how to make it work faster.

mafiesto4 (approved):

PR #1035, opened 2023-04-22
Level::ReversedUnload implementation

Withaust:
Its disabled by default, so no compatibility breaks should arise from this change.


mafiesto4:
Actually, maybe we just unload the scenes in inversed order by default without a need for `ReversedUnload`? I think it's a far better behavior.

Withaust:
@mafiesto4 Done

PR #1039, opened 2023-04-23
Remove fake lighting from DebugDraw shader

Withaust:
I removed fake lighting from the DebugDraw shader due to some elements having poor visibility.
Here is how it looks like right now `WITH` fake lighting:

image

And here is how it now looks `WITHOUT` fake lighting:

image


mafiesto4:
This fake lighting helps to debug more complex things like large terrain collision debug preview:

image

Otherwise, it looks very flat:

image

Withaust:
@mafiesto4 What about having a separate implementation that would allow toggling this behavior, like with the depth test? Would that be suitable?

mafiesto4:
Maybe small adjustments to that lighting could help or separate shader for wireframe meshes.

Withaust:
Superseded by https://github.com/FlaxEngine/FlaxEngine/pull/1076

Withaust:
@mafiesto4 Implemented with a separate PS in https://github.com/FlaxEngine/FlaxEngine/pull/1077

PR #1041, opened 2023-04-23
`ViewportIconsRenderer::AddActorWithTexture` addition

Withaust:
This adds an ability to register actors for display with viewport icons without being bound by types.


Withaust:
There seems to be a problem at rendering multiple custom icons due to having only one InstanceBuffer slot for custom textures

Withaust:
Okay, so basically I had to implement a system of dynamic material creation in order for multiple custom icons to function properly. Original code was using the same `InstanceBuffer` for all custom icons, and kept reassigning image parameter with a new one on every draw, but that was not working at all, and if you had more then 1 custom icon at display, it would draw them all with the same texture that was assigned the latest. I solved this by keeping track of virtually created instances of a default material per custom texture.

PR #1046, opened 2023-04-24
Fix collision assets creation in non asset folders.

Tryibion:
If a collision asset is trying to be created in a folder that can not have assets, the folder will switch to the folder with the model in it.

### DIFF
--- a/Source/Editor/Content/Proxy/CollisionDataProxy.cs
+++ b/Source/Editor/Content/Proxy/CollisionDataProxy.cs
@@ -141,7 +141,13 @@ namespace FlaxEditor.Content
                 });
             };
             var initialName = (modelItem?.ShortName ?? Path.GetFileNameWithoutExtension(model.Path)) + " Collision";
-            Editor.Instance.Windows.ContentWin.NewItem(this, null, create, initialName, withRenaming);
+            
+            // If folder can not have assets then move to folder with the model and create asset
+            var contentWin = Editor.Instance.Windows.ContentWin;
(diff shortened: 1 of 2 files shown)


Source/Editor/Content/Proxy/CollisionDataProxy.cs
```diff
@@ -141,7 +141,13 @@ public void CreateCollisionDataFromModel(Model model, Action<CollisionData> crea
                 });
             };
             var initialName = (modelItem?.ShortName ?? Path.GetFileNameWithoutExtension(model.Path)) + " Collision";
-            Editor.Instance.Windows.ContentWin.NewItem(this, null, create, initialName, withRenaming);
+            
+            // If folder can not have assets then move to folder with the model and create asset
+            var contentWin = Editor.Instance.Windows.ContentWin;
+            if (!contentWin.CurrentViewFolder.CanHaveAssets)
+                contentWin.Navigate(modelItem?.ParentFolder?.Node);
```
mafiesto4, line 148: I think `ContentWindow.NewItem` could call `proxy.CanCreate` over that folder to check if it's a valid location, and then attempt to change it to Content so we have a more generic solution.

mafiesto4 (changes requested):

Source/Editor/Content/Proxy/CollisionDataProxy.cs
Tryibion: Changed to now handling in `ContentWindow.NewItem`

mafiesto4 (approved):

PR #1058, opened 2023-04-29
`NetworkManager::RPCInvoker` implementation

Withaust:
This implements an ability to monitor who the latest RPC invoker was inside of the RPC method body and write a logic accordingly, like this:
```
API_FUNCTION(NetworkRpc = "Server, Reliable") void SendMessage(const String& text)
{
    NETWORK_RPC_IMPL(Chat, SendMessage, text);
    RecieveMessage(NetworkManager::RPCInvoker, text);
}

API_FUNCTION(NetworkRpc = "Client, Reliable") void RecieveMessage(uint32 sender, const String& text)
{
    NETWORK_RPC_IMPL(Chat, RecieveMessage, sender, text);
    
    if (sender == NetworkManager::LocalClientId)
    {
        LOG(Info, "Called RecieveMessage locally with text : {0}", text);
    }
    else
    {
        LOG(Info, "Got RecieveMessage remote call with text: {0}", text);
    }
}
```
This snippet of code demonstrates potential chat implementation, which allows for the server to refire the incoming chat RPC with the id of the person that sent this message for other clients, so that they could run custom formatting, like `%username%: %chat message%` for example.

### DIFF
--- a/Source/Engine/Networking/NetworkManager.cpp
+++ b/Source/Engine/Networking/NetworkManager.cpp
@@ -25,6 +25,7 @@ uint32 NetworkManager::Frame = 0;
 uint32 NetworkManager::LocalClientId = 0;
 NetworkClient* NetworkManager::LocalClient = nullptr;
 Array<NetworkClient*> NetworkManager::Clients;
+uint32 NetworkManager::RPCInvoker = 0;
 Action NetworkManager::StateChanged;
 Delegate<NetworkClientConnectionData&> NetworkManager::ClientConnecting;
 Delegate<NetworkClient*> NetworkManager::ClientConnected;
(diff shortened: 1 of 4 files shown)


Withaust:
Already implemented but undocumented by https://github.com/FlaxEngine/FlaxEngine/commit/e4804db160c2783266a241cd7bf0b8e4e9813e22

mafiesto4:
Here are docs for that new `NetworkRpcParams` feature: https://github.com/FlaxEngine/FlaxDocs/blob/master/manual/networking/high-level.md#rpc-context-parameters

It allows for both sending RPC to specific clients (from server) or getting info who sends that data - works also in network serialization functions from the stream. Potentially in future, we want to run objects replication in async (eg. via Job System) thus storing network message-related data within `NetworkStream` instance will work fine.

Issue #713, opened 2022-05-12
Suggestion: Add exist-check to operator ==


mlavik1:
**Issue description:**
Using the `==` operator to do nullchecks on actors that have been destroyed may currently produce confusing results.
Example:
```csharp
Actor myActor = otherActor;
Actor.Destroy(otherActor);
// ... some frames later
if (myActor != null)
   Debug.Log(myActor.Name); // This will throw exceptions
```
Since the managed `Actor` instance is still valid (but not the underlying object), the nullcheck will fail.
However if I change it to `if (myActor)` it works, because [Object.cs overloads `operator bool(Object obj)`](https://github.com/FlaxEngine/FlaxEngine/blob/389dfb44de722cafa1abe5184f1b41938572e38a/Source/Engine/Scripting/Object.cs#L192)

**Suggested fix:**
Apply a similar same existance-check to `operator ==` as is used in `operator bool(Object)`, which does the following:
```csharp
return obj != null && obj.__unmanagedPtr != IntPtr.Zero;
```
By the way, Unity also does this.

**Steps to reproduce:**
```csharp
Actor myActor = otherActor;
Actor.Destroy(otherActor);
// ... some frames later
if (myActor != null)
   Debug.Log(myActor.Name);
```

**Flax version:**
1.3.6228

**Test project:**
[Nullcheck.zip](https://github.com/FlaxEngine/FlaxEngine/files/8682349/Nullcheck.zip)

mafiesto4:
Fixed in #1061 and 5e31a678bd678def61c3c9db527f157c719c7fbb

PR #1061, opened 2023-04-30
Add == operator for Actors and Scripts

RuanLucasGD:
Fix a problem when i compare two diferent actors, always returned true.

With this operator turns more easily compare different or null actors.


RuanLucasGD:
#713

Issue #252, opened 2021-02-19
Objects registry issue when setting parent to null during game


stefnotch:
**Issue description:**
When setting an actor parent to null during a running game, Flax ends up complaining about `Objects registry already contains object with ...` once you stop the game.

Here is a log of such a run
```
[ 00:00:58.548 ]: [Info] [PlayMode] Start
[ 00:00:58.564 ]: [Info] Changing editor state from FlaxEditor.States.EditingSceneState to FlaxEditor.States.PlayingState
[ 00:00:58.564 ]: [Info] Apply game settings
[ 00:00:58.564 ]: [Info] Collecting scene data
[ 00:00:58.564 ]: [Info] Saving scene Main to bytes
[ 00:00:58.569 ]: [Info] Scene saved! Time 5 ms
[ 00:00:58.569 ]: [Info] Cleanup graph for scene 'Main'
[ 00:00:58.569 ]: [Info] Unloding 1 lightmap(s)
[ 00:00:58.570 ]: [Info] Gathered 1 scene(s)!
[ 00:00:58.570 ]: [Info] Creating scenes
[ 00:00:58.573 ]: [Info] Loading scene...
[ 00:00:58.573 ]: [Info] Loading 1 lightmap(s)
[ 00:00:58.586 ]: [Info] Created graph for scene 'Main' in 0 ms
[ 00:00:58.586 ]: [Info] Scene loaded in 13 ms
[ 00:00:58.590 ]: [Info] [PlayMode] Enter
[ 00:01:02.262 ]: [Info] [PlayMode] Stop
[ 00:01:02.295 ]: [Info] Changing editor state from FlaxEditor.States.PlayingState to FlaxEditor.States.EditingSceneState
[ 00:01:02.295 ]: [Info] Restoring scene data
[ 00:01:02.295 ]: [Info] Cleanup graph for scene 'Main'
[ 00:01:02.296 ]: [Info] Unloding 1 lightmap(s)
[ 00:01:02.303 ]: [Info] Loading scene...
[ 00:01:02.304 ]: [Info] Loading 1 lightmap(s)
[ 00:01:02.304 ]: [Error] Objects registry already contains object with ID=398e37fc4db178f2fcfae298038b6e0c (type 'FlaxEngine.HingeJoint')! Trying to register object  (FlaxEngine.HingeJoint; 398e37fc4db178f2fcfae298038b6e0c) (type 'FlaxEngine.HingeJoint').
[...]

PR #1078, opened 2023-05-06
Fix SetParent with null value

RuanLucasGD:
fix for #252


mafiesto4:
Actually, that breaks some existing logic, like unparenting actors for custom object pooling.

So when game sets parent to null, then it might not be null due to auto-set which is not intended.

RuanLucasGD:
@mafiesto4  I understood. when the game sets it to null it generates a strange behavior, the object disappears.

PR #1079, opened 2023-05-06
PE: Delayed Shadows.

plemsoft:
Delayed shadows only update cascade shadows each second frame, double the speed of cascade shadows.
Currently set to update all cascades at same speed , looks the best (30 fps).
Only if you have a really fast moving object your able to see the difference.

@mafiesto4 Build errors looks like something from VisjectMeta.cpp , not related to this pull :)


Source/Engine/Level/Actors/DirectionalLight.h
```diff
@@ -18,6 +18,12 @@ class FLAXENGINE_API DirectionalLight : public LightWithShadow
     API_FIELD(Attributes="EditorOrder(65), DefaultValue(4), Limit(1, 4), EditorDisplay(\"Shadow\")")
     int32 CascadeCount = 4;
 
+    /// <summary>
+    /// Delayed shadows only update cascade shadows each second frame, double the speed of shadows. For best result set shadow distance to 8000+ and change resolution to 2048.
+    /// </summary>
+    API_FIELD(Attributes = "EditorOrder(66), DefaultValue(false), EditorDisplay(\"Shadow\")")
+    bool DelayedShadows = false;
```
mafiesto4, line 25: I think it would be a better idea to add `CacheShadows` property to `LightWithShadow` which would allow us to reuse it in the future for other light types (eg. for point/spot lights by using shadow map atlas). Also, static lights could use `CacheShadow` property to bake shadow or shadow map to the lgihtmaps. Plus we could expose function for users to dirty cached shadow (eg. by script during cinematic sequence to have better shadows).

plemsoft: @mafiesto4 Renamed DelayedShadows to CacheShadows and moved it to LightWithShadows.
Yes atlas would be the way to go , not only to cache depth, but also allocated atlas size depending on distance , so near by can have way better resolution , and far away light could use smaller atlas size, would really give AAA shadows :)

mafiesto4:
Shadows were refactored on `1.`9 branch to use new shadow map atlas with variable rate (708fba5136aa07a22bfd84a484a006dfcfe48a3f).

Whole changelist:
https://github.com/FlaxEngine/FlaxEngine/compare/f7470af42dad...00f2a0b825c0

PR #1085, opened 2023-05-07
Format Content Window Tooltips Better, Including Types On Everything.

Menotdan:
This PR aims to make tooltips in Flax easier to read, as well as more consistent. I have added the following functionality:

- Relative paths for all tooltips.
- Basic "Type:" descriptions for several files types (Not Assets)
- "Type:" description generation for Assets. This is done by adding a space before every capital letter. Plenty of edge cases it can hit, room for improvement for sure.
- More consistent tooltips. Folders, Assets, and other Content now all have "Path:" and "Type:", and Assets and Content both have "Size:"

BTW, I forgot to make a new branch for this, hopefully this does not interfere with the PR process.

### DIFF
--- a/Source/Editor/Content/Items/ContentItem.cs
+++ b/Source/Editor/Content/Items/ContentItem.cs
@@ -2,6 +2,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
+using System.Text;
 using FlaxEditor.Content.GUI;
 using FlaxEditor.GUI.Drag;
 using FlaxEngine;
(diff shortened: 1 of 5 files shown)


Source/Editor/Content/Items/ContentItem.cs
```diff
@@ -353,11 +355,19 @@ public virtual void RefreshThumbnail()
...
 
         /// <summary>
-        /// Updates the tooltip text text.
+        /// Updates the tooltip text.
         /// </summary>
         public virtual void UpdateTooltipText()
         {
-            TooltipText = "Path: " + Path;
+            string fileExtension = System.IO.Path.GetExtension(Path);
+            string fileDescription = Utilities.Utils.TranslateFileExtension(fileExtension);
```
mafiesto4, line 363: Actually, we could add `virtual` property getter for file description in `ContentItem` so each type would override it with proper info (eg. including custom assets made by game/plugin). Then `TranslateTypeName` could be implemented in `CSharpScriptItem`/`CppScriptItem`/etc. for each asset type - easier to extend. And I think that TranslateTypeNames` interface option can be removed as it is enough to have this filetype always visible (the extension is already in Path below).

Menotdan: Yeah, I'm cool with that, BUT this solution is what I used because I don't want to go through every single type and try to make them all implement TranslateTypeName. Wouldn't that also break old stuff and stuff from other projects?

Menotdan: Actually to add to this point, I'd like it if there were some way to only implement translatetypename on specific types, and for the other types it will fallback to the original thingy that I implemented, maybe that could be done through a base class. Actually reading your comment now, that might be what you mean lol

mafiesto4: Okay, no problem, I'll do it :)

PR #1086, opened 2023-05-08
Make the UI for linking scale values in a transform more intuitive and easier to use.

Menotdan:
This PR changes the interface for linking the values in a transform scale to make it more intuitive and easier to use. I am using a new class I've created, `IconButton` which lets you easily create a button with an icon.

I have added a button that appears next to the scale option as the link icon, which is a lot easier to find and use than the context menu that was previously in place. (I have left this context menu in place for now)

BTW, There are extra commits in this PR that also were reverted. This is because I forgot to make a new branch for my previous PR, and I'm learning the lesson of why to do that now lol


Menotdan:
I think I have correctly resolved the conflicts in this. Choosing to keep master in these conflicting files as the content in scale-icon-changes is outdated

Menotdan:
Alright, I was wrong something is broken for sure.

Menotdan:
Ok, I got a working build now. It should work. Right? Also nice changes to my previous PR, thank you!

Menotdan:
@mafiesto4 This should be ready to be pulled now, btw

PR #1096, opened 2023-05-11
Input phases

Swiggies:
This PR allows you to poll specific input phases (Waiting, Pressed, Held, Released) rather than having to create many different actions that might share the same Input. This also doesn't take into account what "Mode" the Action uses so you can create significantly less actions. This doesn't change how input currently works and can seamlessly be used with any current project.

An example of how this may be used is if I want to create multiple actions on the same key but I don't want to have to create multiple Actions for them since it gets tedious to maintain. So instead of, for example, creating `JumpPressed`, `JumpHeld`, `JumpReleased`, I could just sample the phase of the Jump Action like so:
devenv_2023-05-11_21-16-43
This would give me the following output
FlaxEditor_2023-05-11_21-16-32

### DIFF
--- a/Source/Engine/Input/Enums.h
+++ b/Source/Engine/Input/Enums.h
@@ -264,6 +264,31 @@ API_ENUM() enum class InputActionMode
 };
 
 /// <summary>
+/// The input action event trigger modes.
+/// </summary>
+API_ENUM() enum class InputActionPhase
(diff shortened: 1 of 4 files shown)


Swiggies:
I've updated it to be State rather than Phase since that makes more sense

PR #1105, opened 2023-05-16
Fix `NetworkManager::ClientConnected` dispatch regression

Withaust:
https://github.com/FlaxEngine/FlaxEngine/commit/953ae3e9bbbfd9d353bd7d2b0b390edbf784b308 has fixed an issue, but also added an initialization problem on its own. Here is a breakdown of the regression:

Currently, if you were spawning any networked actor/component on `NetworkManager::ClientConnected` action (lets say a host player) then checking for ownership would always yield `None`. This is due to the introduction of `NetworkManager::IsConnected()` check in `NetworkReplicator::GetObjectRole` since https://github.com/FlaxEngine/FlaxEngine/commit/953ae3e9bbbfd9d353bd7d2b0b390edbf784b308. It is no longer be possible to request object's role because at that moment `State = NetworkConnectionState::Connected;` was not yet executed, and so `NetworkReplicator::GetObjectRole` would return `None` as a fallback. This PR simply shifts `State` assignment to be executed prior to dispatching `NetworkManager::ClientConnected` in order to eliminate this problem.

### DIFF
--- a/Source/Engine/Networking/NetworkManager.cpp
+++ b/Source/Engine/Networking/NetworkManager.cpp
@@ -318,10 +318,10 @@ bool NetworkManager::StartHost()
 
     // Auto-connect host
     LocalClient->State = NetworkConnectionState::Connected;
-    ClientConnected(LocalClient);
-
     State = NetworkConnectionState::Connected;
     StateChanged();
+
+    ClientConnected(LocalClient);
     return false;
 }
 


mafiesto4:
Actually, this change made Arizona Framework spawn the player twice when playing the local host but that's because there was a bug in the code there. Fixed: https://github.com/FlaxEngine/ArizonaFramework/commit/705741281c1c2995758ccfaca1b9fc7a655032ca

Also, in 0d3bae376157fce471fac7b7fb52b886e482930e I added `LocalClient->State == Connecting` so it might be useful.

PR #1109, opened 2023-05-19
Eyedropper color picker implementation for Windows and Linux

Menotdan:
This PR implements an eyedropper color picker in the `ColorPickerDialog` dialog, for Windows and Linux. Mac support has been lay out for what needs to be implemented.

Also, there should be a unique icon for the color picker, but I'm not sure how to do so, so hopefully it can be figured out.

BTW this commit adds back IconButton, I don't know how, I don't know why, but git decided it should come back. If you feel the need to remove, feel free to do so, it just makes it easier for me.


mafiesto4:
Very cool feature, thanks!

In 9cdd1cbc45a14cd0b5df8f974b37fa1f960cf9db I added `Platform.MousePosition` property for all platforms to access screen-space cursor location so the `ScreenUtils` doesn't need to implement it.

In 62946f63598ae501bd3c217dc7b602a280a44d35 I did some cleanup:
* merge per-platform implementation files to reduce build time (C++ compilation is slow)
* removed`IconButton` and some leftovers in include files ordering
* simplified a lot of code to be slimmer
* added ability to hit `Escape` key when using eye dropper to cancel the action
* changed eye dropper icon to `Search` with a tooltip
* changes color value node in Materials to not recompile the whole material every frame when using the picker

Menotdan:
Thanks for the changes!

PR #1115, opened 2023-05-22
Add the option to import materials for a model as instances of another material in the project.

Menotdan:
This PR adds the option, on model import, to import a model's materials as instances of a specified material so you can make a base material. I found the previous workflow of having to manually create material instances annoying, and this PR fixes that.


Menotdan:
@mafiesto4 just a reminder that this exists! ive been missing this feature every time I've tried to use the engine since then haha

Source/ThirdParty/OpenFBX/ofbx.h
```diff
@@ -1,4 +1,5 @@
 #pragma once
+#include <Engine/Core/Types/Guid.h>
```
NoriteSC, line 2: ???

PR #1119, opened 2023-05-26
Implement `INetworkObject::OnNetworkSync`

Withaust:


Withaust:
BTW, @mafiesto4, are there any specific reason why `INetworkObject` is purely virtual? I think it would have been better to drop `= 0` from all those methods, since it is quite cumbersome having to implement all those methods when in most cases you only really need `Spawn`/`Despawn` and also maaaaybe `Sync`.

Issue #1123, opened 2023-05-26
Ctrl C+Ctrl V ing a prefab will not create a new prefab ID


Bits360:
**Issue description:**
Duplicating a prefab works fine, but ctrl c+ctrl v will cause it to be an exact copy, which will copy changes between the 2 prefabs, which has caused me so many problems.

**Steps to reproduce:**
Read above

**Minimal reproduction project:**
Can be replicated in any project

**Flax version:**
Can be replicated in any version, but am currently am on 1.5

RuanLucasGD:
I confirm this, you can paste the prefab only in another folder, when you try modify the prefab all changes go to original asset, this is very bad

mafiesto4:
Fixed in #1124

PR #1124, opened 2023-05-26
Copy content files using Content Database instead of Content Importing

GoaLitiuM:
Fixes an issue where copying asset files does not change the asset ID, causing the copied assets to remain linked to the original files.


GoaLitiuM:
Fixes #1123

Issue #1081, opened 2023-05-06
Suggestion: Change the direction of the Capsule Collider without changing the rotation


PhyresiCompany:
https://user-images.githubusercontent.com/101017436/236641925-33abdbb8-a3e5-4f39-bd31-8629443c48fa.mp4

mafiesto4:
There is `UpDirection` property in Character Controller but it needs some improvements (eg. debug preview of the current state).

PR #1141, opened 2023-05-31
Add collider direction for orienting capsule collider.

Tryibion:
Adds the ability to rotate capsule collider to the x, y, or z axis. This also adds the ability to orient any collider without changing the orientation of the actor (although it is not exposed to the editor as the main purpose of this PR is for the capsule collider). #1081

image
Image shows the name "Direction" but it has been changed to "Collider Direction" because otherwise it will hide Actor.Direction. if there is a better name, let me know and I will update it.

image


PhyresiCompany:
@Tryibion Thank you, this resource will be very useful
- Can you also add this feature to the CharacterController capsule?

mafiesto4:
Reverted in 41cff47ca36a0d64da7438f39b2ea8d0990505fa as it was causing issues in some existing user projects.

PR #1148, opened 2023-06-04
Usability improvements in Model/SkinnedModel/Material windows

PhyresiCompany:
**Improvements to Model / SkinnedModel windows**
- I added a Button for the **CurrentLOD**.
- I added a Button that centers the model.
- I fixed the **ScreenSize** and added a callsign for Auto LOD.
- I added a DropdownButton to the **PreviewLOD** that works dynamically based on the amount of LODS in the model.

Captura de Tela (276)

**Improved material window**
- I added a DropdownButton to the model

Captura de Tela (277)

⚠️**This commit of mine `Separation of some Previews to fix bugs` corrects some problems with the `AnimationGraph` / `SkeletonMask` windows and adds a `Button CenterModel` for the `SkeletonMask` and for the `CollisionData`**

**Before:**

**After:**

### DIFF
--- a/Source/Editor/Viewport/Previews/AnimatedModelPreview.cs
+++ b/Source/Editor/Viewport/Previews/AnimatedModelPreview.cs
@@ -5,6 +5,7 @@ using FlaxEditor.GUI.ContextMenu;
 using FlaxEngine;
 using FlaxEditor.GUI.Input;
 using FlaxEngine.GUI;
+using FlaxEditor.Viewport.Widgets;
 using Object = FlaxEngine.Object;
 
 namespace FlaxEditor.Viewport.Previews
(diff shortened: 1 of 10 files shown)


Menotdan:
Thank you, this is great! 
Only criticism is that typically you should split your code into more commits

PhyresiCompany:
> Thank you, this is great! Only criticism is that typically you should split your code into more commits

@Menotdan Ok, thank you very much for your feedback, in my next Pull Requests I will split more commits

mafiesto4:
Why this adds `SkeletonMaskPreview` instead of using `AnimatedModelPreview`? This introduces lots of code duplication (more stuff to maintain).

PhyresiCompany:
> Why this adds `SkeletonMaskPreview` instead of using `AnimatedModelPreview`? This introduces lots of code duplication (more stuff to maintain).

@mafiesto4 I had added my `PreviewLOD` modifications to the **AnimatedModelPreview** but this caused a problem in **AnimationGraph** and **SkeletonMask**.

- As shown in these images, the **PreviewLOD** was also added to the AnimationGraph and SkeletonMask previews, which makes no sense to change the `LOD` in these windows
image

- With the new view, the problem has been fixed. In addition to removing the PreviewLOD from these windows with the new preview I also removed the CurrentLOD from these windows

image

**I could have solved the PreviewLOD problem another way. But creating a new preview I found more organized**

PR #1154, opened 2023-06-09
Remove spline editor

RuanLucasGD:
After #1142, the editor is not needed because its options are only for defining smooth or flat curve points. Before you needed these options when duplicating a spline point, now these options are irrelevant as this is already done automatically.


mafiesto4:
Actually, when working with splines for content-generation (eg. procedural road) the tangents switching between linear/smooth makes sense (eg I used it).

PR #1155, opened 2023-06-10
Don't let rename source/content folders on content editor

RuanLucasGD:


Source/Editor/Content/Items/ContentFolder.cs
```diff
@@ -113,7 +113,16 @@ public bool ContainsChild(string path)
         public override ContentItemSearchFilter SearchFilter => ContentItemSearchFilter.Other;
 
         /// <inheritdoc />
-        public override bool CanRename => ParentFolder != null; // Deny rename action for root folders
+        public override bool CanRename
+        {
+            get
+            {
+                var hasParentFolder = ParentFolder != null;
+                var isContentFolder = IsFolder && Path == Editor.Instance.GameProject.ContentFolderPath;
```
mafiesto4, line 121: This doesn't work for `Content`/`Source` in plugin/engine projects. It would be easier with `!(Node is MainContentTreeNode)` - `MainContentTreeNode` is created for main project folders such as Content and Source (see `ContentDatabaseModule.OnInit`).

mafiesto4 (changes requested):

Source/Editor/Content/Items/ContentFolder.cs
RuanLucasGD: @mafiesto4 thanks mafi!
Exist any way to detect `Source `folder? 
Do you want i remove `ContentFolderPath` and `SourceFolderPath` ?

RuanLucasGD: I think it's done then.

mafiesto4 (approved):

PR #1204, opened 2023-06-24
Add Realloc with alignment.

Tryibion:
A function like this was requested by Epic when putting together the Epic online services.


mafiesto4:
Actually, maybe we could name it `Realloc` and add `uint64 oldSize` param and remove `Realloc` that doesn't take `old` size value? Because the existing code seems invalid to copy data via `newSize` while it could be higher than `oldSize`

Tryibion:
> 

Hmmm. maybe, except that the epic games services params need the parms of pointer, size, and alignment, but not old size for its callback.... unless you know of a way to get the old size that I would need for that extra param. `void* EOS_MEMORY_CALL EOSReallocateMemory(void* pointer, size_t sizeInBytes, size_t alignment)`

mafiesto4:
Okay makes sense. So maybe we name it `ReallocAligned`?

PR #1206, opened 2023-06-25
Implement `NetworkReplicator::GetSyncClientId()`

Withaust:


Withaust:
This will only fire once per object with the first and last client getting set to be a `SynchedClientId`. This has no use.

PR #1210, opened 2023-06-27
Add *BoundsBox.IsOnView*

RuanLucasGD:
This feature allow verify if a boundBox is visible by any camera. Util to check if a model is on screen for example.

### DIFF
--- a/Source/Engine/Core/Math/BoundingBox.cpp
+++ b/Source/Engine/Core/Math/BoundingBox.cpp
@@ -5,6 +5,7 @@
 #include "Matrix.h"
 #include "Transform.h"
 #include "../Types/String.h"
+#include "../../Level/Actors/Camera.h"
 
 const BoundingBox BoundingBox::Empty(Vector3(MAX_float), Vector3(MIN_float));
 const BoundingBox BoundingBox::Zero(Vector3(0.0f));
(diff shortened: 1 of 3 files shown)


Source/Engine/Core/Math/BoundingBox.cs
```diff
@@ -115,6 +115,57 @@ public BoundingBox(Vector3 minimum, Vector3 maximum)
...
+                }
+            }
+
+            // verify all box points to check if is on field of view
+            for (int i = 0; i < GetCorners().Length; i++)
+            {
+                // even if the corner isn't on screen return true if the box is on front of camera
+                // and on field on screen
+
+                float distanceOfCorner = Vector3.Distance(cameraPosition, GetCorners()[i]);
```
Withaust, line 155: Replace `float` with `Real` to get rid of CI errors from testing it with UseLargeWorlds

RuanLucasGD: @Withaust  I need use Float3 instead of Vector3?

Withaust: > @Withaust I need use Float3 instead of Vector3?

After looking at implementation of other methods in `BoundingBox`, `Vector3` should be fine.

Source/Engine/Core/Math/BoundingBox.cpp
```diff
@@ -14,6 +15,56 @@ String BoundingBox::ToString() const
     return String::Format(TEXT("{}"), *this);
 }
 
+ bool BoundingBox::IsOnView(Camera* camera)
+{
+     if (&camera == nullptr)
```
Withaust, line 20: I'd probably also change this to `if (camera)` to be consistent with the rest of the engine

Withaust:
Mafi is probably going to clean up the code on merge anyway, so it's fine

Withaust (approved):

PR #1212, opened 2023-06-27
Release custom assets before binary modules on scripting release

GoaLitiuM:
The engine crashes when custom assets are being disposed after the binary modules are unloaded.


Withaust:
This also fixes some random crashes with `SoftAssetReference` cleanup that I've had for a while now

PR #1214, opened 2023-06-28
Fix vehicle wheel debug orientation

RuanLucasGD:
this fix a problem on vehicle wheel debug, showing the cilinder with incorrect orientation.

**before**
image

**after**
image


mafiesto4:
Actually, there must have been another solution for this issue for my test car it's not correct with this PR:
image

mafiesto4:
Maybe this will work for you car too?
```cpp
const Quaternion wheelDebugOrientation = Quaternion::Euler(-data.State.RotationAngle, data.State.SteerAngle, 0) * Quaternion::Euler(90, 0, 90);
```

RuanLucasGD:
@mafiesto4 I had done it that way because Debug didn't work if the wheels were with identity rotation.

this code works for me
`const Quaternion wheelDebugOrientation = GetOrientation() * Quaternion::Euler(-data.State.RotationAngle, data.State.SteerAngle, 0) * Quaternion::Euler(90, 0, 90);`

image

Issue #1229, opened 2023-07-01
Problem with collider hierarchy


PhyresiCompany:
Depending on how the hierarchy is, collider does not work. I tested it with Box Collider but it may happen with other actors.

https://github.com/FlaxEngine/FlaxEngine/assets/101017436/6e43007e-0910-4eee-b69b-e13adee72e65

PR #1230, opened 2023-07-01
Add functionality to add collider anywhere as a child a rigidbody.

Tryibion:
This is some nice functionality to not limit the rigid body of only havening colliders directly as a child but can have a collider as a child of a child. #1229


Menotdan:
@mafiesto4 This should probably be merged if you can be sure the write conflicts can be fixed without causing a regression.

cNori:
this is incompile implementacion

Tryibion:
Closing in favor of #3757 or #2325

PR #1284, opened 2023-07-30
Require semicolon after most engine API macros

GoaLitiuM:
The documentation seems to suggest adding semicolon after most API macros, but this was not enforced in any way. IDEs seems to nag about the unnecessary semicolons after these macros, so why not enforce the usage of semicolons?

The trick to enforce this uses `static_assert` at the end of the macro to trigger a compile time error when the semicolon is missing. For same cases just removing the semicolon at the end was enough.

### DIFF
--- a/Source/Engine/Content/BinaryAsset.h
+++ b/Source/Engine/Content/BinaryAsset.h
@@ -10,7 +10,8 @@
 #define DECLARE_BINARY_ASSET_HEADER(type, serializedVersion) \
 	DECLARE_ASSET_HEADER(type); \
 	static const uint32 SerializedVersion = serializedVersion; \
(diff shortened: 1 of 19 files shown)


NoriteSC:
all is fine but usage of static_assert(true, "")
to enforce semicolon, is ok but is not striped away from Distrybution
builds the compiler dos not care what is at athe end of macros
static_assert(true, "") acualty teaks time on cpu

and i have feeling some IDE's will not like the semicolon at the end of macro because of global functions

GoaLitiuM:
`static_assert(true, "")` does not generate any additional code, but it probably would slow down the compile times but only by a very miniscule amount.

PR #1266, opened 2023-07-24
Content Browser Improvement [ready]

NoriteSC:
content browser is no longer showing source files (.png,/.fbx)
image
inside ContentFilter class there is SuportedFileExtencionsInContentFolder/SuportedFileExtencionsInSourceFolder list as a config if something is missing add extencion there 

folder switcher opened by pressing on separator in navigation bar
image
Settings opcions
image
it will be hard to expand because how multi selection works

PR #1290, opened 2023-08-03
Default material override

NoriteSC:
allows user to override Default material with out editing engine source
image
when the https://github.com/FlaxEngine/FlaxEngine/pull/1266 will be merged i will hide material instance in setting folder permanently


Menotdan:
> when the #1266 will be merged i will hide material instance in setting folder permanently
Could you explain this better? I'm not sure what you will be hiding based on this.

NoriteSC:
> > Could you explain this better? I'm not sure what you will be hiding based on this.
code is creating a material instance in setting when material is assigned to the field

PR #1291, opened 2023-08-04
Add recursive methods for layers/tags

Withaust:


mafiesto4:
I think it should be `API_FUNCTION()`

Withaust:
> I think it should be `API_FUNCTION()`

Whoops, lemme change that.

Withaust:
Fixed it, CI errors are unrelated to the PR.

Withaust:
@mafiesto4 Fixed it, CI errors are unrelated to the issue though.

Withaust:
@mafiesto4 Fixed

Withaust:
I decided to also add recursive methods for tags, but I noticed that `String`/`StringView` versions of setting/getting tags have `/// [Deprecated in v1.5]` comments. This is a long standing topic that I never had the chance of asking about. How is deprecation handled? Should I consider simply removing them (considering we are on the 1.6 for a month now), or is this some sort of a unique compatibility issue?

Withaust:
I also decided to extend recursive addition to the tags as well.

PR #1292, opened 2023-08-05
Add options to edit splines

RuanLucasGD:
Hello, edit splines is a hard work on Flax XD

I made some edit options to be more easy edit.

[Desktop 2023.08.05 - 11.27.48.01.webm](https://github.com/FlaxEngine/FlaxEngine/assets/79365912/6e355fe4-df46-4188-b8c9-ccd03ed9436f)

* Is more easy select spline point or tangent on long distances.
* Added "Linear" option to set selected point (or tangent) as linear.
* On set linear, the spline point is automatically selected.
* Added "Smooth In" option to set Tangent In as linear of current point selected.
* Added "Smooth Out" option to set Tangent Out as linear of current point selected.
* When i set as smooth in, the tangent in is automatically selected.
* When i set as smooth out, the tangent out is automatically selected.
* Added "Aligned" option to set tangent in and tangent out aligneds.
* Added "Free" option to move some tangent independent of another tangent.

**Obs:**

I made a class for each edit option because is more easy to add anothers options.
I don't made the spline point gizmo size bigger for longer distances because OnDebugDraw  don't work on Spline.cs, i need to pass [DrawSpline ](https://github.com/FlaxEngine/FlaxEngine/blob/f29cd1b7b855b49712b3aeeffdd5cb230c4918a9/Source/Engine/Level/Actors/Spline.cpp#L464C14-L464C14) from cpp to cs.  

@mafiesto4 can add this icons pls?

Icons


Source/Editor/SceneGraph/Actors/SplineNode.cs
```diff
@@ -279,6 +286,9 @@ public override void OnDispose()
             }
         }
 
+        private const float PointNodeSize = 1.5f;
+        private const float TangentNodeSize = 1.0f;
```
Withaust, line 290: Replace `float` with `Real` here.

Source/Editor/SceneGraph/Actors/SplineNode.cs
```diff
@@ -398,6 +408,43 @@ internal static void OnSplineEdited(Spline spline)
             }
         }
 
+        internal static float NodeSizeByDistance(Vector3 nodePosition, float nodeSize)
```
Withaust, line 411: Same thing here, you need to replace all new mentions of `float` with `Real` instead, this way you will fix CI/CD compilation issue that you get from `LargeWorlds` compilation mode.

Withaust (changes requested):

Source/Editor/SceneGraph/Actors/SplineNode.cs
RuanLucasGD: @Withaust thanks

Withaust (approved):

mafiesto4:
Cool stuff! I changed `OnDebugDrawSelected` in `Spline.cpp` so it's properly drawn as full white when selected. I added icons and changed those buttons into nice tabs with `icon+name` pair for usability and cool visuals:

image

PR #1294, opened 2023-08-07
`NetworkReplicator::EndInvokeRPC` fix annoying error that happens due to an oversight

Withaust:
Currently, there is an annoying error that can be triggered completely legitimately. This happens due to fact that this check is done immediately upon calling `RPC` on an object that might as well already be registered, and is already in `SpawnQueue` simply waiting to be added to the `Object` list on the next frame. Detouring this check and moving it in a place where it actually matters and makes more sense fixes the issue.

### DIFF
--- a/Source/Engine/Networking/NetworkReplicator.cpp
+++ b/Source/Engine/Networking/NetworkReplicator.cpp
@@ -1127,13 +1127,6 @@ bool NetworkReplicator::EndInvokeRPC(ScriptingObject* obj, const ScriptingTypeHa
     rpc.Info = *info;
     rpc.ArgsData.Copy(Span<byte>(argsStream->GetBuffer(), argsStream->GetPosition()));
     rpc.Targets.Copy(targetIds);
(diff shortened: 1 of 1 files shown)


Withaust:
Actually, now that I think about it, this error does not even have to be present at all, even in update loop check due to the fact that, for example, when we are dispatching RPCs and some of the objects got despawned, it wont make sense to print out the error since the behavior of dropping RPCs for despawned objects is intended. Because of this, I believe it is better to just get rid of this error message altogether then.

mafiesto4:
Maybe we could go back to the initial version of this PR and simply cache (editor-only) list of object IDs that were despawned so user who wants to call RPC on an object not added to networking will get nice error message in Editor?

Withaust:
> Maybe we could go back to the initial version of this PR and simply cache (editor-only) list of object IDs that were despawned so user who wants to call RPC on an object not added to networking will get nice error message in Editor?

Sounds reasonable, I'll make a change that would cover this.

Withaust:
@mafiesto4 Done

PR #1305, opened 2023-08-13
Cache ManagedDictionary types and helper methods

GoaLitiuM:


GoaLitiuM:
I believe it's not necessary, the information stored in the dictionary is just managed handles to types, which do not get reused after hot-reload (the type references should be different between reloaded assemblies, so different managed handles as well).

PR #1312, opened 2023-08-13
.NET7: Optimize ScriptingObject managed initialization and marshalling

GoaLitiuM:
Some notable changes in this PR:
- `MClass` is now stored in managed side, and is directly returned through some interop methods to avoid excessive dictionary lookups.
- Most of the `ScriptingObject`s initialization in managed side is now done in `ScriptingObjectCreate` interop method, and changing the internal values are done via `SetInternalValues` to combine small managed method calls into larger ones.
- Split marshalling delegates (`ToManaged*`/`ToNative*` methods) to further apart in order to avoid runtime type checks in marshalling methods.
- `Type` is internally wrapped with `TypeHolder` to contain commonly accessed information of the type.
- Field offsets are cached and used in the marshalling methods over `FieldInfo`. The `__unmanagedPtr` field offset is cached in native side and accessed with more direct method now.
- Minor changes like `GetTypeGCHandle` -> `GetTypeManagedHandle`


GoaLitiuM:
Rebased to latest master

Issue #1298, opened 2023-08-10
Add runtime switch to borderless and fullscreen borderless window


Tryibion:
Hello, I have not seen a way to switch to borderless and fullscreen borderless window mode while the game is running. Can this be added?

mafiesto4:
Added in #1315

PR #1315, opened 2023-08-14
Add borderless window switch at runtime on Windows platform

Tryibion:
Add changing to borderless window at game runtime. This also adds a nice utility function in the `Screen` class to easily change window mode. I do not have Linux or Mac, so I did not implement on those platforms. #1298


Tryibion:
I totally forgot a getter method. Thanks for adding that.

PR #1317, opened 2023-08-16
Add to be able to choose sloppy mesh simplification or not in model import settings for LODs

Tryibion:
This adds the ability do do a non sloppy mesh optimization for LODs if the user wants. It can give some better results than doing a sloppy simplification.

### DIFF
--- a/Source/Engine/Tools/ModelTool/ModelTool.h
+++ b/Source/Engine/Tools/ModelTool/ModelTool.h
@@ -322,6 +322,12 @@ public:
         // The target amount of triangles for the generated LOD (based on the higher LOD). Normalized to range 0-1. For instance 0.4 cuts the triangle count to 40%.
         API_FIELD(Attributes="EditorOrder(1130), EditorDisplay(\"Level Of Detail\"), VisibleIf(nameof(ShowGeometry)), Limit(0, 1, 0.001f)")
         float TriangleReduction = 0.5f;
+        // Whether to do a sloppy mesh optimization. This is faster and does not follow the topology of the original mesh.
+        API_FIELD(Attributes="EditorOrder(1140), EditorDisplay(\"Level Of Detail\"), VisibleIf(nameof(ShowGeometry))")
+        bool Sloppy = true;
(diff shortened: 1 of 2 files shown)


Source/Engine/Tools/ModelTool/ModelTool.h
```diff
@@ -322,6 +322,12 @@ API_CLASS(Namespace="FlaxEngine.Tools", Static) class FLAXENGINE_API ModelTool
         // The target amount of triangles for the generated LOD (based on the higher LOD). Normalized to range 0-1. For instance 0.4 cuts the triangle count to 40%.
         API_FIELD(Attributes="EditorOrder(1130), EditorDisplay(\"Level Of Detail\"), VisibleIf(nameof(ShowGeometry)), Limit(0, 1, 0.001f)")
         float TriangleReduction = 0.5f;
+        // Whether to do a sloppy mesh optimization. This is faster but does not follow the topology of the original mesh.
+        API_FIELD(Attributes="EditorOrder(1140), EditorDisplay(\"Level Of Detail\"), VisibleIf(nameof(ShowGeometry))")
+        bool SloppyOptimization = true;
+        // Only used if Sloppy is false. Target error is an approximate measure of the deviation from the original mesh using distance normalized to [0..1] range (e.g. 1e-2f means that simplifier will try to maintain the error to be below 1% of the mesh extents).
+        API_FIELD(Attributes="EditorOrder(1150), EditorDisplay(\"Level Of Detail\"), VisibleIf(nameof(SloppyOptimization), true), Limit(0.01f, 1, 0.001f)")
```
Withaust, line 329: Maybe it would be reasonable to add `!SloppyOptimization` check in `VisibleIf` here to prevent confusion, since it does not seem to be accounted for in sloppy call of the simply function?

Tryibion: Correct, the `LODTargetError` will only show if `SloppyOptimization` is false hence the `true` in the `VisibleIfAttribute`

Withaust: Oh, so the logic on that is yet again inverted, my mistake.

PR #1318, opened 2023-08-16
PE: Fix - Game Cooker Crash

plemsoft:
.NET will crash (LibraryImports.g.cs) if you sent a string[] == null , if .length == 0 there is no problem so: public string[] CustomDefines = { };


mafiesto4:
Fixed in more generic way here: b78db755ec1b807abaeafad1a5ca394dc722e040 but thanks for spotting the problem in the recent `master`

PR #1319, opened 2023-08-16
Use mutex over atomic operations in `Delegate`

GoaLitiuM:
Using mutex seems to perform faster in heavy operations like deserializing thousands of `Model`s in prefab at once.


mafiesto4:
Cool! I tested various scenes and indeed it makes the scenes load faster. Even objects/prefabs spawning went high up!
I left a safety define in 936018c49ea3caa085e8b876ec74bc560379b140 for anyone who wants it backward compatible (because now `Bind==BindUnique`).

PR #1334, opened 2023-08-23
Prevent AbstractWrapper from showing in Add AnimEvent CM.

Tryibion:


Tryibion:
Thanks! I did not realize they had that attribute.

PR #1335, opened 2023-08-24
Add automated plugin project creation and Git cloning.

Tryibion:
This adds an automated way into the engine to create a plugin project and to git clone a plugin project. I am sure there are improvements that can be made to it.


Tryibion:
Hmm fails due to ZipFile not existing in current context... Do I need to add something somewhere?

mafiesto4:
Awesome tool!

Issue #1080, opened 2023-05-06
Suggestion: Progress Bar Improvements


PhyresiCompany:
**Filled vertical/horizontal image**

https://user-images.githubusercontent.com/101017436/236641602-17c0db9b-c4bd-44ff-b58a-3704da22a53d.mp4

**Filled radial 360 image**

https://user-images.githubusercontent.com/101017436/236641609-66841c0d-5ff6-4269-98e0-4d8c56d23958.mp4

RuanLucasGD:
#1355

Myterian:
So, what happend to the radial filled progress bar?

PR #1355, opened 2023-08-30
Improve progress bar

Tryibion:
Improve progress bar to add switch between clipping and stretch and add direction. Let me know if it should be broken down into more options or if the options are fine as I put them.


RuanLucasGD:
#1080

PR #1360, opened 2023-09-02
Optimize PhysX collision event handling and onContact callback

GoaLitiuM:
Few improvements to collision event handling and callbacks:

- Use PhysX `PxContactPairFlag::eACTOR_PAIR_HAS_FIRST_TOUCH` and `PxContactPairFlag::eACTOR_PAIR_LOST_TOUCH` flags to determine enter and exit collision information.
- Unsubscribe from persisting contact information as this is not used by the engine, only initial and last contact information are needed.
- Extract collision velocities more efficiently (this assumes the pairs in extra data stream are in the same order as in the contact stream)


mafiesto4:
Great!

Issue #1328, opened 2023-08-20
Crash playing scene animation


Tryibion:
**Issue description:**
Create scene animation, add UI control, add control, add a variable. enable live preview. press play. Flax will crash.

```
Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
Repeat 2 times:
 --------------------------------
at FlaxEngine.SceneAnimationPlayer.Internal_Tick(IntPtr, Single)
--------------------------------
at FlaxEngine.SceneAnimationPlayer.Tick(Single)
at FlaxEditor.Windows.Assets.SceneAnimationWindow.Update(Single)
at FlaxEngine.GUI.ContainerControl.Update(Single)
at FlaxEngine.GUI.ContainerControl.Update(Single)
at FlaxEngine.GUI.ContainerControl.Update(Single)
at FlaxEngine.GUI.ContainerControl.Update(Single)
at FlaxEngine.GUI.ContainerControl.Update(Single)
at FlaxEngine.GUI.ContainerControl.Update(Single)
at FlaxEngine.GUI.ContainerControl.Update(Single)
at FlaxEngine.GUI.RootControl.Update(Single)
at FlaxEngine.Window.Internal_OnUpdate(Single)
at FlaxEngine.Interop.NativeInterop+Invoker+InvokerNoRet1`2[[System.__Canon, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.Single, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].MarshalAndInvoke(System.Object, FlaxEngine.Interop.ManagedHandle, IntPtr)
at FlaxEngine.Interop.NativeInterop.InvokeMethod(FlaxEngine.Interop.ManagedHandle, FlaxEngine.Interop.ManagedHandle, IntPtr, IntPtr)
```

**Flax version:**
1.6

Tryibion:
![image](https://github.com/FlaxEngine/FlaxEngine/assets/71274967/8c1660f3-1489-4692-876e-be517e24e72e)

PR #1363, opened 2023-09-03
.NET7: Fix managed UTF-8 string allocation not using correct encoding

GoaLitiuM:
Serializing/deserializing scene objects with strings which had Unicode characters did not get encoded properly (for example the text in UI labels).


Tryibion:
Does this fix <https://github.com/FlaxEngine/FlaxEngine/issues/1328>?

mafiesto4:
@Tryibion I don't know it it fixes scene anims for UI but I'll fix that bug later on.

Issue #1365, opened 2023-09-03
Incorrect focus on Character Controller


RuanLucasGD:
**Issue description:**
Focus on character controller is not working well. The bounds seem to be the wrong size and the focus goes to the center of the world.

**Steps to reproduce:**
[Development 2023.09.03 - 10.58.05.03(1).webm](https://github.com/FlaxEngine/FlaxEngine/assets/79365912/8923c71f-4207-41c3-9628-9c01c37aef5d)

**Flax version:**
1.6

MineBill:
I've been doing some digging to figure this out. Here is what i learned so far:

The editor uses the actor's bounding box and focuses on the center of it. This bounding box is calculated by calculating the box of the actor itself (if it has one) and then expanded by merging the bounding boxes of the children. The CharacterController asks the PhysicsBackend for its bounding box and then merges the boxes of its children. However, the PhysicsBackend doesn't update the bounding box when the game is not running, which is why focusing works correctly if you press play and then switch to the Editor view. So, the PhysicsBackend calculates the box once during the initialization of the scene(in your case near 0, 0, 0) and then expands it with the boxes of the children, which ends up creating a bigger and bigger box the more you move your character from its original position.

Since i don't really know the engine that much, i'm not sure what would be the best way to solve this.

PR #1369, opened 2023-09-07
Fix for #1365

RuanLucasGD:
#1365


mafiesto4:
I improved it a little in 9da686cd525c3ff76fc5b8722b697d26c64688a1 so it matches the CC bounds:

image

PR #1376, opened 2023-09-12
DebugDraw extra

NoriteSC:
fixes
ray direction was dictating the length of ray
now ray is normalized and multiplied by length
added default vaule for
Color& color to be Color::White
[info] c# because
Color color = Color.White
is not valid think to do on c# side the default value is not genereted 
[to do] generate extra overload binding in c#


mafiesto4:
I think the trick here was that `direction` parameter could be `ray.Direction * length` in user code so we don't need to normalize and multiply it.

NoriteSC:
debug stuff should be quick to set up
now is 
DrawRay(origin, direction * length, Color.White);
and if direction is multiplied by large value someone is asking to get float overflow
the GetNormalized() in this case is just a safe guard
in case of Ray struct ray needs to be broken in to base components

GoaLitiuM:
The previous overload for `DrawRay` without the length parameter should remain unchanged.

NoriteSC:
length is opcional default is float.max u can set to magnitude of direction if u like the same behavior as old one or i can add draw direction

RuanLucasGD:
@NoriteSC i think remove "length" from DrawRay is better because is more confortable only pass direction with magnetude instiead of pass direction and lenght separately, or you can add another method for this

NoriteSC:
@RuanLucasGD i added the DrawDirection it will look the same way as old code

Source/Engine/Debug/DebugDraw.h
```diff
@@ -69,15 +69,47 @@ API_CLASS(Static) class FLAXENGINE_API DebugDraw
...
+
+    /// <summary>
+    /// Draws the line in a direction.
+    /// </summary>
+    /// <param name="ray">The ray.</param>
+    /// <param name="color">The color.</param>
+    /// <param name="length">The length of the ray.</param>
+    /// <param name="duration">The duration (in seconds). Use 0 to draw it only once.</param>
+    /// <param name="depthTest">If set to <c>true</c> depth test will be performed, otherwise depth will be ignored.</param>
+    API_FUNCTION() static void DrawRay(const Ray& ray,const Color& color = Color::White, float length = 3.402823466e+38f, float duration = 0.0f, bool depthTest = true);
```
mafiesto4, line 112: This could use `MAX_float` instead of `3.402823466e+38f` - bindings generation should handle it.

NoriteSC: last time i tried it it will not compile c# side but i can change it

PR #1378, opened 2023-09-12
optymaliscion for #1358 and tiles now has consistent x,y pixel offset

NoriteSC:


mafiesto4:
@Tryibion could take a look into this and give some feedback?

Source/Editor/Content/GUI/ContentView.cs
```diff
@@ -714,54 +714,68 @@ public override bool OnCharInput(char c)
...
+                    itemsHeight = ContentItem.DefaultHeight * viewScale;
+                    //width - element / element gives elementRowCount, * it by element width gives actual size
+                    //where last elemet is - it by from width gives offset to the edge now because 
+                    //the for loop contans x += itemsWidth + xSpace;
+                    //so if all of it gets / by element count it will give the scaling
+                    float elementRowCount = Mathf.Floor(width / (itemsWidth + xSpace));
+                    float lastElement = Mathf.Max((elementRowCount * (itemsWidth + xSpace)), 0);
+                    if (_children.Count > elementRowCount)
+                    {
+
```
Tryibion, line 735: Can remove extra line

Source/Editor/Content/GUI/ContentView.cs
```diff
@@ -714,54 +714,68 @@ public override bool OnCharInput(char c)
...
-                    if (x + itemsWidth > width)
+                    var flooredItemsWidth = Mathf.Floor(itemsWidth);
+                    var flooredItemsHeight = Mathf.Floor(itemsHeight);
+                    x = ContentItem.DefaultMarginSize;
+                    for (int i = 0; i < _children.Count; i++)
                     {
-                        x = itemsToFit == 1 ? 1 : itemsWidth / itemsToFit + xSpace;
-                        y += itemsHeight + 7;
+                        var c = _children[i];
+                        c.Bounds = new Rectangle(x,y, flooredItemsWidth, flooredItemsHeight);
```
Tryibion, line 749: It may be worth double checking to make sure that x and y are always ints. May have to Floor them here so we dont get half pixels. Ran into this on the first round (granted we were dividing) where they were now always whole numbers.

Source/Editor/Content/GUI/ContentView.cs
```diff
@@ -714,54 +714,68 @@ public override bool OnCharInput(char c)
...
+
+
```
Tryibion, line 751: Can remove extra line space

Source/Editor/Content/Items/ContentItem.cs
```diff
@@ -706,25 +708,18 @@ public override void Draw()
...
-                    }
+                        Render2D.FillRectangle(clientRect, Parent.ContainsFocus ? style.BackgroundSelected : style.LightBackground);
                     else if (IsMouseOver)
-                    {
-                        Render2D.FillRectangle(textRect, style.BackgroundHighlighted);
-                        Render2D.DrawRectangle(clientRect, style.BackgroundHighlighted);
-                    }
+                        Render2D.FillRectangle(clientRect, style.BackgroundHighlighted);
+                    else
+                        Render2D.FillRectangle(clientRect, style.LightBackground);
```
Tryibion, line 720: The issue that I see with this is that the highlighting on some of the tiles looks different. What if we did what was there before with the "window" around the thumbnail but also add it to the folder item to keep it consistent?

Source/Editor/Content/Items/ContentItem.cs
```diff
@@ -747,7 +742,7 @@ public override void Draw()
             
             // Draw short name
             Render2D.PushClip(ref textRect);
-            Render2D.DrawText(style.FontMedium, ShowFileExtension || view.ShowFileExtensions ? FileName : ShortName, textRect, style.Foreground, nameAlignment, TextAlignment.Center, TextWrapping.WrapWords, 1f, 0.95f);
+            Render2D.DrawText(style.FontMedium, ShowFileExtension || view.ShowFileExtensions ? FileName : ShortName, textRect, style.Foreground, nameAlignment, TextAlignment.Center, TextWrapping.WrapWords, 0.75f, 0.95f);
```
Tryibion, line 745: Any reason for this change? There was a complaint a while about about spacing between the lines was not enough and some letters were too close to the next line.

Tryibion (commented):
Here are some comments. I also think it would be good to make sure that the empty space between the first element of a row to the panel and the empty space on the last element of a row to the panel are the same so that way the items sit in the center.

Source/Editor/Content/GUI/ContentView.cs
NoriteSC: 735 got it

Source/Editor/Content/GUI/ContentView.cs
```diff
@@ -714,54 +714,68 @@ public override bool OnCharInput(char c)
...
-                var flooredItemsWidth = Mathf.Floor(itemsWidth);
-                var flooredItemsHeight = Mathf.Floor(itemsHeight);
-                x = itemsToFit == 1 ? 1 : itemsWidth / itemsToFit + xSpace;
-                for (int i = 0; i < _children.Count; i++)
-                {
-                    var c = _children[i];
-                    c.Bounds = new Rectangle(Mathf.Floor(x), Mathf.Floor(y), flooredItemsWidth, flooredItemsHeight);
+                case ContentViewType.Tiles:
+                    float xSpace = ContentItem.DefaultMarginSize * viewScale;
+                    float ySpace = ContentItem.DefaultMarginSize * viewScale;
```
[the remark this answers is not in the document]
NoriteSC: so 10 pixels spacing between tiles it kinda a lot

Source/Editor/Content/GUI/ContentView.cs
NoriteSC: yup

Source/Editor/Content/Items/ContentItem.cs
NoriteSC: idk i will leave it at 1

Source/Editor/Content/GUI/ContentView.cs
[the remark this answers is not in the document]
Tryibion: Yeah maybe not. But they do seem to be very close together. Maybe 1 or 2 more then?

Source/Editor/Content/GUI/ContentView.cs
NoriteSC: the Bounds is rect with has float2 as x and y ruining the Floor for every element can add up and having stuff shifted by 1 pixel it will happen for all elements it is not something critical

Source/Editor/Content/GUI/ContentView.cs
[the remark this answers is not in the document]
NoriteSC: what is now 
image
+1
image
+2
image
+5
image

i will leave it at +1 @mafiesto4 can change it if he dont like 
mafiesto4 it just dont go above 5 or spacing will get kinda sylly😆

Source/Editor/Content/Items/ContentItem.cs
NoriteSC: i just made consistent Highlights on gray scale offset, nothing more and remove some stuff with dont add to overall feel of the tile
and i doubt the [ContentItem.cs](https://github.com/FlaxEngine/FlaxEngine/pull/1378/files/85081a095d72f36a9bf4ac2e9a3bcb38df866b1a#diff-1885657851b3c22966481a12b5641354c604bf69618a98f118dc2bb758b895fb) is being not drawnd outside the bounds of scroll virew so less draw calls will be better

PR #1383, opened 2023-09-13
Lazy allocate mutex in `Delegate`

GoaLitiuM:


Withaust:
Its probably not my cup of tea, but I still have to ask out of the curiosity. Why are we using `int32* _lockerRefCount` instead of `int32 _lockerRefCount` here? Considering that on x64 arch `sizeof(int32*) == 8`, while on both x32/x64 would be always `sizeof(int32) == 4`, maybe it would be reasonable to replace it in favor of `int32`? Or maybe I am not getting the reason for it being a pointer in the first place.

GoaLitiuM:
After reviewing this again, it looks like the reference counter is not needed at all. When the delegate is copied, the bound functions are copied over to new delegate, so there is no reason to share the reference counter from original delegate.

Issue #1386, opened 2023-09-14
Jetbrains Rider not showing on editors list


ctigrisht:
Since Jetbrains has changed their install directories via toolbox, rider is not detected as an editor anymore.
Please fix the runtime check.

eLeSTRaGo-Dev:
For Linux platform in script `Source/Editor/Scripting/CodeEditors/RiderCodeEditor.cpp` add new search path.

```
void RiderCodeEditor::FindEditors(Array<CodeEditor*>* output)
{
    ...
    SearchDirectory(&installations, localAppDataPath / TEXT(".local/share/JetBrains/Toolbox/apps/rider/"));
    ...
}
```

mafiesto4:
Merged #1390 and #1391 + backported them to `master` branch.

van800:
https://github.com/FlaxEngine/FlaxEngine/pull/1390#issuecomment-1724960495

PR #1390, opened 2023-09-15
Fix detect Rider on Linux platform installed from newest versions toolbox

eLeSTRaGo-Dev:
**Issue:** https://github.com/FlaxEngine/FlaxEngine/issues/1386


van800:
You may consider using the following code: https://github.com/JetBrains/RiderSourceCodeAccess/tree/main/Source/RiderSourceCodeAccess/Private/RiderPathLocator

At the .NET side there is nuget https://www.nuget.org/packages/JetBrains.Rider.PathLocator

PR #1398, opened 2023-09-16
.NET7: Include EditorModules in Assembly lookup paths

GoaLitiuM:
Fixes https://github.com/FlaxEngine/FlaxDocs/pull/106


MineBill:
Tested on Windows 11 and it does fix the issue.

Issue #1331, opened 2023-08-21
Script Randomly Deleted


Swiggies:
**Issue description:**

I was editing my C# script with Visual Studio 2022 when VS prompted me that my project had been modified externally and prompted me to reload. This usually happens after compiling (alt-tabbing back to Flax) but I did not prompt a compile this time and I pressed ignore and my entire script had been deleted from my PC.

**Steps to reproduce:**

Unable to repro consistently. Happened twice in one session.

**Minimal reproduction project:**

**Flax version:**

1.6

Menotdan:
I have had the same issue once, after a similar sequence of events. I was editing a C++ script though.

Swiggies:
Just had it happen again. Nothing in the logs specifies why it might have happened.

`[ 24/08/2023 5:36:01 PM ]: [Info] Content item 'D:\Projects/Flax/FlaxPrototypingTools/Source/PrototypeTools/PrototypingActor.cs' has been removed
[ 24/08/2023 5:36:01 PM ]: [Info] Command: "C:\Program Files (x86)/Flax/Flax_1.6/Binaries/Tools/Flax.Build.exe" -log -genproject  
[ 24/08/2023 5:36:01 PM ]: [Info] Flax.Build 1.0.0
[ 24/08/2023 5:36:01 PM ]: [Info] Generating project files...
[ 24/08/2023 5:36:01 PM ]: [Info] Missing Android SDK. Cannot build for Android platform.
Missing Linux Toolchain. Cannot build for Linux platform.
[ 24/08/2023 5:36:01 PM ]: [Info] Total time: 00:00:00.4623194
[ 24/08/2023 5:36:03 PM ]: [Info] Starting scripts compilation...
[ 24/08/2023 5:36:03 PM ]: [Info] Command: "C:\Program Files (x86)/Flax/Flax_1.6/Binaries/Tools/Flax.Build.exe" -log -logfile= -build -mutex -buildtargets=PrototypeToolsEditorTarget -skiptargets=FlaxEditor -platform=Windows -arch=x64 -configuration=Development -hotreload=".HotReload.6" 
[ 24/08/2023 5:36:03 PM ]: [Info] Flax.Build 1.0.0
[ 24/08/2023 5:36:03 PM ]: [Info] Building targets...
[ 24/08/2023 5:36:03 PM ]: [Info] Missing Android SDK. Cannot build for Android platform.
[...]

PR #1402, opened 2023-09-16
Fix assets getting deleted when modified by external applications

GoaLitiuM:
There is a very small window of files getting erroneously deleted after external application like Visual Studio modifies the file upon saving by creating temporary files and then renaming those to original filename. We should never attempt to delete the file detected as being deleted unless it was prompted by the user.

The way to reproduce the issue can be done by modifying a file in Visual Studio 2022 for example by adding a space and then hitting CTRL+S until the Flax Editor detects the file removed and then generates the solution files again. Thanks to MineBill from Discord for discovering the steps for reproducing this.

### DIFF
--- a/Source/Editor/Modules/ContentDatabaseModule.cs
+++ b/Source/Editor/Modules/ContentDatabaseModule.cs
@@ -643,7 +643,8 @@ namespace FlaxEditor.Modules
         /// Deletes the specified item.
         /// </summary>
         /// <param name="item">The item.</param>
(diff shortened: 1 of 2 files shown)


RuanLucasGD:
@GoaLitiuM maybe related with #1331 ?

@Swiggies

XeroWolf:
Happened to me twice now sigh. I'll had to switch to daily builds then till 1.7 is released.

PR #1404, opened 2023-09-16
Don't treat '.' in folder names as extensions.

### LINKED ISSUE #1392: Renaming folder with dots in name automatically appends an extension
**Issue description:**

When renaming a folder with a dot in its name (Ex: "aa.bb" to "a.b"), the engine automatically appends the "extension" to the renamed string even when user removes it. Only way to remove the auto-appended extension is by recreating the folder.

**Steps to reproduce:**

1) Create a folder. Example: "aa.bb"
2) Rename the folder. Example: "a.b"
3) The engine auto appends ".bb", resulting in "a.b.bb"

**Flax version:**

1.6.6344

https://github.com/FlaxEngine/FlaxEngine/assets/22917863/9c49d15b-589e-44a4-8891-8bb256a05370

MineBill:
Fixes #1392.

### DIFF
--- a/Source/Editor/Modules/ContentEditingModule.cs
+++ b/Source/Editor/Modules/ContentEditingModule.cs
@@ -90,6 +90,12 @@ namespace FlaxEditor.Modules
                     hint = "Too long name.";
                     return false;
                 }
+                
+                if (item.IsFolder && shortName.EndsWith("."))
+                {
+                    hint = "Name cannot end with '.'";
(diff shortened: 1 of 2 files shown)


Tryibion:
lol. i also just submitted a fix for this. #1405. I dont think it is bad for people to have '.' in their folder names.

MineBill:
Lol, 2 minutes apart 😋

PR #1405, opened 2023-09-16
Fix issue with renaming folder if it has '.' in it.

Tryibion:
Fix for #1392.


mafiesto4:
I merged #1404 instead.

PR #1407, opened 2023-09-16
SmoothDamp for Vector2 and Vector3

AndrejStojkovic:
Implemented the SmoothDamp function for Vector2 and Vector3 in C#.

### DIFF
--- a/Source/Engine/Core/Math/Vector2.cs
+++ b/Source/Engine/Core/Math/Vector2.cs
@@ -822,24 +822,6 @@ namespace FlaxEngine
         }
 
         /// <summary>
-        /// Returns the length of a vector.
-        /// </summary>
-        /// <param name="vector">The vector to get the length from.</param>
-        public static Real Length(Vector2 vector)
-        {
-            return (Real)Mathf.Sqrt(vector.X * vector.X + vector.Y * vector.Y);
-        }
-
-        /// <summary>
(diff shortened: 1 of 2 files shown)


MineBill:
Hey, nice work! Since c++ is also a supported language for scripting, I think it's a nice idea to add SmoothDamp to c++ too. If you need help implementing it in c++ feel free to reach out on the discord server, I go by the same name.

NoriteSC:
error CS0266: Cannot implicitly convert type 'double' to 'float'. An explicit conversion exists (are you missing a cast?)
 @AndrejStojkovic 
 just a info because vector can be a double or float u need to use "Real" type definiton

AndrejStojkovic:
@NoriteSC Should be fixed now, can you run the tests?

yjplapian:
Hey Andrej! I wonder if you can also add the SqrMagnitude and Magnitude for these vectors? I was about to do that in a PR myself, but I saw that you already edited the scripts

AndrejStojkovic:
@yjplapian, there is already an implementation for magnitude and squared magnituted - it's Length and LengthSquared. But I implemented static functions for Length and LengthSquared so you can use them as Vector3.Length(vector).

AndrejStojkovic:
Fixed the type casting issue, tests pass on forked repo. Ready for merge. 👍

NoriteSC (approved):

PR #1412, opened 2023-09-16
Add more Editor input binding options

GoaLitiuM:
Highlighted the new input binding options in the settings:
image

Completely new options:

- Added a binding for toggling the camera viewport right-click functionality (Toggle Camera Rotation)
  - This is very useful binding when using laptops with no dedicated or hard to activate mouse right-click
  - Can be cancelled by pressing escape or hitting the toggle button again
- Fixed a bug with bindings with no modifiers getting triggered when any modifiers are pressed
- Added a binding for toggling the Gizmo transform space between world/local spaces (defaults to 4-key)
- Added keyboard bindings for changing the viewport camera movement speed
- Added bindings for opening the Profiler window and triggering the Start/Stop and Clear buttons in the Profiler window with a shortcut

Most new bindings have no default binding unless they had hardcoded binding previously.

### DIFF
--- a/Source/Editor/Editor.cs
+++ b/Source/Editor/Editor.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Runtime.InteropServices.Marshalling;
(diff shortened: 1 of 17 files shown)


GoaLitiuM:
Added following File-menu bindings:
image

PR #1414, opened 2023-09-17
Fixing x64 editor/runtime on M1/2 macs

wackoisgod:
This resolves part of #1254 where when running x64 compiled versions of the runtime/editor on M1/2 macs, it would not properly check the correct location for the x64 version of the dotnet runtime.

This does not fix many of the issues that exist in the build system for building x64 binaries on an M1/2 macs but this does allow folks who download the editor from the CI page to actually be able to run it on a M1/2 mac.

### DIFF
--- a/Source/Engine/Scripting/Runtime/DotNet.cpp
+++ b/Source/Engine/Scripting/Runtime/DotNet.cpp
@@ -1541,7 +1541,16 @@ bool InitHostfxr()
     get_hostfxr_params.size = sizeof(hostfxr_initialize_parameters);
     get_hostfxr_params.assembly_path = libraryPath.Get();
 #if PLATFORM_MAC
-    get_hostfxr_params.dotnet_root = "/usr/local/share/dotnet";
+    ::String macOSDotnetRoot = TEXT("/usr/local/share/dotnet");
+#if defined(__x86_64) || defined(__x86_64__) || defined(__amd64__) || defined(_M_X64)
+    // When emulating x64 on arm
+    const ::String dotnetRootEmulated = macOSDotnetRoot / TEXT("x64");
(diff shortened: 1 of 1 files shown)


Menotdan:
Thank you for this fix! There were several people who wanted to use Mac and couldn't because of this.

mafiesto4:
Thanks!

slipster216:
Just tried the latest build artifact and am still getting the same error.

wackoisgod:
> Just tried the latest build artifact and am still getting the same error.

I downloaded the latest build and it works correctly for me and the application now loads where as before this change it did not. It should be noted @slipster216 that you do need to have the x64 version of the dotnet sdk installed as well as the arm version.

slipster216:
Ah, I only had the arm version installed, not x64 as well. Someone should
note that on the install instructions, as I wouldn't expect to need both..

slipster216:
Well, that lets me load the editor, but loading any sample project gives me an error that says the game scripts can't be loaded, so nothing works..

wackoisgod:
@slipster216 as I mentioned in the comment of the PR this doesn't fix the build system related issues around building an x64 binary on an arm64 machine. This allows the editor to actually run, I can load a project just fine and run it in the editor just fine but it may not be possible to cook and run at this time from an arm64 based machine. That requires other fixes which were out of the scope of this PR. 

The above fix would also resolve issues that if someone did build an x64 game on an x64 mac it would run properly on a arm64 based machine.

slipster216:
I'm building - just running the editor and trying to load a sample project.

wackoisgod:
> Well, that lets me load the editor, but loading any sample project gives me an error that says the game scripts can't be loaded, so nothing works..

I put together this PR https://github.com/FlaxEngine/FlaxEngine/pull/1422 which properly does compile the dlls and I can run and see debug stuff from the compiled dll.

PR #1420, opened 2023-09-17
Better missing script handling

RedTheKitsune:
Currently when we change name or namespace of Script it disappears completely and we can only see warning in output log.
This MR adds DummyScript that gets created automatically when deserialization fails.
image
It holds serialized json data, displays missing script type and allows to transfer serialized data to different Script.

TODO:

- [x] Add XML comments

(I know this is planned for 1.8, but disappearing Scripts during refactoring were annoying for me)

### DIFF
--- a/Source/Engine/Level/Components/DummyScript.h
+++ b/Source/Engine/Level/Components/DummyScript.h
@@ -1,0 +1,67 @@
+﻿#pragma once
+
+#include "Engine/Core/Cache.h"
+#include "Engine/Scripting/Script.h"
+#include "Engine/Scripting/ScriptingObjectReference.h"
+#include "Engine/Serialization/JsonWriters.h"
+
+API_CLASS(Attributes="HideInEditor") class FLAXENGINE_API DummyScript : public Script
+{
+    API_AUTO_SERIALIZATION();
+    DECLARE_SCRIPTING_TYPE(DummyScript);
+
+public:
+    API_PROPERTY(Attributes="ReadOnly")
+    FORCE_INLINE String GetMissingTypeName() const
+    {
+        if(Data.IsEmpty()) return TEXT("");
+        
+        rapidjson_flax::Document doc;
+        doc.Parse(Data.ToStringAnsi().GetText());
+        
+        String str (doc["TypeName"].GetString());
+        
+        return str;
+    }
+    
+    API_PROPERTY()
+    void SetMissingTypeName(String value)
+    {
+        _missingTypeName = value;
+    }
+    
+    API_FIELD(Hidden, Attributes="HideInEditor") String Data;
+    
+    API_PROPERTY()
+    FORCE_INLINE ScriptingObjectReference<Script> GetReferenceScript() const
+    {
+        return _referenceScript;
+    }
+    
+    API_PROPERTY()
+    void SetReferenceScript(ScriptingObjectReference<Script> value)
+    {
+        _referenceScript = value;
+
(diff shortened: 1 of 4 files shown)


Source/Engine/Level/SceneObjectsFactory.cpp
```diff
@@ -238,6 +239,8 @@ void SceneObjectsFactory::HandleObjectDeserializationError(const ISerializable::
     {
         const Guid parentId = JsonTools::GetGuid(parentIdMember->value);
         Actor* parent = Scripting::FindObject<Actor>(parentId);
+        DummyScript* dummyScript = parent->AddScript<DummyScript>();
+        dummyScript->Data = String(buffer.GetString());
```
mafiesto4, line 243: `MissingTypeName` could be cached here from the `const ISerializable::DeserializeStream& value`.

Source/Engine/Level/SceneObjectsFactory.cpp
```diff
@@ -238,6 +239,8 @@ void SceneObjectsFactory::HandleObjectDeserializationError(const ISerializable::
     {
         const Guid parentId = JsonTools::GetGuid(parentIdMember->value);
         Actor* parent = Scripting::FindObject<Actor>(parentId);
+        DummyScript* dummyScript = parent->AddScript<DummyScript>();
```
mafiesto4, line 242: Check if `parent ` is valid as it might be `null`.

Source/Engine/Level/Components/DummyScript.h
```diff
@@ -0,0 +1,67 @@
+﻿#pragma once
+
+#include "Engine/Core/Cache.h"
+#include "Engine/Scripting/Script.h"
+#include "Engine/Scripting/ScriptingObjectReference.h"
+#include "Engine/Serialization/JsonWriters.h"
+
+API_CLASS(Attributes="HideInEditor") class FLAXENGINE_API DummyScript : public Script
```
mafiesto4, line 8: Maybe `MissingScript` would be a better name?

RedTheKitsune:
Done all the requested changes, currently the only problem I have is ``MissingTypeName`` being visible in C# (you can get and set it) because otherwise its not gonna render in Properties. The best solution to this is to allow displaying only getters in UI as readonly fields.

Source/Engine/Level/SceneObjectsFactory.cpp
RedTheKitsune: Done + it cleaned up some code in MissingScript

Source/Engine/Level/SceneObjectsFactory.cpp
RedTheKitsune: Moved everything couple lines down into already existing `if(parent)`

Tryibion:
It would be good to create a custom editor for this and change the header text color of the script GUI to a different color. OrangeRed maybe?

RedTheKitsune:
I digged into the code of the engine and I dont think it is possible to do this through GenericEditor. I either gonna need to add special handling for it in ScriptsContainer or give LayoutElementsContainer ability to directly edit its parent (or atleast add support for per custom editor styling options)

Tryibion:
> I digged into the code of the engine and I dont think it is possible to do this through GenericEditor. I either gonna need to add special handling for it in ScriptsContainer or give LayoutElementsContainer ability to directly edit its parent (or atleast add support for per custom editor styling options)

Can do something like this:
image

PR #1423, opened 2023-09-17
Add automatic module creation.

Tryibion:
This PR adds a way to automatically create a new module via right clicking on the source folder in the editor.

### DIFF
--- a/Source/Editor/Windows/ContentWindow.ContextMenu.cs
+++ b/Source/Editor/Windows/ContentWindow.ContextMenu.cs
@@ -288,15 +288,6 @@ namespace FlaxEditor.Windows
                             popup.Hide();
                         };
                     };
-                    /*
-                    button.Size = new Float2(200, 100);
-                    var nameLabel = new Label()
-                    {
-                        Text = "Name:",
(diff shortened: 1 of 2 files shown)


Source/Editor/Windows/ContentWindow.ContextMenu.cs
```diff
@@ -146,6 +148,148 @@ private void ShowContextMenuForItem(ContentItem item, ref Float2 location, bool
             }
 
             cm.AddSeparator();
+            
+            // Check if is source folder to add new module
+            if (item is ContentFolder sourceFolder && sourceFolder.ParentFolder.Node is ProjectTreeNode node)
```
AlanWills, line 153: Is this restriction on the ParentFolder being the ProjectTreeNode required?  A lot of the docs, for example, have an editor module created in a sub folder called `Editor` in the `Source` folder.

Tryibion: This will only add the context menu option to be added to the source folders right click menu. I figured the best place to create modules is directly under the source folder.

PR #1426, opened 2023-09-18
Fix typo in PxCustomGeometry.h

eltociear:
Assigment -> Assignment


mafiesto4:
This code file comes directly from https://github.com/FlaxEngine/PhysX-5 which is a fork of https://github.com/NVIDIA-Omniverse/PhysX

PR #1446, opened 2023-09-19
Expose GPUTexture::DownloadData to Scripting API

GoaLitiuM:
TODO:
- [x] `GPUTexture::DownloadData`
- [ ] `GPUTexture::DownloadDataAsync`
- [ ] `GPUTexture::UploadMipMapAsync`
- [ ] Other structures?

### DIFF
--- a/Source/Editor/Utilities/EditorUtilities.h
+++ b/Source/Editor/Utilities/EditorUtilities.h
@@ -7,7 +7,7 @@
 
 enum class PixelFormat : unsigned;
 enum class DirectorySearchOption;
-class TextureData;
+struct TextureData;
 
 /// <summary>
 /// Helper functions for the editor.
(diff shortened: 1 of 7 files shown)


mafiesto4:
Implemented via new `MeshAccessor` for 1.10 that is more flexible for any vertex format:

https://github.com/FlaxEngine/FlaxEngine/blob/1.10/Source/Engine/Graphics/Models/MeshAccessor.h

https://github.com/FlaxEngine/FlaxEngine/blob/1.10/Source/Engine/Graphics/Models/MeshAccessor.cs

mafiesto4:
Added in 15da2d59e511c1bd0d4c6fdf710cabd3a44343c6. With some basic medata about image and ability to get all pixels as a `Color` or `Color32` array.

PR #1454, opened 2023-09-20
macOS support fixes

wackoisgod:
* Adding macOS FileSystemWatcher, this should allow files to be monitored and update like the other OSs
* Reworked how macOS launches processes to use NSTask which just deals with escaped and unescaped paths better
* Made a change to the ScriptsBuilder::RunBuildTool, this was adding the escaped values to the path, in reality it should be up to the underlying OS to make sure things are properly escaped, so removed those as they just end up causing issues. Also instead of appending the args to the fileName we just properly use the Arguments variable on the CreateProcessSettings
* No longer use open in order to show files in the finder, we use the proper method selectFile
* made a slight cleanup change to the MacPlatform Tick function
* Added ToNSString functions just to make that easier
* Added a ParseArguments function that will take a string and turn it into an array for NSTask
* Resolves a set of issues where the BuildTool would not properly run and compile the scripts.


wackoisgod:
@mafiesto4 am I to assume the tests here have proper coverage ?

mafiesto4:
Yeah CI runs all tests for PRs/pushes so it's good to merge.

mafiesto4:
Good improvements, thanks! I did some tweaks here: https://github.com/FlaxEngine/FlaxEngine/compare/038c67c819e0f62a8d68ea3458f602f3e153dee4...5c5c64cf761e

PR #1456, opened 2023-09-20
Visject grid snap

Menotdan:
I have added grid snap to visject graphs. (Previous version of PR contained changes relating to automatically calculating sizes of nodes, but since that was buggy, I decided to decrease the scope of this PR.)
The icon for the toolbar isn't ideal, so if someone could find a better one, that would be appreciated.


Source/Editor/Surface/Archetypes/Material.cs
```diff
@@ -281,7 +281,7 @@ internal enum MaterialTemplateInputsMapping
                 Title = "Material",
                 Description = "Main material node",
                 Flags = NodeFlags.MaterialGraph | NodeFlags.NoRemove | NodeFlags.NoSpawnViaGUI | NodeFlags.NoSpawnViaPaste | NodeFlags.NoCloseButton,
-                Size = new Float2(150, 300),
+                Size = new Float2(180, 300),
```
Tryibion, line 284: I know this is being increased for increased font reasons. You could make it more dynamic doing something like this: Style.Current.FontSmall.MeasureText("Tessellation Multiplier").X + 30. As then it should size with the font.

Menotdan: In that case, I would like to make every one of the nodes more dynamic. Maybe there is a way to calculate the minimum size horizontally? Like `max(HeaderText.Measure, (horizontal_port_count * (PortSize + PortPadding) + longest_text_left.Measure + longest_text_right.Measure + text_padding*2))` but it would be more complicated than that..

Tryibion: It may be good to standardize on some equation like that though. Might have to play with it a little to get it right.

Menotdan: I added auto-resizing.

RuanLucasGD:
@Menotdan i'm testing your pr and i finded some problems 
image

MineBill:
I've been liking this so far, currently working a visject grid to complement this PR. However, i would like the ability to change the grid size, preferably through the Editor settings.

RuanLucasGD:
@MineBill idk, but i think this is ideal too

MineBill:
I'm asking because having to change the size every time you open a material would be incredibly annoying.

MineBill:
Also, new nodes created from the context menu are not initially snapped to the grid.

Menotdan:
> Also, new nodes created from the context menu are not initially snapped to the grid.

I can change that if you want.

MineBill:
Yeah that seems like a good idea. Also, we need to reach a consensus about the settings.

Menotdan:
> Yeah that seems like a good idea. Also, we need to reach a consensus about the settings. We need to get this merged if mafi agrees so I can open a PR for my grid.

What settings do we need to agree on? Also @mafiesto4 if this could be reviewed and/or merged, that would be great.

MineBill:
I believe that the grid size should be configurable and persistent, at least when opening closing windows. Should this be an Editor setting or can it be attached in the material window? If there is a chance for this to get in 1.7 then we should merge it as is and i'll include the settings in my PR for the grid.

Menotdan:
> I believe that the grid size should be configurable and persistent, at least when opening closing windows. Should this be an Editor setting or can it be attached in the material window? If there is a chance for this to get in 1.7 then we should merge it as is and i'll include the settings in my PR for the grid.

I think it should be an Editor setting, but maybe as part of the grid PR.

mafiesto4:
Thanks! I did some minor code cleanup. Now, nodes will be snapped (both location and size) when opening graph with snapping enabled (via Editor settings). Also, grid size can be adjusted,

PR #1458, opened 2023-09-20
Allow for better support for running on m1/2 machines

wackoisgod:
* So we need to account for 2 possible situations where you are running under an emulated process and a native process with a different target host in this case x64


mafiesto4:
Cannot wait for Github Runners to support cloud-hosted macOS arm64 for m1/m2 builds https://github.com/github/roadmap/issues/528

wackoisgod:
> Cannot wait for Github Runners to support cloud-hosted macOS arm64 for m1/m2 builds [github/roadmap#528](https://github.com/github/roadmap/issues/528)

Agreed 100%

PR #1460, opened 2023-09-20
Fix white window flickering in context menus

GoaLitiuM:
Avoid showing the window before the first frame is rendered to that window.


Tryibion:
Could you add this in `ContextMenuBase` where the context menus are being created? or do you think it is better here for an "overall" fix?

mafiesto4:
I think we can change the default value to prevent flickering in general. It's mostly Ediotr-only because games not use custom windows creation thart much.

Issue #11, opened 2020-12-19
Win32 in solution configuration


stefnotch:
**Issue description:**
It's possible to select Win32 from the solution configuration. This causes errors to pop up everywhere.

e.g. BasicTemplate
![image](https://user-images.githubusercontent.com/10220080/102695960-2105fe80-422b-11eb-8931-14ddbdbbb614.png)

Is Win32 intentionally left in the solution configuration? Would it be better to change [Flax.Build](https://github.com/FlaxEngine/FlaxEngine/blob/master/Source/Tools/Flax.Build/Build/Builder.Projects.cs) so that it's excluded? (And under which circumstances?)

**Steps to reproduce:**
1. Open the basic template Flax Sample in Visual Studio
2. Open a random script
3. Switch from `Win64` or `Any CPU` to `Win32`
4. Wait a bit

**Minimal reproduction project:**
[Flax Samples](https://github.com/FlaxEngine/FlaxSamples/)

**Flax version:**
Flax 1.0

mafiesto4:
I guess we could improve configurations picking for win32 or drop support for it anyway soon.

mafiesto4:
Finally fixed in #1466

PR #1466, opened 2023-09-20
Prevent generating VS configurations for Win32 and engine projects

GoaLitiuM:
Removes the `Debug` and `Release` solution configurations and `Any CPU` and `Win32` solution platforms from generated Visual Studio solution files sourced from Flax projects. This should hopefully fix Intellisense issues with default solution settings and effectively changes the defaults to `Editor.Debug` and `Win64`.


GoaLitiuM:
Fixes #11

mafiesto4:
I added some more to this: d94c0d3323d50a8110200c5ed47dbc691e049c60 and a420ca72d670def12cd4fd8c2675291f159d8ecc to make VS usage even smoother.

PR #1475, opened 2023-09-21
Remove flawed MSBuild search in favor of using dotnet build

ontrigger:
The MSBuild/Visual Studio detection logic is very likely to fail and is redundant as the .NET SDK already contains it's own version of MSBuild that can be invoked by calling 'dotnet build'. 

This pull completely removes the MSBuild detection/building logic and replaces it with a 'dotnet build' command.
A link to download the latest .NET SDK is displayed in case dotnet is not found


ontrigger:
closing in favor of #1485

PR #1476, opened 2023-09-21
Customizable statusbar

davevanegdom:
This PR adds **customization settings for the Statusbar**, which are available when using a custom Theme. 

It also changes the default color for the Statusbar during PlayMode to a green color instead of the scary orange-red.

For each Theme, the user can edit:
- Statusbar playmode color 
- Statusbar loading color (used when importing, building, etc)
- Statusbar failed color (used on compilation fail, etc)

**Note:** options "Normal" and "Text Color" have been removed (and reverted to their old properties) for the sake of theme consistency.

statusbaroptions

statusbarstyleshort

### DIFF
--- a/Development/Scripts/Windows/CallBuildTool.bat
+++ b/Development/Scripts/Windows/CallBuildTool.bat
@@ -11,16 +11,6 @@ for %%I in (Source\Logo.png) do if %%~zI LSS 2000 (
 call "Development\Scripts\Windows\GetMSBuildPath.bat"
 if errorlevel 1 goto Error_NoVisualStudioEnvironment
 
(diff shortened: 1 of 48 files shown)


davevanegdom:
Maybe something to change back is the fact that you can set the "Normal" color. Originally the "normal"  color was the backgroundSelected color property in the Theme. 

A potential point of frustration here is that the Statusbar would be an element the user has to set separately for it to match all other backgroundSelected styles should they change that color...

Tryibion:
I agree with your thoughts of leaving the normal color as being changed with the default theme colors.

davevanegdom:
@Tryibion I agree. I removed customization for "Normal" as well as the "Text Color" for theme consistency.

PR #1478, opened 2023-09-21
Improved RandomUtil for easier use.

DevHrytsan:
I have added new random methods for easier use without initializing RandomStream.
Now, RandomUtil provides methods for generating random numbers within specified ranges, random points on the surface of the sphere and in the sphere. 
It encapsulates common randomization tasks, making it easier to work with randomness.


Source/Engine/Engine/RandomUtil.cs
```diff
@@ -24,5 +24,80 @@ public static float Rand()
...
+            float v = RandRange(-1, 1); 
+            float w = RandRange(-1, 1); 
+
+            // Normalize the vector to ensure it's inside the unit sphere
+            float length = Mathf.Sqrt(u * u + v * v + w * w);
+            u /= length;
+            v /= length;
+            w /= length;
+
+            return new Float3(u, v, w);
```
stefnotch, line 81: Doesn't this return a biased sample of random points inside a sphere?

Source/Engine/Engine/RandomUtil.cs
```diff
@@ -24,5 +24,80 @@ public static float Rand()
...
         }
+
+        /// <summary>
+        ///Generates a random number within a specified range.
+        /// </summary>
+        /// <param name="min">The minimum value of the range.</param>
+        /// <param name="max">The maximum value of the range.</param>
+        /// <returns>The random number.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static float RandRange(float min, float max)
```
stefnotch, line 35: It might be worth mimicking one of the existing methods for consistency :) https://github.com/FlaxEngine/FlaxEngine/blob/97ed46bc4ff36a0269b50b138c08908461ee9f96/Source/Engine/Utilities/Extensions.cs#L265

stefnotch (changes requested):

Source/Engine/Engine/RandomUtil.cs
stefnotch: Oh, we actually have an unbiased one over there https://github.com/FlaxEngine/FlaxEngine/blob/97ed46bc4ff36a0269b50b138c08908461ee9f96/Source/Engine/Engine/RandomStream.cs#L131

Source/Engine/Engine/RandomUtil.cs
DevHrytsan: I just remade it easier to be inside one class. 
Especially, for people who came from the other game engines.
 It's distributed  but seems I a little overcomplicated it by recreating my own version xD

mafiesto4:
Good utilities, we could add similar `RandomUtil` in C++.

On thing tho: `InsideSphere` and `OnSphere` names don't match `GetUniformOrientation`, maybe we unify them?
Like `GetPointInsideSphere`/`GetPointOnSphere`?

DevHrytsan:
> Good utilities, we could add similar `RandomUtil` in C++.
> 
> On thing tho: `InsideSphere` and `OnSphere` names don't match `GetUniformOrientation`, maybe we unify them? Like `GetPointInsideSphere`/`GetPointOnSphere`?

Yeah, I unified them. I will implement similar functions in Random.cpp class.

DevHrytsan:
I have updated it and added to Random.h. But seems some similar functions are available in RandomStream.h

stefnotch (changes requested):
This one is tricky. So randomness is a very mean topic with lots of little things that can go wrong.

I do apologise for not thinking of some of these things earlier. I could definitely have saved you a lot of time and trouble. :(


The C# RandomUtil class is convenient, because it has a static Random object. My proposal would be
1. Move the class to the `FlaxEngine.Utilities` namespace.
2. Tell users that they can do `RandomUtil.Random.NextFloat()` and whatnot.

Then this class would become a really straightforward one
```csharp
using System;
using System.Runtime.CompilerServices;
using FlaxEngine.Utilities;

namespace FlaxEngine
{
    /// <summary>
    /// Basic pseudo numbers generator utility.
    /// </summary>
    public static class RandomUtil
    {
        /// <summary>
        /// A static random numbers generator.
        /// If you need a thread safe one, or one that can be seeded, please use [`Random`].
        /// </summary>
        public static readonly Random Random = new Random();
    }
}
```

Source/Engine/Engine/RandomUtil.cs
```diff
@@ -24,5 +25,79 @@ public static float Rand()
...
+        }
+
+        /// <summary>
+        /// Generates a random point inside a sphere with radius 1.0.
+        /// </summary>
+        /// <returns>A random <see cref="Float3"/> within the sphere.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static Float3 GetPointInsideSphere()
+        {
+            return Random.NextUnitVector3();
```
stefnotch, line 72: I'm so sorry, I mixed up the "inside sphere" and "on sphere" cases.

On sphere: Random.NextUnitVector3();
Inside sphere: Flax is missing that method. I'd propose adding one to https://github.com/FlaxEngine/FlaxEngine/blob/master/Source/Engine/Utilities/Extensions.cs

The method could look like this. (Generalises to other dimensions as well)
I don't immediately have a good name for it though.
```csharp
Vector3 output;
            Real l;
            do
            {
                // Create random float with a mean of 0 and deviation of ±1
                output.X = NextFloat(random) * 2.0f - 1.0f;
                output.Y = NextFloat(random) * 2.0f - 1.0f;
                output.Z = NextFloat(random) * 2.0f - 1.0f;

                l = output.LengthSquared;
            } while (l > 1);
            return output;
```

Source/Engine/Core/Random.h
```diff
@@ -14,4 +14,92 @@ namespace Random
...
+
+    /// <summary>
+    /// Generates a random number within a specified range.
+    /// </summary>
+    /// <param name="min">The minimum value of the range.</param>
+    /// <param name="max">The maximum value of the range.</param>
+    /// <returns>The random <see cref="int"/>.</returns>
+    inline int RandRange(int min, int max)
+    {
+        return min + (max - min) * Rand();
```
stefnotch, line 37: My remark here is that the C standard library `rand()` function isn't quite as nice as what C# has. (C's rand has so many weird edge cases that don't matter here.) And more importantly, they behave differently.

In C#, `random.Next(0,2)` will generate the values 0 and 1. [The maximum value is exclusive.](https://learn.microsoft.com/en-us/dotnet/api/system.random.next?view=net-7.0)
This also applies to the `Random.NextDouble` function in C#. `Random.NextDouble` will never generate a `1.0f`.

Meanwhile in C world, this class has 
```c
inline float Rand()
{
    return (float)rand() / (float)RAND_MAX;
}
```
Which absolutely can generate `1.0f`, even if the odds of it happening are really low.

This annoyingly leads to `return min + (max - min) * Rand();` having a super low chance of generating the `max` value. I can sketch up a diagram or a better explanation if you're curious.

stefnotch: (I hope someday someone adds a proper random number generation library to Flax, just to put me out of my misery. :P)

DevHrytsan: No worries. but I really was confused with OnSphere and Inside Sphere :D
Yeah, seems rand() is really the problem here. Maybe we would use <uniform_real_distribution> for it?
https://stackoverflow.com/questions/288739/generate-random-numbers-uniformly-over-an-entire-range

Seems Flax Engine really needs some custom random solution, like Unity does.

Source/Engine/Utilities/Extensions.cs
```diff
@@ -224,6 +224,16 @@ public static void Shuffle<T>(this Random random, IList<T> collection)
...
 
+        /// <summary>
+        /// Generates a pseudo-random number from normalized range [0;1].
+        /// </summary>
+        /// <param name="random">An instance of <see cref="Random"/>.</param>
+        /// <returns>The random number.</returns>
+        public static float Rand(this Random random)
+        {
+            return random.Next(0, int.MaxValue) / (float)int.MaxValue;
+        }
```
stefnotch, line 235: I think this can be safely removed

Source/Engine/Engine/RandomUtil.cs
```diff
@@ -2,6 +2,7 @@
 
 using System;
 using System.Runtime.CompilerServices;
+using FlaxEngine.Utilities;
 
 namespace FlaxEngine
```
stefnotch, line 7: The idea behind the "move to FlaxEngine.Utilities" namespace was that users wouldn't end up being confused about where the `.NextInsideUnitVector3`, `NextBool`, `NextByte` methods are.

We could also leave it in the FlaxEngine namespace, but in that case I'd add a documentation comment that says something like `FlaxEngine.Utilities` has a lot of helper methods for random number generation.

Source/Engine/Utilities/Extensions.cs
```diff
@@ -326,6 +336,27 @@ public static Vector3 NextUnitVector3(this Random random)
             return output;
         }
 
+        /// <summary>
+        /// Generates a uniformly distributed random unit vector point inside a unit sphere.
+        /// </summary>
+        /// <param name="random">An instance of <see cref="Random"/>.</param>
+        /// <returns>A random <see cref="Vector3"/>.</returns>
+        public static Vector3 NextInsideUnitVector3(this Random random)
```
stefnotch, line 344: Nice name :)

Source/Engine/Core/Random.h
```diff
@@ -14,4 +15,92 @@ namespace Random
...
+    /// <returns>A random <see cref="Quaternion"/>.</returns>
+    /// <param name="randomRoll">Should the roll value be randomized.</param>
+    inline Quaternion GetUniformOrientation(bool randRoll = false)
+    {
+        float randomYaw = Rand() * 360.0f;
+        float randomPitch = Rand() * 360.0f; 
+        float randomRoll = (randRoll) ? Rand() * 360.0f : 0.0f;
+
+        return Quaternion::Euler(randomYaw, randomPitch, randomRoll);
+    }
```
stefnotch, line 104: Feel free to skip this one for now. Here's an issue https://github.com/FlaxEngine/FlaxEngine/issues/1563 that I just opened.

Source/Engine/Core/Random.h
stefnotch: I'm not a C++ expert, so I can't tell whether that's the best solution. It does seem reasonable though.

Feel free to split up this pull request into one that improves the C# side, and then leave improving the C++ for another pull request. After all, if we have to introduce a proper randomisation function or library to the C++ side, then that definitely warrants a bit more attention.

Source/Engine/Utilities/Extensions.cs
DevHrytsan: It was so hard to find name for it :D

Source/Engine/Utilities/Extensions.cs
DevHrytsan: Done!

Source/Engine/Core/Random.h
DevHrytsan: I have reverted C++ Random.h changes. So now it`s only C# pull request

Arcnor: I'm probably missing something, but shouldn't one of the C++ standard random generators suffice to replace C's `rand`? https://en.cppreference.com/w/cpp/numeric/random

stefnotch (approved):
Very neat! Thank you so much for putting up with all of my comments. <3

Source/Engine/Utilities/Extensions.cs
```diff
@@ -286,7 +286,28 @@ public static float NextFloat(this Random random, float max)
...
+            float sqrtU1 = Mathf.Sqrt(u1);
+
+            float x = sqrt1MinusU1 * Mathf.Sin(rad * u2);
+            float y = sqrt1MinusU1 * Mathf.Cos(rad * u2);
+            float z = sqrtU1 * Mathf.Sin(rad * u3);
+            float w = sqrtU1 * Mathf.Cos(rad * u3);
+
+            Quaternion result = new Quaternion(x, y, z, w);
+
+            if (randomRoll)
```
stefnotch, line 305: Should be `if (!randomRoll)` I suppose? I noticed it just now.

DevHrytsan: Yeah. I was tired when I wrote this function.
I thought I put first to return with a roll and then without.
It's funny that we both didn't notice it at first :D

PR #1485, opened 2023-09-21
Remove flawed MSBuild search in favor of using dotnet build

ontrigger:
This is a recreated version of #1475 without merge conflicts

The MSBuild/Visual Studio detection logic is very likely to fail and is redundant as the .NET SDK already contains it's own version of MSBuild that can be invoked by calling 'dotnet build'.

This pull completely removes the MSBuild detection/building logic and replaces it with a 'dotnet build' command.
A link to download the latest .NET SDK is displayed in case dotnet is not found


Arcnor:
Any reason this was closed instead of merged?

ontrigger:
The version of MSBuild bundled with 'dotnet build' does not support building C++ projects, meaning you still need the Visual Studio version of MSBuild to build the engine, which kind of defeats the point.

The proper solution to this is to create a console app that uses https://github.com/microsoft/MSBuildLocator to find the needed version of MSBuild and use it to set up the project. This console app needs to be prebuilt (how are you gonna build if you dont have MSBuild lmao), and I don't think the engine dev will approve an .exe file being added as part of a pull

Arcnor:
That's really unfortunate. I ran into this while building from source, and found really annoying that you need to install .NET inside VS, I was expecting installing the dotnet SDK to suffice, but what you say makes sense of course.

That said, maybe there is a way of using VS's MSBuild and dotnet's SDK somehow...

Anyway, thanks for the explanation.

PR #1493, opened 2023-09-22
Fix not capturing the the Debug.Log line in the DebugLogWindow

MineBill:
In the debug log window, when using regex to analyze stack traces from Debug.Log, the incorrect group was used to get the line. As a result, when clicking on the log entries on the window, no code editor would open the file at the correct line.


GoaLitiuM:
Needs to be fixed here as well
https://github.com/FlaxEngine/FlaxEngine/blob/faeb21a77a8d3d020fd3c0bcd743391c667c62ce/Source/Editor/Windows/DebugLogWindow.cs#L576

GoaLitiuM:
This also affects Visual Studio, not just Rider

PR #1495, opened 2023-09-22
Fixing non-windows platforms csproj files

wackoisgod:
* Seems there was a hardcoded path ? for the prebuilt binaries for the FlaxEngine dll
* Also fixed some hardcoded paths for the editor


mafiesto4:
You could do sth like to have even better path to those prebuilt Flax C# assembly on all platforms:
```cs
project.CSharp.FileReferences.Add(Path.Combine(Globals.EngineRoot, Platform.GetEditorBinaryDirectory(), "Development/FlaxEngine.CSharp.dll"));
```

PR #1501, opened 2023-09-22
TextBox: Added ctrl+backspace support

Chikinsupu:
In a lot of text editors it is common that ctrl+backspace deletes the word before the caret/cursor. Personally i use this key combination a looooot.
But at the moment pressing ctrl+backspace in Flax will only delete the last character and insert the DEL control character (The box character in the before gif).

This PR filters out the DEL character and deletes the word before the caret.

Before:
Before3

After:
After3

Note: When putting a ton of spaces after a word this doesn't work as expected, since spaces count as seperator characters and the "FindPrevWordBegin()" function stops at seperator chars. The same happens when there are other seperator chars like . , / etc.
This can probably be fixed pretty easily.


MineBill:
image

PR #1506, opened 2023-09-23
Fix invalid characters in generated C# launch settings paths

GoaLitiuM:


Source/Tools/Flax.Build/Projects/VisualStudio/VisualStudioProjectGenerator.cs
```diff
@@ -558,8 +558,8 @@ public override void GenerateSolution(Solution solution)
             {
                 var profiles = new Dictionary<string, string>();
                 var profile = new StringBuilder();
-                var editorPath = Utilities.NormalizePath(Path.Combine(Globals.EngineRoot, Platform.GetEditorBinaryDirectory(), $"Development/FlaxEditor{Utilities.GetPlatformExecutableExt()}"));
-                var workspacePath = Utilities.NormalizePath(solutionDirectory);
+                var editorPath = Utilities.NormalizePath(Path.Combine(Globals.EngineRoot, Platform.GetEditorBinaryDirectory(), $"Development/FlaxEditor{Utilities.GetPlatformExecutableExt()}")).Replace('\\', '/');
```
wackoisgod, line 561: I mean looking at the code in `Utilities.NormalizePath` isn't this already being done? If not it should be and to be honest I don't even understand why we are special casing the C:\ drive in that code as its not required

GoaLitiuM: It's probably needed elsewhere but not in this case. I think all path related utility functions needs refactoring to use available `System.Path` methods instead of using string manipulation all over the place.

mafiesto4: Hmm strange, because `NormalizePath` fixes all path separators to be `/` except for `C:/` to `C:\`

wackoisgod: I guess my question is @mafiesto4 why does it special case the C drive? Like why is that required?

mafiesto4: okay let's merge this

PR #1510, opened 2023-09-24
Input event improvements

Swiggies:
This PR will allow developers to use a new InputEventState class to subscribe to input events which pass in the state of the input, where they can then handle what happens per-state, rather than having to make multiple input actions, and then creating multiple InputEvents.

This is pretty similar to my last input PR but just adds the functionality to InputEvent (through a new class).

Decided on making it a new C# class as I didn't want to break current functionality with InputEvents but open to suggestions.

Example usage:
```csharp
public InputEventState inputEvent = new InputEventState("Jump");

public override void OnEnable()
{
    // Here you can add code that needs to be called when script is enabled (eg. register for events)
    inputEvent.Triggered += OnTriggeredState;
}

private void OnTriggeredState(InputActionState state)
{
    if (state == InputActionState.Press) Debug.Log("Press");
    if (state == InputActionState.Release) Debug.Log("Release");
}

public override void OnDisable()
{
    // Here you can add code that needs to be called when script is disabled (eg. unregister from events)
    inputEvent.Triggered -= OnTriggeredState;
}
```

### DIFF
--- a/Source/Engine/Engine/InputEventState.cs
+++ b/Source/Engine/Engine/InputEventState.cs
@@ -1,0 +1,70 @@
+// Copyright (c) 2012-2023 Wojciech Figat. All rights reserved.
+
+using System;
+
+namespace FlaxEngine
+{
+    /// <summary>
+    /// Virtual input action binding. Helps with listening for a selected input event.
+    /// </summary>
+    public class InputEventState
+    {
+        /// <summary>
+        /// The name of the action to use. See <see cref="Input.ActionMappings"/>.
+        /// </summary>
+        [Tooltip("The name of the action to use.")]
+        public string Name;
+
+        /// <summary>
+        /// Returns true if the event has been triggered during the current frame (e.g. user pressed a key). Use <see cref="Triggered"/> to catch events without active waiting.
+        /// </summary>
+        public bool Active => Input.GetAction(Name);
+
+        /// <summary>
+        /// Occurs when event is triggered (e.g. user pressed a key). Called before scripts update.
+        /// Has an additional parameter to read the input state.
+        /// </summary>
+        public event Action<InputActionState> Triggered;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="InputEvent"/> class.
+        /// </summary>
+        public InputEventState()
+        {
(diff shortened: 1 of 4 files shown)


Tryibion:
Just a thought for discussion. Would it make sense to have a different event/Action per state?

Source/Engine/Engine/InputEventState.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <summary>
+        /// Returns true if the event has been triggered during the current frame (e.g. user pressed a key). Use <see cref="Triggered"/> to catch events without active waiting.
+        /// </summary>
+        public bool Active => Input.GetAction(Name);
+
+        /// <summary>
+        /// Occurs when event is triggered (e.g. user pressed a key). Called before scripts update.
+        /// Has an additional parameter to read the input state.
+        /// </summary>
+        public event Action<InputActionState> Triggered;
```
GoaLitiuM, line 27: So why can't this be under `InputEvent` as `public event Action<InputActionState> StateChanged;` or something like that? I don't think adding a new action would be a breaking change.

Swiggies: I can have a look at adding this, but normal InputEvents are handled the same way as regular InputActions and are only activated when in an exact state, the one that's specified in Input Settings. I initially tried to do this but I can have another look at adding it that way.

Swiggies:
> Just a thought for discussion. Would it make sense to have a different event/Action per state?

This is how Unity handles it, somewhat. I'm happy to implement that if it makes more sense to others.

mafiesto4:
@Swiggies very cool

Do you think it's better to use this version or #1496 ? I'm fine with this small API breakage.

Tryibion:
> @Swiggies very cool
> 
> Do you think it's better to use this version or #1496 ? I'm fine with this small API breakage.

I assume you mean #1529

mafiesto4:
Ah yeah, sorry my bad #1529

mafiesto4:
Thanks again @Swiggies - I merged code from @Tryibion to keep the API slicker (less types) - but I'll include you as a contributor in the changelog.

PR #1512, opened 2023-09-24
Fix error on try invoke actions by childrens on OnDestroy 

RuanLucasGD:
i needed redo this pr #1419


mafiesto4:
Maybe call `OnDestroy` for scripts after `OnDisable` but before actor's `OnEndPlay` so scripts will be disabled first, then e can start destruction?

RuanLucasGD:
@mafiesto4 It's right now?

PR #1513, opened 2023-09-24
Add showing only collision wireframe in `CollisionDataWindow`

RuanLucasGD:
This Remove background Sky from Collision Data Window, this help a lot when you don't have a mesh because collision data wire have a too light color with background.

Remade of #1413


mafiesto4:
I think it would be better to do sth like this:
```cs
_preview.Task.ViewFlags &= ~ViewFlags.Sky;
```

RuanLucasGD:
@mafiesto4 done

PR #1516, opened 2023-09-24
Add `GetRoot` method to actor for finding top most parent.

Tryibion:


wackoisgod:
So this code can cause a crash, when a script is put on to the "Scene" object because that will have no parent. Now its a good question if one should allow scripts on the Scene actor but at the moment you can and thus if you put something on it that calls this function it will crash the engine.

Tryibion:
> So this code can cause a crash, when a script is put on to the "Scene" object because that will have no parent. Now its a good question if one should allow scripts on the Scene actor but at the moment you can and thus if you put something on it that calls this function it will crash the engine.

Thanks for finding that mistake. Should be fixed now.

PR #1518, opened 2023-09-25
Add button to position camera to editor viewport view.

Tryibion:
This adds a button on the camera actor node context menu to quickly position the camera to the editor's viewport view.


mafiesto4:
Maybe override `OnContextMenu` method in `CameraNode` to implement this in cleaner way? We could pass `EditorWindow` there and check for `SceneTreeWindow` if used from main view (it could ignore `PrefabWindow` I guess).

Tryibion:
I agree that would be a cleaner way and opens up for more customizability in the actor nodes. Submitted.

PR #1520, opened 2023-09-25
Fix potential incorrect null checks in FlaxEngine.Objects

GoaLitiuM:
The null-conditional operator checks for reference equality of the `FlaxEngine.Object`, but doesn't check the validity of the unmanaged pointer. This check is corrected in cases where the object was not immediately returned from the bindings layer and may have been destroyed earlier.

Fixes https://github.com/FlaxEngine/FlaxEngine/issues/1328


Tryibion:
Unfortunately, I still get a crash when trying to play a scene animation.

Tryibion:
Looks like that recent change fixes the scene animation issue.

Issue #1114, opened 2023-05-22
Improvements to the Visual Graph


PhyresiCompany:
**Add Context Sensitive to Node Search**
Dragging a node shows only the possible nodes this parameter can connect to
![Captura de Tela (98)](https://github.com/FlaxEngine/FlaxEngine/assets/101017436/da3a1824-bacb-453c-9af9-0db7481e09ce)

mafiesto4:
Added in #1522 (in future we will keep improving it to provide smarter contextual options)

PR #1522, opened 2023-09-25
Visject: Context sensitive node search

Chikinsupu:
When dragging a connection out of a node port into the void, the context menu to add nodes gets shown.
Context sensitive node search makes it so that only nodes with compatible ports get shown in the context menu. This mimics the behaviour of visual editors from other engines, e.g. Unreal.

### Tasks
- [x] Implement context menu item filtering by type
- [x] Implement context menu group filtering by type
- [x] Support default archetypes and groups
- [x] Support particle emitter graph specific archetypes and groups
- [x] Support visual scripting specific archetypes and groups 
- [x] Support animation graph specific archetypes and groups
- [x] Add "context sensitive" toggle to context menu
- [x] Cleanup
- [x] Add XML comments
- [x] A lot of testing 
- [x] New tasks ahead
- [x] Rewrite input/output compatibility check for custom nodes
- [x] Cleanup 2
- [x] Testing 2
- [x] Profiling (against master)
- [x] Done 🥳
- [x] #1114 

**Known Bugs**
- [x] 'Comment' archetype shows up when typing in the search bar
- [x] Filtering by group name activates all items (turned off at the moment)
- [x] Asynchronously loaded nodes that aren't InvokeMethodNodes currently get filtered out even when compatible

**Optional**
- [x] Make context menu taller and less wide
- [x] Make item entries a tiny bit taller

**Didn't do**
- [x] Change entry score based on the port type (e.g. when dragging out of a vector3 port, options that have Vector3 as input should have a higher score than Float3, Vector2, int etc) -> Will probably be its own PR

### Profiling Results
I didn't do craaaazzy profiling and testing. But enough to get an idea what the costs of this feature are.
I used the in engine profiler to check how much longer in time the opening and fetching of the items take, and how much more memory is getting used. Also these times are just the spikes visible in the profiler, since custom nodes get fetched asynchronously, so the time for all items to appear varies.

First of all, when right clicking into the void, so when no context is really needed, the speeds are rather identical. 
Opening the CM took in both cases around **85ms** on average while taking up **~4MB of memory**

Now, the more interesting part: dragging a connection out of a port. Since here context sensitivity really just starts doing work.

On master with no context sensitivity it took around **85ms** on average while taking up **~4MB of memory**, what a surprise
And on my branch with context sensitivity enabled it took around **116ms** while taking up **~7MB of memory**

I think it comes to no surprise that filtering through all archetypes does take some more time. With an average of around 30ms more, which means it's effectively ~35% slower now and takes up 3MB of memory more.

Like i said, i didn't do crazy profiling here, i just read numbers from the in engine profiler. Let's just say i was a bit lazy to open dotTrace now and do in depth profiling.
Also the whole filtering process could probably be a lot more optimized and faster, but would need some refactoring.
I think sacrificing a couple of milliseconds more for such a QoL feature is okay, when opening a context menu. Especially since it can still be optimized.

Just keep in mind that my PC is on the stronger side, with an Ryzen 9 3900X, 32GB RAM etc. But i think even weaker PCs won't have a problem with that.

### Feature preview

Material Editor
ContextualNodes

Visual Scripting Editor
ContextSensitiveVisualScripting

Context Menu Toggle
ContextSensitiveToggle

### DIFF
--- a/Source/Editor/Scripting/ScriptType.cs
+++ b/Source/Editor/Scripting/ScriptType.cs
@@ -1393,5 +1393,32 @@ namespace FlaxEditor.Scripting
                 return _custom.GetMembers(name, MemberTypes.Method, bindingAttr).FirstOrDefault();
             return ScriptMemberInfo.Null;
         }
+        
+        /// <summary>
+        /// Basic check to see if a type could be casted to another type
+        /// </summary>
+        /// <param name="from">Source type</param>
+        /// <param name="to">Target type</param>
+        /// <returns>True if the type can be casted</returns>
+        public static bool CanCast(ScriptType from, ScriptType to)
+        {
+            if (from == to)
+                return true;
+            if (from == Null || to == Null)
+                return false;
+            return (from.Type != typeof(void) && from.Type != typeof(FlaxEngine.Object)) &&
+                   (to.Type != typeof(void) && to.Type != typeof(FlaxEngine.Object)) &&
+                   from.IsAssignableFrom(to);
+        }
+
+        /// <summary>
+        /// Basic check to see if this type could be casted to another type
+        /// </summary>
+        /// <param name="to">Target type</param>
+        /// <returns>True if the type can be casted</returns>
+        public bool CanCastTo(ScriptType to)
+        {
+            return CanCast(this, to);
+        }
     }
 }
--- a/Source/Editor/Surface/Archetypes/Function.cs
+++ b/Source/Editor/Surface/Archetypes/Function.cs
@@ -773,7 +773,7 @@ namespace FlaxEditor.Surface.Archetypes
                     Values[4] = GetSignatureData(memberInfo, memberInfo.GetParameters());
                 }
             }
-
+            
             private SignatureInfo LoadSignature()
             {
                 var signature = new SignatureInfo();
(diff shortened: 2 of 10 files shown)


mafiesto4:
@Chikinsupu How ready is this? Do you need another testing week or can we merge this week?

@stefnotch Would you mind taking a look and review?

Chikinsupu:
>@Chikinsupu How ready is this? Do you need another testing week or can we merge this week?
>
>@stefnotch Would you mind taking a look and review?

I tested all kinds of graphs. Material, Particle, Visual Scripting and Animation and i didn't encounter any more issues after my last tests. So i would say it does work generally, or at least it won't break anything.
If needed i can test for another week of course and check for edge cases.

We just have to keep in mind that this is a rather stupid context filter, since all it does as of now, is to check if an archetype is compatible with the current port type. No more, no less.

I think a more interesting question would be, if this is sufficient for now and can be updated over time, or if this feature needs some additional functionality.
Examples:

- Right now it doesn't update item scores, meaning that nodes that would make more sense in the given context get pushed up the list. So, right now when you have a port with a Vector3 type, and drag a connection out of there, it would still show float2, float3, float4, double2, double3, double4, vector2 methods before the vector3 methods.

- Automatic casting is not implemented. Right now a ton of nodes are valid because the underlying system thinks that a type can be casted to another type. When you drag a connection out of an 'Actor' port, it shows you a ton of items that will not connect after spawning the node, because the Actor needs to be casted to another type first. This was already commented in the code, prior to this PR, just not implemented and instead it throws an exception right now.

So, depending on if this PR should fix things like the 2 mentioned examples, it's ready or not.
I was thinking of making 2 different PRs for automatic casting, and smarter sorting, but fixing those things in this PR is fine for me as well.

Any thoughts?

stefnotch:
I only took a quick glance at the code and haven't thoroughly tested this. So here are my preliminary 2 cents:
The code does look well written, finding bugs would probably be a matter of testing it.
In terms of features, it's definitely sufficient for now. I'd love to see the smarter sorting and automatic casting being implemented in future PRs.

mafiesto4:
Thanks!

PR #1529, opened 2023-09-26
Add input states to `InputEvent` and `AxisChanged` to `InputAxis`

Tryibion:
This is just a more compact version of #1510. This also adds an `AxisChanged` event to the `InputAxis` class.

### DIFF
--- a/Source/Engine/Input/Input.h
+++ b/Source/Engine/Input/Input.h
@@ -296,6 +296,12 @@ public:
     API_EVENT() static Delegate<StringView, const InputActionState&> ActionTriggered;
 
     /// <summary>
+    /// Event fired when virtual input axis is changed. Called before scripts update. See <see cref="AxisMappings"/> to edit configuration.
+    /// </summary>
+    /// <seealso cref="InputAxis"/>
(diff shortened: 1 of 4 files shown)


Source/Engine/Input/Input.h
```diff
@@ -293,7 +293,13 @@ API_CLASS(Static) class FLAXENGINE_API Input
     /// Event fired when virtual input action is triggered. Called before scripts update. See <see cref="ActionMappings"/> to edit configuration.
     /// </summary>
     /// <seealso cref="InputEvent"/>
-    API_EVENT() static Delegate<StringView> ActionTriggered;
+    API_EVENT() static Delegate<StringView, const InputActionState&> ActionTriggered;
```
mafiesto4, line 296: This can be just `InputActionState` as it's enum value

Tryibion: Like instead of `const InputActionStat&`?

mafiesto4: Yea, we would pass just value of enum instead of const reference. Not a big deal, just a minor suggestion for readability fo this event in C++.

PR #1533, opened 2023-09-27
Use Guids for sprite identification instead of indexing

Arcnor:
Atlas sprites are currently being identified using an index. Problem is, modifying the atlas list of sprites can invalidate all used sprites, or use the wrong ones, etc. Worse still, this happens without any error or warning, and the user won't even notice until they save and load their project again (because `SpriteHandle` instances are cached, so the editor will still show the correct Sprites even when their index has changed).

This PR makes Sprites use a `Guid` for identification instead.

Various things:
1. Missing is a way of processing existing projects so indices can be replaced with generated `Guid`s. There is a warning that shows up when a project is opened with a newer version of the editor, and I'm assuming something runs to update the project. I'm assuming that mechanism can be used to do this. Otherwise this will be a breaking change...
2. The `Array<Sprite> Sprites` inside `SpriteAtlas` has been replaced with a `Dictionary<Guid, Sprite> Sprites` instead. This has obvious performance implications when `Draw`ing (at least), but the tradeoff of possibly using the wrong index is worse in my opinion. Caching layers or passing `Sprite` references instead might improve this, but benchmarking should probably be applied before fixing what might not be a real issue.

### DIFF
--- a/Source/Editor/CustomEditors/Editors/SpriteHandleEditor.cs
+++ b/Source/Editor/CustomEditors/Editors/SpriteHandleEditor.cs
@@ -1,5 +1,7 @@
 // Copyright (c) 2012-2023 Wojciech Figat. All rights reserved.
 
+using System;
+using System.Linq;
 using FlaxEditor.GUI;
 using FlaxEditor.Scripting;
 using FlaxEngine;
@@ -15,7 +17,8 @@ namespace FlaxEditor.CustomEditors.Editors
     {
         private ComboBox _spritePicker;
         private ValueContainer _atlasValues;
-        private ValueContainer _indexValues;
+        private ValueContainer _idValues;
+        private Guid[] _guids;
 
         /// <inheritdoc />
         public override void Initialize(LayoutElementsContainer layout)
@@ -27,8 +30,8 @@ namespace FlaxEditor.CustomEditors.Editors
             _atlasValues = atlasValues;
 
             // Sprite
-            var spriteIndexField = typeof(SpriteHandle).GetField("Index");
-            _indexValues = new ValueContainer(new ScriptMemberInfo(spriteIndexField), Values);
+            var spriteIdField = typeof(SpriteHandle).GetField("Id");
+            _idValues = new ValueContainer(new ScriptMemberInfo(spriteIdField), Values);
             var spriteLabel = layout.AddPropertyItem("Sprite", "The selected sprite from the atlas.");
 
             // Check state
@@ -54,10 +57,20 @@ namespace FlaxEditor.CustomEditors.Editors
             var spritesCount = value.SpritesCount;
             var spritePicker = spriteLabel.ComboBox();
             spritePicker.ComboBox.Items.Capacity = spritesCount;
-            for (int i = 0; i < spritesCount; i++)
-            {
-                spritePicker.ComboBox.AddItem(value.GetSprite(i).Name);
+            _guids = new Guid[spritesCount];
+
+            var sortedSprites = value.Sprites.Values.ToList();
+            sortedSprites.Sort((s0, s1) => {
+                var xComp = s0.Area.Location.X.CompareTo(s1.Area.Location.X);
+
+                return xComp == 0 ? s0.Area.Location.Y.CompareTo(s1.Area.Location.Y) : xComp;
+            });
+            var idx = 0;
+            foreach (var sprite in sortedSprites) {
+                spritePicker.ComboBox.AddItem(sprite.Name);
+                _guids[idx++] = sprite.Id;
             }
+
             spritePicker.ComboBox.SupportMultiSelect = false;
             spritePicker.ComboBox.SelectedIndexChanged += OnSelectedIndexChanged;
             _spritePicker = spritePicker.ComboBox;
@@ -70,7 +83,7 @@ namespace FlaxEditor.CustomEditors.Editors
 
             SpriteHandle value;
             value.Atlas = (SpriteAtlas)_atlasValues[0];
-            value.Index = _spritePicker.SelectedIndex;
+            value.Id = _guids[_spritePicker.SelectedIndex];
             SetValue(value);
         }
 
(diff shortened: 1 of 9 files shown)


Source/Editor/CustomEditors/Editors/SpriteHandleEditor.cs
```diff
@@ -54,10 +57,20 @@ public override void Initialize(LayoutElementsContainer layout)
...
-            for (int i = 0; i < spritesCount; i++)
-            {
-                spritePicker.ComboBox.AddItem(value.GetSprite(i).Name);
+            _guids = new Guid[spritesCount];
+
+            var sortedSprites = value.Sprites.Values.ToList();
+            sortedSprites.Sort((s0, s1) => {
+                var xComp = s0.Area.Location.X.CompareTo(s1.Area.Location.X);
+
+                return xComp == 0 ? s0.Area.Location.Y.CompareTo(s1.Area.Location.Y) : xComp;
```
Arcnor, line 66: This sorting is more or less duplicated here and in the `SpriteAtlasWindow`, but I have a PR in the works that completely replaces this editor anyway, so I decided not to spend a lot of time in refactors for now.

Source/Editor/Windows/Assets/SpriteAtlasWindow.cs
```diff
@@ -291,6 +295,9 @@ public override void Save()
 
             ClearEditedFlag();
             _item.RefreshThumbnail();
+
+            _properties.UpdateSprites();
+            _propertiesEditor.BuildLayout();
```
Arcnor, line 300: This is actually added for a separate issue the editor has: When saving the list of sprites, names were not being changed and new ordering was not being applied. This fixes the inconsistency.

Source/Engine/Render2D/SpriteAtlas.cpp
```diff
@@ -153,11 +159,26 @@ bool SpriteAtlas::LoadSprites(ReadStream& stream)
...
+        if (tilesVersion >= 2)
+        {
+            stream.ReadUint32(&t.Id.Values[0]);
+            stream.ReadUint32(&t.Id.Values[1]);
+            stream.ReadUint32(&t.Id.Values[2]);
+            stream.ReadUint32(&t.Id.Values[3]);
+        }
+        else
+        {
+            t.Id = Guid::New();
```
Arcnor, line 178: This should be changed according to the aforementioned missing "project upgrading" processing. Before/after creating a new id for older format `SpriteAtlas` instances, replacements should be made where needed (i.e. `SpriteRender` instances).

Arcnor:
Forgot to mention: Sprites are no longer sorted by insertion order (because that's now undefined by using a `Dictionary`), but by X/Y position.

This can be "fixed" in at least 2 ways:
1. Keeping track of insertion order by adding another "index" field to the Sprite itself, but this needs handling when sprites are added/removed, which is error prone.
2. Replacing the `Dictionary` with an array, so sorting is "natural". This will impact Sprite instances fetching, though.

SinnersSum:
You should also expose an indexer in the form of `GetSpriteAt(int idx)` or something like that. I imagine in certain cases (like enumerating the atlas) it is still far more practical to get by index rather than id.

Arcnor:
> You should still expose an indexer in the form of `GetSpriteAt(int idx)` or something like that. In certain cases (like enumerating the atlas) it is still far more practical to get by index rather than id.

There was one like that, but because of the `Dictionary` change, there is only `GetSprite(Guid)` now. Enumeration is done simply by iterating over the dictionary itself, right now.

SinnersSum:
> > You should still expose an indexer in the form of `GetSpriteAt(int idx)` or something like that. In certain cases (like enumerating the atlas) it is still far more practical to get by index rather than id.
> 
> There was one like that, but because of the `Dictionary` change, there is only `GetSprite(Guid)` now. Enumeration is done simply by iterating over the dictionary itself, right now.

I'm just thinking how's that going to work on the C# side though? I see that the docs mention a [Sprites](https://docs.flaxengine.com/api/FlaxEngine.SpriteAtlas.html#FlaxEngine_SpriteAtlas_Sprites) property, is that still generated in your revision? I guess that's a workable solution if so

Arcnor:
> > > You should still expose an indexer in the form of `GetSpriteAt(int idx)` or something like that. In certain cases (like enumerating the atlas) it is still far more practical to get by index rather than id.
> > 
> > There was one like that, but because of the `Dictionary` change, there is only `GetSprite(Guid)` now. Enumeration is done simply by iterating over the dictionary itself, right now.
> 
> I'm just thinking how's that going to work on the C# side though? I see that the docs mention a [Sprites](https://docs.flaxengine.com/api/FlaxEngine.SpriteAtlas.html#FlaxEngine_SpriteAtlas_Sprites) property, is that still generated in your revision? I guess that's a workable solution if so

If you check my changes you'll see the Sprites Dictionary is already being used to iterate. It's not the ideal API, having `AddSprite`, `RemoveSprite`, then direct access to the underlying `Dictionary`, but that field was already there, so I'm assuming it's used in other places.

Ideally we'd expose a read-only iterator instead, but I haven't found anything like that exposed from the C++ side anywhere.

SinnersSum:
> If you check my changes you'll see the Sprites Dictionary is already being used to iterate. It's not the ideal API, having `AddSprite`, `RemoveSprite`, then direct access to the underlying `Dictionary`, but that field was already there, so I'm assuming it's used in other places.
> 
> Ideally we'd expose a read-only iterator instead, but I haven't found anything like that exposed from the C++ side anywhere.

Ah I see, you actually replaced the underlying array with a dictionary, I didn't notice that in my original reading.

My only concern about this change is what it means for the C# side's ability to enumerate; it seems that originally the sprites array was getting marshalled directly as an array on the C# side; this raises the question whether the dictionary get exposed as a property on the C# side as well? Impressive work on the part of the marshaller if so, but if not it's probably worth adding a pair of functions that take/produce an array to reproduce the functionality of [Sprite](https://docs.flaxengine.com/api/FlaxEngine.SpriteAtlas.html#FlaxEngine_SpriteAtlas_Sprites).

Arcnor:
> > If you check my changes you'll see the Sprites Dictionary is already being used to iterate. It's not the ideal API, having `AddSprite`, `RemoveSprite`, then direct access to the underlying `Dictionary`, but that field was already there, so I'm assuming it's used in other places.
> > Ideally we'd expose a read-only iterator instead, but I haven't found anything like that exposed from the C++ side anywhere.
> 
> Ah I see, you actually replaced the underlying array with a dictionary, I didn't notice that in my original reading.
> 
> My only concern about this change is what it means for the C# api's ability to enumerate; it seems that originally the sprites array was getting marshalled directly as an array to the C# api; this raises the question whether the dictionary get exposed as a property on the C# side as well? Impressive work on the part of the marshaller if so, but if not it's probably worth adding a pair of functions that take/produce an array to reproduce the functionality of [Sprite](https://docs.flaxengine.com/api/FlaxEngine.SpriteAtlas.html#FlaxEngine_SpriteAtlas_Sprites).

If you read the code, you'll see that the C# side has access to the Sprite Dictionary, which again is what I'm using to do the enumeration in at least one place. That comes from the C++ side, so the whole marshalling process works just fine.

Issue #1535, opened 2023-09-27
Flax Editor Rebuilds csproj File, Causing Loss of Solution-Wide Analyze Data in Rider


schmidt-florian:
**Issue Description**
When the Flax Editor rebuilds the .csproj file, Rider loses all the solution-wide analyze data. This forces Rider to re-analyze all files, which is a time-consuming process, especially for large projects.

This also happens with Visual Studio and Resharper.

**Steps to Reproduce**
Open a Flax Engine solution in Rider.
Perform some actions in Rider that trigger a .csproj file rebuild (e.g., adding a new script).
Switch back to Flax Editor. (sometimes not needed because flax will try to rebuild in the background)
Switch back to Rider.

**Expected Behavior**
Rider should retain the solution-wide analyze data even after the .csproj file is modified by the Flax Editor.

**Actual Behavior**
Rider loses all solution-wide analyze data and starts re-analyzing all files, which takes a significant amount of time for large projects.

**System Information**
Flax Engine Version: master branch 248304a
Rider Version: 2023.2.2

mafiesto4:
Fixed in #1538

PR #1538, opened 2023-09-27
Always generate additional Visual Studio solution files for all platforms

GoaLitiuM:
Fixes ~~# 1537~~ (sorry wrong issue) and possibly other Intellisense issues regarding VSCode, especially in non-Windows platforms where VSCode workspace files are always generated with default `-genproject` options.


GoaLitiuM:
Fixes https://github.com/FlaxEngine/FlaxEngine/issues/1535

PR #1541, opened 2023-09-28
Better light theme (Style) support, added a 2nd Default (light) theme

Radiangames:
Three changes to make light themes/Styles work better and to include one by default.

1) Added ForegroundViewport as a new color.  It is used in the main game viewport (ViewportWidgetButton), and the viewport for rendering of particles and materials.  It is needed because the default foreground in a Light theme is black, but black does not work well in a viewport.  A new color seemed appropriate.

2) Fixed the profiler window to use the Foreground color in multiple text elements, instead of Color.White (or no default TitleColor).  This includes  the Row class, Asset class, SingleChart class, Timeline Class, and more.

3) Added a second theme/Style (DefaultLight) to include with the engine.  It uses RGB float values because those were easier to transfer from the saved values that I had created (and they're easier for me to edit if necessary).  I tried to emulate how the Default theme is created/loaded/etc as closely as possible.

### DIFF
--- a/Source/Editor/Content/Tree/ContentTreeNode.cs
+++ b/Source/Editor/Content/Tree/ContentTreeNode.cs
@@ -86,6 +86,7 @@ namespace FlaxEditor.Content
                 Folder.ParentFolder = parent.Folder;
                 Parent = parent;
             }
+            IconColor = Style.Current.Foreground;
         }
 
         /// <summary>
--- a/Source/Editor/GUI/AssetPicker.cs
+++ b/Source/Editor/GUI/AssetPicker.cs
@@ -395,8 +395,8 @@ namespace FlaxEditor.GUI
             else
             {
                 // No element selected
-                Render2D.FillRectangle(iconRect, new Color(0.2f));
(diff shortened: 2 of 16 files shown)


MineBill:
Hey, i really like this! The default light theme looks so good!
I'd like to point out some areas i found that need some more work:

The "No asset select" background is a bit transparent, so on a white background there is not a lot of contrast to read the message.
image

Inspectors for arrays and dictionaries still use a dark background color.
image
image

The icons on the Game Cooker window are very dark.
image

The folder icons on the tree view of the content window are white. Not sure if you can just change the color of icon from code or a different icon set is needed for this.
image

Again, thank you for this!

Radiangames:
> Hey, i really like this! The default light theme looks so good! I'd like to point out some areas i found that need some more work:

Hi MineBill, thanks for finding these cases and including pics!  I will update the branch with fixes for these a little later today.  Will again try to keep the code changes as minimal as possible.

Cheers,
Luke

Radiangames:
> I'd like to point out some areas i found that need some more work:

I believe I have fixed all the issues.  Please let me know if I missed anything.  

Thanks again for the help in improving light theme support!

mafiesto4:
This looks nice! But hurts eyes out of brightness xD

PR #1511, opened 2023-09-24
Fix mouse leave gameWindow

RuanLucasGD:
This is a usability tweak, with this i can easily edit editor properties in game mode even if Screen.CursorVisible is false by any script

This fix #1370  and is a small change from #859

Issue #1370, opened 2023-09-08
Show mouse on all windows except Game Window


RuanLucasGD:
**Issue description:**
Hello, this is a suggestion about usability. I'm making a game where i need the Screen.Visible = false, but Screen.CursorLock = CursorLockMode.None. I made this because i made a custom cursor icon.

The problem happen when i need change any value on editor on game mode why i cannot see the cursor XD.
Please , if possible set cursor as visible (in all windows except GameWindow) always on game mode., this can help a lot.

**Flax version:**
1.6

mafiesto4:
Fixed in #859 and f2a3338dce85b2b9a6fb9c7957349b27fdf92e72

PR #1544, opened 2023-09-28
Fix mouse leave gameWindow

RuanLucasGD:
#1511


mafiesto4:
Actually, this should check if the game control is actually focused before changing focus because if user presses `Shift+F11` to defocus the game window then the window focuses again when moving cursor outside it.

RuanLucasGD:
I tried ```Shift+F11``` and this fix my problem on #1370 . I think this pr is not necessary

PR #1553, opened 2023-09-28
Automatically set SpriteRender size when changing either its Texture or Sprite

Arcnor:
When assigning a new texture or sprite to a `SpriteRender` element, the size needs to be set by hand, which is really annoying. This PR simply sets the right size after a change of Texture/Sprite has been made.


Arcnor:
Changing the material or other parameter also changes the size, I need to find a way to differentiate the "event".

Issue #1552, opened 2023-09-28
Visual Scripting: Get LengthSquared (and other nodes) throw exception and return null


Chikinsupu:
**Issue description:**

Depending on the version:

Version 1.6:
Using Get LengthSquared, Get Length, Distance node, and probably many other nodes of the same kind on vector3 throws Failed to call Method exception.

> [ 00:00:07.988 ]: [Error] Failed to call method 'FlaxEngine.Vector3.get_LengthSquared' (args count: 0) with invalid object instance of type 'FlaxEngine.Float3'
> [ 00:00:07.988 ]: [Error] Visual Script stack trace:
> [ 00:00:07.988 ]: [Error]     at Culprit:b41ab51744e77610f0ff009855f0b55b in node FlaxEngine.Vector3.get_LengthSquared()
>     at Culprit:b41ab51744e77610f0ff009855f0b55b in node FlaxEngine.Debug.Log()
>     at Culprit:b41ab51744e77610f0ff009855f0b55b in node OnEnable()
> 

Version Master:
Using Get LengthSquared, Get Length, Distance node, and probably many other nodes of the same kind on vector3 throws native interop exception: Invalid ManagedHandle exception.

> 
> [ 00:00:17.712 ]: [Error] Native interop exception!
> [ 00:00:17.725 ]: [Warning] Exception has been thrown during InvokeMethod. Invalid ManagedHandle
> Stack strace:
>    at FlaxEngine.Interop.ManagedHandle.ManagedHandlePool.GetObject(IntPtr handle) in D:\a\FlaxEngine\FlaxEngine\Source\Engine\Engine\NativeInterop.Managed.cs:line 537
>    at FlaxEngine.Interop.NativeInterop.Invoker.InvokerRet0`2.MarshalAndInvoke(Object delegateContext, ManagedHandle instancePtr, IntPtr paramPtr) in D:\a\FlaxEngine\FlaxEngine\Source\Engine\Engine\NativeInterop.Invoker.cs:line 841
>    at FlaxEngine.Interop.NativeInterop.InvokeMethod(ManagedHandle instanceHandle, ManagedHandle methodHandle, IntPtr paramPtr, IntPtr exceptionPtr) in D:\a\FlaxEngine\FlaxEngine\Source\Engine\Engine\NativeInterop.Unmanaged.cs:line 704
> [ 00:00:17.725 ]: [Error] Exception has been thrown during InvokeMethod.
> Invalid ManagedHandle
[...]

PR #1555, opened 2023-09-28
Fix VisualScript method calls not working with value types

GoaLitiuM:


GoaLitiuM:
Fixes https://github.com/FlaxEngine/FlaxEngine/issues/1552

PR #1561, opened 2023-09-29
Added support for in parameters to the csharp binding generator

SinnersSum:
`in` parameters allow for large value type objects to be passed by reference with semantics very similar to `const &`, such that `r-values` can be passed seamlessly as well without needing to create a variable for the value. Functions such as [Camera.GetMatrices](https://github.com/FlaxEngine/FlaxEngine/blob/fd3f10864b8c09289971d8cb2cfd64090d0cd026/Source/Engine/Level/Actors/Camera.h#L219) or [Camera.Project](https://github.com/FlaxEngine/FlaxEngine/blob/fd3f10864b8c09289971d8cb2cfd64090d0cd026/Source/Engine/Level/Actors/Camera.h#L151) currently receive certain arguments by-ref, which in many cases requires the caller to explicitly create an `l-value` variable for the argument in order to pass it in by `ref`. If these functions instead used `in` values they could enjoy all of the advantages of passing by-ref without needing an explicit `l-value` to reference.

I have run some light tests of this change in a small project, and it appears to work correctly. As yet I haven't added any uses of `API_PARAM(In)` to the engine, but I will attempt to hit some of these obvious cases tomorrow,

### DIFF
--- a/Source/Editor/CustomEditors/CustomEditor.cs
+++ b/Source/Editor/CustomEditors/CustomEditor.cs
@@ -637,7 +637,7 @@ namespace FlaxEditor.CustomEditors
                 if (text.Length != 32)
                     return false;
                 JsonSerializer.ParseID(text, out var id);
-                obj = FlaxEngine.Object.Find<FlaxEngine.Object>(ref id);
+                obj = FlaxEngine.Object.Find<FlaxEngine.Object>(in id);
             }
             else if (Color.TryParseHex(text, out var color))
             {
--- a/Source/Editor/CustomEditors/Dedicated/ActorEditor.cs
+++ b/Source/Editor/CustomEditors/Dedicated/ActorEditor.cs
@@ -58,7 +58,7 @@ namespace FlaxEditor.CustomEditors.Dedicated
                 if (prefab && !prefab.WaitForLoaded())
                 {
                     var prefabObjectId = actor.PrefabObjectID;
-                    var prefabInstance = prefab.GetDefaultInstance(ref prefabObjectId);
+                    var prefabInstance = prefab.GetDefaultInstance(in prefabObjectId);
                     if (prefabInstance != null)
                     {
                         // Use default prefab instance as a reference for the editor
@@ -168,7 +168,7 @@ namespace FlaxEditor.CustomEditors.Dedicated
                 {
                     var actor = (Actor)Values[0];
                     var prefabObjectId = actor.PrefabObjectID;
-                    var prefabInstance = prefab.GetDefaultInstance(ref prefabObjectId);
+                    var prefabInstance = prefab.GetDefaultInstance(in prefabObjectId);
                     if (prefabInstance != null)
                     {
                         Values.SetReferenceValue(prefabInstance);
--- a/Source/Editor/CustomEditors/Editors/FlaxObjectRefEditor.cs
+++ b/Source/Editor/CustomEditors/Editors/FlaxObjectRefEditor.cs
@@ -114,7 +114,7 @@ namespace FlaxEditor.CustomEditors.Editors
         public Guid ValueID
         {
             get => _value ? _value.ID : Guid.Empty;
-            set => Value = Object.Find<Object>(ref value);
+            set => Value = Object.Find<Object>(in value);
         }
 
         /// <summary>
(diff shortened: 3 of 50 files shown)


SinnersSum:
First time creating a pr so let me know if I've committed any faux pas.

Changing library functions to take `in` instead of `ref` will be a minor breaking change.

I'll note that I wasn't completely sure what some of the checks to `ParameterInfo.IsRef` were trying to accomplish so I kept my modifications relatively minimal. When I attempt to add `in` support to some library functions tomorrow I'm sure I'll get a better idea whether there will be any issues.

mafiesto4:
You might want to update `BindingsGenerator.CacheVersion` to reflect format change of serialized bindings info for build system.

SinnersSum:
> You might want to update `BindingsGenerator.CacheVersion` to reflect format change of serialized bindings info for build system.

Updated the cache version. Thanks for noting that, I would have totally missed that detail.

Made a couple of minor changes, namely changed `FunctionInfo`/`ParameterInfo` to use bit-flags instead of bool fields, as a TODO comment in the code had requested.


More majorly though, I changed a couple of apis —namely `Object` and `Model`/`ModelLod`/`Mesh`— to use `in` instead of `ref`. I was planning on updating the `Matrix` api (which has a lot of candidate parameters) but decided to hold off on that for now as the candidate functions have a ton of references.

This does constitute a breaking change, but all that needs to be done to update affected callsites is swap `ref` for `in`. Long term this should improve ergonomics, as you will no longer need to assign a value to a temp variable before passing it in, which I noticed many callsites in the engine do.

SinnersSum:
I'd like to get a review. I expect the tests (at least the showcase projects) will fail to compile, but the engine should compile and run without issue.

If it's not possible to merge the engine api changes, then I'd like to get just the support for in-refs in `Flax.Build` merged so that afterwards new apis can begin to freely use in-refs.

SinnersSum:
Wow, I'm glad to see that all the checks passed.

> 1. Add `FlaxEngine.Object.Find<T>(ref id)` marked as deprecated so users can migrate easily

I agree that would be preferable, but the issue is that `in` and `ref` are the same under the hood (only differentiated by an attribute that only the compiler sees IIRC) so `Find(in GUID)` and `Find(ref GUID)` are the same function as far as .Net is concerned, and thus can't coexist.

I can see two options to get around this;

1) Create `Find(GUID)` (non-ref), mark `Find(ref GUID)` deprecated, and then at some point in the future add `in` to the new `Find(GUID)` function and remove the other. Not great, since it would imply a performance hit in the interm.

2) Create a temporary `Find_(in GUID)` function (name pending), mark the old one as deprecated, then at a later date change `ref` to `in` in the original, and then at a *later later date* possibly remove the unnecessary temporary function. Workable, but awkward.

> 2. Add or just verify that generated C++ bindings code doesn't convert the value back to C# data - for `T&` we might convert custom structure from C# to C++ data, perform API call and then back from C++ to C#. For `in` params we could skip conversion back.

I'll give it a look. I tried to avoid changing anything significant about the logic in order to avoid breaking anything, but I added the `IsByRefIn` and `IsByRefOut` properties for just that purpose so I'll see if there're some places where I can elide marshalling.

> 3. I would skip Flax 1.7 as it's pretty late in the production schedule so this could be merged for `1.8` (still target master but merge would happen in 2-3 weeks closest)

Makes sense 👍🏿

mafiesto4:
We gonna need to do perf comparision between ref and in for bindings methods like FindObject(id) to be sure there is no perf regression (I would expect improvements as otherwise it's not so important feature if usability only for such drastic api change). And maybe leaving ref version of existing methods should be implement too so existing code works properly.

SinnersSum:
> We gonna need to do perf comparision between ref and in for bindings methods like FindObject(id) to be sure there is no perf regression (I would expect improvements as otherwise it's not so important feature if usability only for such drastic api change). And maybe leaving ref version of existing methods should be implement too so existing code works properly.

I'm waiting for C# 12 to finish this PR, as with C# 12 changing apis to use `in` will no longer be a breaking change, but only just a warning. https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/ref-readonly-parameters.md

In terms of perf though, for maximum effect it will be necessary to add `readonly` annotation to struct members where appropriate. It might be possible to automate the generation of `readonly` annotation from C++ `const` annotation, and this wouldn't represent a breaking change either way. I'll investigate this afterwards.

PR #1562, opened 2023-09-29
Fix build project generation under Arch linux

Zode:
Just a quick patch to stop the "Exception: Missing NET SDK 7.0" error while generating project files under a situation where the dotnet environment path exists and is defined, but the dotnet pack is named differently (such as Microsoft.NETCORE.App.Host.arch-x64 vs Microsoft.NETCORE.App.Host.linux-x64). In such case the ridFallback would never get filled and thus fail later on with the TryAddHostRuntime function call resulting in the missing sdk error.


aidar31:
Thanks!!!, but error "Missing Android SDK. Cannot build for Android Plaform" remain. I installed android sdk with command "sudo apt install android-sdk".

PR #1564, opened 2023-09-29
Add VisibleIfValue attribute

Arcnor:
The `VisibleIf` attribute allows you to hide fields in the editor for your type based on another `bool` field without having to create a custom editor.

I've expanded support and added attributes for `Enum` and `int`. This can be generalized more, for example, having a single `VisibleIfValue` attribute where you pass the type of the field you're depending on, or even ignoring the type of the field altogether, just comparing its value, instead of having `VisibleIfEnum`, `VisibleIfInt`, etc... I've just followed the existing pattern for now but will be nice hearing some feedback about it.

### DIFF
--- a/Source/Editor/CustomEditors/Editors/GenericEditor.cs
+++ b/Source/Editor/CustomEditors/Editors/GenericEditor.cs
@@ -71,6 +71,11 @@ namespace FlaxEditor.CustomEditors.Editors
             public VisibleIfAttribute VisibleIf;
 
             /// <summary>
+            /// The "visible if value" attribute.
+            /// </summary>
+            public VisibleIfValueAttribute VisibleIfValue;
+
+            /// <summary>
             /// The read-only attribute usage flag.
             /// </summary>
             public bool IsReadOnly;
@@ -129,6 +134,7 @@ namespace FlaxEditor.CustomEditors.Editors
                 Space = (SpaceAttribute)attributes.FirstOrDefault(x => x is SpaceAttribute);
                 Header = (HeaderAttribute)attributes.FirstOrDefault(x => x is HeaderAttribute);
                 VisibleIf = (VisibleIfAttribute)attributes.FirstOrDefault(x => x is VisibleIfAttribute);
+                VisibleIfValue = (VisibleIfValueAttribute)attributes.FirstOrDefault(x => x is VisibleIfValueAttribute);
                 IsReadOnly = attributes.FirstOrDefault(x => x is ReadOnlyAttribute) != null;
                 ExpandGroups = attributes.FirstOrDefault(x => x is ExpandGroupsAttribute) != null;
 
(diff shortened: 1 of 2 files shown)


stefnotch:
Alternative proposal:

Update the VisibleIf documentation with an example along the lines of
```csharp
[VisibleIf(nameof(_isCustomResolution )), Limit(1, 8192)]
public int Width = 1280;

private bool _isCustomResolution => Resolution == ResolutionModes.Custom;
```

That's what the Flax editor does internally. https://github.com/FlaxEngine/FlaxEngine/blob/fd3f10864b8c09289971d8cb2cfd64090d0cd026/Source/Editor/Windows/Assets/SceneAnimationWindow.cs#L124C13-L129C37

Arcnor:
Like @stefnotch said, `VisibleIf` can evaluate a property, so potentially you can add the necessary code to do most checks you might want. However, I think you might want to check for visibility based on a single value of some field, like (in my own case) testing for Enum values. Hence I've modified my PR to support the generic case above (`VisibleIfValue` instead of different `VisibleIf<Type>`). That means that instead of doing:

```
enum Mode { ModeA, ModeB, ModeC }

public Mode _mode;

public bool IsModeA => _mode == Mode.ModeA;
public bool IsModeB => _mode == Mode.ModeB;
public bool IsModeC => _mode == Mode.ModeC;

[VisibleIf(nameof(IsModeA))]
public object ThingsToShowWhenA;

[VisibleIf(nameof(IsModeB))]
public object ThingsToShowWhenB;

[VisibleIf(nameof(IsModeC))]
public object ThingsToShowWhenC;
```
...you should be able to do:
```
enum Mode { ModeA, ModeB, ModeC }

public Mode _mode;

[VisibleIfValue(nameof(Mode), Mode.ModeA)]
public object ThingsToShowWhenA;

[VisibleIfValue(nameof(Mode), Mode.ModeB)]
public object ThingsToShowWhenB;

[VisibleIfValue(nameof(Mode), Mode.ModeC)]
public object ThingsToShowWhenC;
```
...which again, I think will happen more often than wanting to do specialized checks (and for those you still have `VisibleIf`).

Also, there is the `invert` parameter like in `VisibleIf`, for checking `!=` instead of `==`.

mafiesto4:
Sounds nice but the purpose of current `VisibleIf` with custom property getter is to move all custom logic checks into the user code (getter) which can include very complex checks (eg. if several different values are matching criteria).

Arcnor:
> Sounds nice but the purpose of current `VisibleIf` with custom property getter is to move all custom logic checks into the user code (getter) which can include very complex checks (eg. if several different values are matching criteria).

I understand, and that's what was discussed as well. The new `VisibleIfValue` was made only to help with the common case of doing a direct comparison with a value, which I think happens the most often. Creating separate functions for each value is a bit too much boilerplate IMHO.

That said, if you don't think there is value please close instead of moving for later, I'm ok losing this if it's not the desired path.

Issue #1565, opened 2023-09-29
IDE Intellisense not working when using Plugin Projects and Master Branch Engine


schmidt-florian:
**Issue description:**
If you use Plugins Projects with the Flax Engine from the Master branch, the IDE does not detect fully the C# Engine code.
If you remove the plugin Projects everything is okay. If you switch to 1.6 installed from the lancher, everything is okay too.

**You can watch the issue in this [Video](https://github.com/FlaxEngine/FlaxEngine/assets/14238126/01cdc88f-5091-4534-a0d0-269831423e2a)**

In the First seconds you see that everything is fine, because the project was opened with 1.6
Then i switched to master branch engine. At minute 1:00, the IDE shows CS0122 for Script.
Then i switched back to 1.6 and generated project files, all fine.

I think there i some major or multple bugs with the plugin projects. Because i have currently only problems with it.
This maybe correlated to #1550 and with the problem i already spoke to @GoaLitiuM at the Discord.

**Steps to reproduce:**
Open attached project with 1.6 if you want it see working
Open attached project with master branch self build if you want to see the error
Generate project files
Open with your IDE (Tested with VS2022)
See result in MyScript.cs

**Minimal reproduction project:**
[Test.zip](https://github.com/FlaxEngine/FlaxEngine/files/12769540/Test.zip)

**Flax version:**
lancher install 1.6
(dev) master branch fd3f108
Steps i have done to install the master branche engine
- Clone it
- GenerateProjectFiles.bat
- PackageAll.bat
- RegisterEngineLocation.bat

Visual Studio 2022
Rider has probably the same issue too, but i can not check it because the 1.6 wont detect rider installed via toolbox.

schmidt-florian:
I want to add, this does not only happens with plugin projects. 
This happens also if I add a second Module. Like GameEditor.
Or it also happens if I only have the Module Game but enable ```BuildNativeCode = true;```
It looks like it happens every time when there is more than one module.

![image](https://github.com/FlaxEngine/FlaxEngine/assets/14238126/b1447237-a88c-4f7b-b626-a1513138146a)

schmidt-florian:
After building the engine with the current commit [6d92d0b](https://github.com/FlaxEngine/FlaxEngine/commit/6d92d0b58f4a38b1273b236702153d95983b5224) the issue is fixed. Probably with #1550.

PR #1570, opened 2023-09-29
Remap non-native Editor VS build configurations to native configurations

GoaLitiuM:
The Editor build configurations for non-desktop platforms (like `Android_ARM64`) should be remapped to matching desktop configuration when no exact configuration mapping can be found. The incorrect mapping leads to broken Intellisense when any of the Editor configurations are selected with `Android_ARM64` platform.


GoaLitiuM:
Possible fix for https://github.com/FlaxEngine/FlaxEngine/issues/1565

PR #1581, opened 2023-10-01
Improve Editor performance with multiple windows open

GoaLitiuM:
This should improve the overall framerate in editor when multiple windows, tooltips or menus are active at once by reducing the rendering framerate of inactive windows to 2fps.

There is some unexpected behavior with these changes however: `Editor.FramesPerSecond` used by the viewport widget does not show the effective framerate of that window anymore because the `RenderTask` might be running at reduced framerate. This however works accurately in games which does not use multiple windows.

### DIFF
--- a/Source/Engine/Platform/Base/WindowBase.cpp
+++ b/Source/Engine/Platform/Base/WindowBase.cpp
@@ -14,6 +14,7 @@
 #include "Engine/Scripting/ManagedCLR/MUtils.h"
 #include "Engine/Scripting/ManagedCLR/MMethod.h"
 #include "Engine/Scripting/ManagedCLR/MClass.h"
+#include "Engine/Engine/Time.h"
 
 #if USE_CSHARP
 // Helper macros for calling C# events
@@ -439,6 +440,9 @@ void WindowBase::OnGotFocus()
         return;
     _focused = true;
 
+    if (RenderTask && _settings.WindowFPSWhenNotFocused > ZeroTolerance)
+        RenderTask->RenderFPS = 0;
+
     GotFocus();
     INVOKE_EVENT_PARAMS_0(OnGotFocus);
 }
@@ -448,6 +452,9 @@ void WindowBase::OnLostFocus()
     if (!_focused)
         return;
     _focused = false;
+
+    if (RenderTask && _settings.WindowFPSWhenNotFocused > ZeroTolerance)
+        RenderTask->RenderFPS = _settings.WindowFPSWhenNotFocused;
 
     LostFocus();
     INVOKE_EVENT_PARAMS_0(OnLostFocus);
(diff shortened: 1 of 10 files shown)


Source/Engine/Platform/Base/WindowBase.cpp
```diff
@@ -439,6 +440,9 @@ void WindowBase::OnGotFocus()
         return;
     _focused = true;
 
+    if (RenderTask && _settings.WindowFPSWhenNotFocused > ZeroTolerance)
```
stefnotch, line 443: Is `_settings.WindowFPSWhenNotFocused` an integer? If so, just writing `_settings.WindowFPSWhenNotFocused > 0` should do the trick.

Source/Engine/Platform/Windows/WindowsWindow.cpp
```diff
@@ -1178,7 +1178,7 @@ LRESULT WindowsWindow::WndProc(UINT msg, WPARAM wParam, LPARAM lParam)
         }
         OnLostFocus();
         break;
-    case WM_ACTIVATEAPP:
+    case WM_ACTIVATE:
```
stefnotch, line 1181: Forgive my absolute noob question, but what's the difference between the two?
WM_ACTIVATE seems to have a 3rd possible value `WA_CLICKACTIVE 2 Activated by a mouse click.`, which should probably also be handled?

NoriteSC:
all is fine but opening any window just makes anny other window laggy including menus (with are windows)
vs hot realod is wirde with it unless lmb is pressed main window laggs
profiler at 2fps is not good

GoaLitiuM:
> all is fine but opening any window just makes anny other window laggy including menus (with are windows)

Menus shouldn't be laggy as they should receive focus when activated, do you have a case which reproduces this because I haven't noticed this issue?

> vs hot realod is wirde with it unless lmb is pressed main window laggs at 2fps is not good

Not sure about this one, the hot reload isn't triggered in background unless one of the windows are activated?

> profiler

For profiler I was thinking of having custom limits in place due to how resource intensive the window is to update every frame. Maybe when inactive, the framerate should be 10-15fps, and when profiling is active, 60fps?

> material instances and changes of propertys

Not sure what this means.

> particle editor

Same as with profiler window, if the particle effects are active then maybe the view shouldn't be throttled.

GoaLitiuM:
Ideally the views should only be updated (rendered + presented) when the displayed content changes, but the GUI doesn't support this type of use-case at the moment.

Arcnor:
> Ideally the views should only be updated (rendered + presented) when the displayed content changes, but the GUI doesn't support this type of use-case at the moment.

Knowing when the content changes is also a lot of work. Things like updating a sprite atlas change the icon displayed on the UI, but only because it's redraw all the time, there is no "content has changed" event anywhere, so having some sort of messaging system might be the previous step to fixing the UI proper, but that's also a lot of work, although might be worth having a discussion about.

GoaLitiuM:
The update rate of inactive windows which could have animated content probably needs to be handled case-by-case:

- Particle System editor (when simulation is active)
- Particle Emitter editor (when simulation is active)
- Material editor (can have animated materials in preview)
- Animation Graph editor (when simulation is active)
- Scene Animation editor (when live preview is active)

Issue #990, opened 2023-04-11
Clicking a UI button too often causes every second button press to drop


Bits360:
**Issue description:**

When clicking a UI button too often, without moving your mouse, every second button press will be dropped, not displaying that the button has been pressed, or registering input.

No logs (does not throw an error.)

**Steps to reproduce:**

Click any button repeatedly without moving your mouse (there is some strange cooldown when it is not moved, which isnt intended, and is rather annoying)

**Minimal reproduction project:**

[MyDebugProject1.zip](https://github.com/FlaxEngine/FlaxEngine/files/11205557/MyDebugProject1.zip)

**Flax version:**

1.5 latest on launcher (also tested in 1.4, both on a fresh reinstall of the engine to ensure that wasnt the problem)

mafiesto4:
Fixed in #1582

PR #1582, opened 2023-10-01
Handle double clicking in CheckBox, Button, Dropdown and DropPanel.

MineBill:
The aforementioned UI controls don't handle double clicks, which means every second(depends how fast the user is clicking) mouse down event gets swallowed by the `OnMouseDoubleClick` method. This is not *really* a problem per se but it's kind of annoying when you want to spam-click a button or a checkbox and makes the UI feel a bit more responsive.
Below are two videos showcasing the change. In both videos i click as fast as i can:

https://github.com/FlaxEngine/FlaxEngine/assets/30367251/a2465073-781a-4d45-b0be-6f68af071b56

https://github.com/FlaxEngine/FlaxEngine/assets/30367251/333b5800-f077-4795-aa16-d778e19487d2

### DIFF
--- a/Source/Engine/UI/GUI/Common/Button.cs
+++ b/Source/Engine/UI/GUI/Common/Button.cs
@@ -315,6 +315,28 @@ namespace FlaxEngine.GUI
         }
 
         /// <inheritdoc />
+        public override bool OnMouseDoubleClick(Float2 location, MouseButton button)
+        {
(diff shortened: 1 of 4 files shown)


Menotdan:
Haha you have no idea how much I really wanted this fixed
I click around a lot and fidgeting with buttons not working was frustrating to a high degree

Tryibion:
Related: #990

PR #1606, opened 2023-10-04
crash hacy fix when pasting get parameter node to other graph

NoriteSC:
image


NoriteSC:
> Why not simple check?
> 
> ```cs
> if (Surface == null || _input == null)
>     return;
> ```

idk if this is only for material graph so having something saying what is wrong is better the silence

Issue #1602, opened 2023-10-04
Dropdown menu in FlaxEditor sometimes appears in the wrong place


nothingTVatYT:
**Issue description:**
Sometimes the dropdown menu in the FlaxEditor on Linux (KDE/X11) does not appear below the menubar but somewhere in the editor, preferably in the view or game panel.

**Steps to reproduce:**
Load any project and repeatedly open the menus from the main menubar. Once in a while the dropdown menu appears somewhere in the center of the editor instead of below the menubar. This happens sporadically and possibly only with the Edit menu.

**Minimal reproduction project:**
Any project will do.

**Flax version:**
1.6.6345

nothingTVatYT:
I haven't seen this problem for quite some time now so it's probably solved with the recent LinuxWindow changes.

PR #1610, opened 2023-10-04
Fix menu misalignment and window placement on Linux

nothingTVatYT:
This fixes the LinuxPlatform::GetMonitorBounds which caused the menu misalignment problems in multi-screen environments and the ignored placement hints of the window manager.


mafiesto4:
Does it fix issue #1602?

nothingTVatYT:
I'm not sure about #1602 as it happens so sporadically but I guess it won't. OTOH I don't know the reason of #1602 so ... maybe?

PR #1611, opened 2023-10-04
Improve viewport camera settings

solnem:
## Summary
This PR is trying to improve the camera settings widget and related camera settings in `EditorViewport`.

Based on my search, at this time, it does not seem to fix any open issues.

The PR contains:
- [x] Gathering all camera settings into one camera widget with context menu
- [x] Improved panning, including the ability to manually set the panning speed and have panning based on distance to the focused/target object (aka relative panning, thanks @stefnotch!)
- [x] Additional default settings in the Flax Engine options
- [x] Ability to reset all camera settings to default through a button in the context menu of the camera settings widget
- [x] Saving and loading of cached camera settings values (similar to how it is in `MainEditorGizmoViewport`)
- [x] Camera speed easing during camera flight and when using `EditorViewport.AdjustCameraMoveSpeed` using `Mathf.Pow`
- [x] Possibility to disable camera easing completely
- [x] Toggle button to quickly switch between orthographic and perspective mode (still missing dedicated icon, feel free to add one)

## Type of change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [x] Refactoring
- [x] Requires documentation update

### Grouping of related settings
The camera settings were spread across the view widget and the camera speed scale widget. With this change, they are all combined into one camera settings widget.

image

### Camera panning
When panning the camera, the panning speed would always be locked at a certain value. These changes add the possibility to manually set the panning speed or use a relative panning mode, which automatically calculates the panning speed based on the camera target/focused object.

Before:
FlaxEditor_2023-10-05_00-07-25
After:
FlaxEditor_2023-10-05_00-09-36

### Adjustable camera speed and camera speed easing
Previously, the camera speed setting was limited to only a few options. Now, from the context menu, the camera settings widget allows you to set the minimum, maximum and current speed of the camera.

FlaxEditor_2023-10-05_00-12-40

In addition to that, when using the camera flight mode (holding RMB or using `CameraToggleRotation`), the camera speed is now adjusted based on `Mathf.Pow` to slow down the camera speed in the beginning (min speed).

FlaxEditor_2023-10-05_00-45-34

The possibility to fully disable camera speed easing was added to the camera settings widget.
image

The camera settings widget now uses a fixed size to prevent being resized whenever the string length of the speed value changes. On top of that, the string gets formatted using `string.Format`. In the `FloatValueBox` inside the context menu, the movement speed will show up rounded up to the third decimal to make it look more appealing to the user.

Before:
FlaxEditor_2023-10-05_20-08-43
After:

### New default settings based on changes
In the options of Flax Engine, additional settings have been added based on the changes that were made to the possible camera settings. These changes get synced with the editor viewport and are also the ones that are used when the reset button is pressed.

image

The new settings described:
| Setting | Description |
| - | - |
|**Total Camera Speed Steps** | Changes the amount of steps it requires for the camera speed to go from minimum to maximum (1-128).|
|**Camera Easing Degree**| When using camera easing, the camera easing degree will determine how slow/fast the camera speed increases/decreases across steps (1.0-8.0).|
|**Min Movement Speed**| The default minimum speed of the camera (0.05-32).|
|**Max Movement Speed**| The default maximum speed of the camera (32-1000).|
|**Use Camera Easing**| Enable/disable camera easing.|
|**Use Relative Panning**| Enable/disable whether the panning speed of the camera should be affected by the distance to the target/focused object or not. |

### Easing value calculation
This was simplified to use `Mathf.Pow` [here](https://github.com/FlaxEngine/FlaxEngine/pull/1611/commits/00aa54cde82da867b2a861cd1b6bf307dafc0a7f), should be fine now.

### ~Additional easing curve setting~
~There doesn't necessarily have to be an ease out on the camera speed. However, a possible change to the camera easing could be to allow changing the easing degree between 0% and 50% and the one between 50% and 100% separately.~

### Orthographic/perspective toggle button
This button is currently missing the correct icon. I had one that I made myself, but I was not too happy with it. Best case, it is an icon that represents the perspective mode, as that one is more easily identifiable.

### DIFF
--- a/Source/Editor/Options/ViewportOptions.cs
+++ b/Source/Editor/Options/ViewportOptions.cs
@@ -47,6 +47,13 @@ namespace FlaxEditor.Options
         public float MaxMovementSpeed { get; set; } = 64f;
 
         /// <summary>
+        /// Gets or sets the degree to which the camera will be eased when using camera flight in the editor window.
+        /// </summary>
+        [DefaultValue(3.0f), Limit(1.0f, 8.0f)]
+        [EditorDisplay("Defaults"), EditorOrder(120), Tooltip("The default degree to which the camera will be eased when using camera flight in the editor window.")]
+        public float CameraEasingDegree { get; set; } = 3.0f;
+
+        /// <summary>
         /// Gets or sets the default near clipping plane distance for the viewport camera.
         /// </summary>
         [DefaultValue(10.0f), Limit(0.001f, 1000.0f)]
--- a/Source/Editor/Options/ViewportOptions.cs
+++ b/Source/Editor/Options/ViewportOptions.cs
@@ -24,47 +24,89 @@ namespace FlaxEditor.Options
         [DefaultValue(1.0f), Limit(0.01f, 100.0f)]
         [EditorDisplay("General"), EditorOrder(101), Tooltip("The mouse wheel sensitivity applied to zoom in orthographic mode.")]
         public float MouseWheelSensitivity { get; set; } = 1.0f;
-
+        
         /// <summary>
-        /// Gets or sets the default movement speed for the viewport camera (must match the dropdown menu values in the viewport).
+        /// Gets or sets the default movement speed for the viewport camera (must be in range between minimum and maximum movement speed values).
         /// </summary>
-        [DefaultValue(1.0f), Limit(0.01f, 100.0f)]
-        [EditorDisplay("Defaults"), EditorOrder(110), Tooltip("The default movement speed for the viewport camera (must match the dropdown menu values in the viewport).")]
+        [DefaultValue(1.0f), Limit(0.05f, 64.0f)]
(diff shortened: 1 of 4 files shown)


solnem:
This PR is pretty much done. I will open it up for reviewing and testing in a bit.
A few things however:
1. Should the orthographic/perspective toggle button stay (currently the arrow button)?
2. Should there be more/other toggle buttons associated to the camera (such as toggle camera easing, toggle relative panning, etc.)?
3. If the orthographic/perspective toggle stays, what should the icon be? The icon I used before was a rectangle that was skewed on one side. Apart from it being maybe overly simplified, the reason why I didn't like that icon was because the on state would mean orthographic mode was enabled. The icon however said the opposite.
As it is now, I have changed it so that the on state means perspective mode is active, so the icon would again fit.
4. Lastly, any final thoughts on what settings/options should be in the view widget and what should be in the camera settings widget? I moved the viewpoints over to the camera, because they are moving the camera. What about brightness? Resolution?

PR #1619, opened 2023-10-05
Visject: Automatic casting

Chikinsupu:
When trying to connect two node ports, which types are theoretically compatible via casting, a casting node should be added automatically in front of the target port. This mimics the behavior of Unreal for example.
image

Currently when trying to connect two ports which are compatible via casting, nothing happens. Instead a NotImplementedException gets thrown. (See before gif)

This PR implements automatically spawning a casting node when needed, setting it's type and positioning it accordingly.
This functionality enhances this PR: #1522 but is not depending on it.

Before:
BeforeCasting

After:
AfterCasting

I already did a bit of testing including Undo/Redo functionality etc.

### DIFF
--- a/Source/Editor/Surface/Archetypes/Tools.cs
+++ b/Source/Editor/Surface/Archetypes/Tools.cs
@@ -787,7 +787,7 @@ namespace FlaxEditor.Surface.Archetypes
             }
         }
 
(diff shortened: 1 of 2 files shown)


Tryibion:
Awesome!

PR #1628, opened 2023-10-06
SOFT_ASSERTs with editor

NoriteSC:
added soft asserts 
for when u are not shure if it should be an asset or cheak or the assert is not critical or code can lock up the editor
currently i encounter a constant crash because object is too far from origin


mafiesto4:
Keep in mind that we already have `CHECK(expression)` and `CHECK_RETURN(expression, returnValue)` that do sth similar.

NoriteSC:
> Keep in mind that we already have `CHECK(expression)` and `CHECK_RETURN(expression, returnValue)` that do sth similar.

i know it is just something specific for editor

PR #1635, opened 2023-10-07
Add color picker option to accept changes on dismissal

Arcnor:
I'm really used to the Unity (and maybe other creative tools) of color picker functionality, where you don't need to press "Ok" for the color to be accepted, just dismissing the dialog is enough.

This PR adds an option to be able to do that, and I'd be grateful if somebody has some feedback on the naming, category location and/or tooltip description, naming is difficult.

### DIFF
--- a/Source/Editor/Options/InterfaceOptions.cs
+++ b/Source/Editor/Options/InterfaceOptions.cs
@@ -149,6 +149,13 @@ namespace FlaxEditor.Options
         public FlaxEngine.GUI.Orientation ContentWindowOrientation { get; set; } = FlaxEngine.GUI.Orientation.Horizontal;
 
         /// <summary>
+        /// Gets or sets the editor content window orientation.
+        /// </summary>
+        [DefaultValue(false)]
+        [EditorDisplay("Interface"), EditorOrder(300), Tooltip("If checked, color pickers will always modify the color unless 'Cancel' if pressed, otherwise color won't change unless 'Ok' is pressed (default)")]
+        public bool ColorPickerAlwaysChangesColor { get; set; } = false;
+
+        /// <summary>
         /// Gets or sets the timestamps prefix mode for output log messages.
         /// </summary>
         [DefaultValue(TimestampsFormats.TimeSinceStartup)]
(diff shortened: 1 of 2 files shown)


Source/Editor/Options/InterfaceOptions.cs
```diff
@@ -148,6 +148,13 @@ public enum PlayAction
         [EditorDisplay("Interface"), EditorOrder(280), Tooltip("Editor content window orientation.")]
         public FlaxEngine.GUI.Orientation ContentWindowOrientation { get; set; } = FlaxEngine.GUI.Orientation.Horizontal;
 
+        /// <summary>
+        /// Gets or sets the editor content window orientation.
```
Tryibion, line 152: Probably a copy and paste error here.

Source/Editor/Options/InterfaceOptions.cs
```diff
@@ -148,6 +148,13 @@ public enum PlayAction
...
+        /// </summary>
+        [DefaultValue(false)]
+        [EditorDisplay("Interface"), EditorOrder(300), Tooltip("If checked, color pickers will always modify the color unless 'Cancel' if pressed, otherwise color won't change unless 'Ok' is pressed (default)")]
+        public bool ColorPickerAlwaysChangesColor { get; set; } = false;
```
Tryibion, line 156: What about like `AutoAcceptOnColorChange` or `AutoPickOnColkrChange` or `AutoAcceptColorPickerColor` or something? Idk naming things is hard.

Source/Editor/Options/InterfaceOptions.cs
Arcnor: Indeed, will fix

Source/Editor/Options/InterfaceOptions.cs
Arcnor: Well, there is nothing really automatic about the option, is it? You still pick the color, is just that the change doesn't need "acknowledgement", but not sure.

Maybe we can say "Color picker magical option 42, surprise!" or something... 😁

Arcnor: That said "Auto accept color picker change" might just work. What do you think?

PR #1659, opened 2023-10-09
Implement editor automation for Asset Importing.

Menotdan:
This PR adds new functionality so a user can register to an event in an `EditorPlugin` that lets them script how an asset is imported.
Demo (compressed due to github file limit):

https://github.com/FlaxEngine/FlaxEngine/assets/32620310/31f83a96-7291-4820-a74f-6808ce40d7af


Menotdan:
Main problem with this PR that I'd like to fix is that, since its a delegate in user code, the exceptions there are somehow causing editor crashes. Not sure how to fix this, but it would be nice to get done before merge.

ontrigger:
I feel like a proper class-based solution not tied to an editorplugin would be much better. Unity's scriptable importers allow you to create custom inspector ui for assets and have many helper methods, including support for working with subassets. You just can't get something like that with a delegate

https://docs.unity3d.com/Manual/ScriptedImporters.html

Menotdan:
> I feel like a proper class-based solution not tied to an editorplugin would be much better. Unity's scriptable importers allow you to create custom inspector ui for assets and have many helper methods, including support for working with subassets. You just can't get something like that with a delegate
> 
> https://docs.unity3d.com/Manual/ScriptedImporters.html

Yeah, this sounds doable. Most of the work is already done for me with Flax internals I think. 
I could create a class that inherits from EditorPlugin instead, since I don't think it makes sense to copy the functionality of EditorPlugin just for this feature. But I would have to look at the code.

Menotdan:
@mafiesto4 could this be moved to 1.8 (or 1.7)? I created a class as discussed in the discord, and this is now easily expandable. It is also usable in it's current state (I use it in my own project), so I think it could be merged in the future after some testing. I don't want to expand the scope of the PR too much just because it will get buried in other things I'm working on.

Maybe someone else has thoughts on this too.

Menotdan:
Closing this for now as hopefully large importing changes are being made.

PR #1665, opened 2023-10-10
Change terrain brush size with scroll

RuanLucasGD:
This adds a way to change terrain brush size with mouse scroll pressing shift.

[Debug 2023.10.10 - 10.49.05.01.webm](https://github.com/FlaxEngine/FlaxEngine/assets/79365912/da9a2ab2-b93b-44f4-8d3c-a448f744edce)


mafiesto4:
Why that brush setter is needed?

RuanLucasGD:
@mafiesto4 fixed, thanks

PR #1671, opened 2023-10-11
Visject: Convert node to parameter

### LINKED ISSUE #1170: Convert to Parameter in Visject
- Allow converting a node to a parameter faster
image
image

Chikinsupu:
This PR adds the ability to convert a node to a parameter with only two clicks - similar to Unreal Engine for example.
245352424-91ac6725-0324-4477-bc01-3802512e3cb9
Closes #1170

### Tasks
- [x] Implement node conversion functionality
- [x] Initialize parameter with node values
- [x] Replace node with ParamGet node
- [x] Add functionality to basic/generic constant nodes (e.g. Float3)
- [x] Connect ParamGet node to previously connected boxes
- [x] Add functionality to more complex nodes (e.g. Texture)
- [x] Cleanup
- [x] Add XML comments
- [x] Testing
- [x] Done 🥳

**Didn't do - postponed to future PRs**
- [x] Put "Convert to Parameter" item in first place in the context menu
- [x] Give user the ability to instantly rename parameter after creation

**Known bugs**
- When converting a node with connections to a parameter node and then undoing that action, all connections to that node get lost. This is due to how the Box.CreateConnection() method works, since it doesn't record any undo actions. This can be fixed easily, but shouldn't be part of this PR

### Feature preview

Convert to parameter
ConvertNodeToParameter

Convert Texture
TextureConvert

Convert Rotation and Uint
ConvertVS

### DIFF
--- a/Content/Shaders/GlobalSignDistanceField.flax
+++ b/Content/Shaders/GlobalSignDistanceField.flax
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:2c8aa181a814d69b15ffec6493a71a6f42ae816ce04f7803cff2d5073b4b3c4f
-size 11790
+oid sha256:e075583620e62407503c73f52487c204ddcad421e80fa621aebd55af1cfb08d5
+size 11798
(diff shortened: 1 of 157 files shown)


mafiesto4:
Thanks!

Issue #1651, opened 2023-10-08
Rider ignors files of plugin projects


schmidt-florian:
**Issue description:**
Rider by default ignores files of plugin projects in his solution-wide analysis.
I do not know if there is another reason. But as long the Project are generated to bo as "not build" for rider, the errors are ignored.
If I set it to build, the errors are not ignored. The issue is that this is overwritten by Flax on Regenerate Project files. Even it does now happen not always, it is annoying to set it every time to build. Especially if you have 5 Modules in 2 Plugin Projects.

**Steps to reproduce:**

1. Open attached project (or create a project with a plugin project and produce a script error in the plugin project)
2. Generate Project files.
3. Open Project in Rider, after rider has finished his analysis, there will be no errors. Image. 1
4. Open Solution Properties and check the build option on the project "Test" Image. 2/3
5. Rider will now show the error. Image. 4
6. On regenerate Project files, it will be set to "not build" and the error will be ignored again.

Image. 1
![image](https://github.com/FlaxEngine/FlaxEngine/assets/14238126/77c3bbe8-81f7-46ef-873a-18d9b19a9778g|width=400px)
Image. 2
![image](https://github.com/FlaxEngine/FlaxEngine/assets/14238126/883f9ed6-c61e-4867-807b-cf455670f0a2|width=400px)
Image. 3
![image](https://github.com/FlaxEngine/FlaxEngine/assets/14238126/5b774492-0dab-4c99-8287-24bbad4f8a20|width=400px)
Image. 4
![image](https://github.com/FlaxEngine/FlaxEngine/assets/14238126/2cf72cc2-2145-4448-8db1-ee86747c2ef4|width=400px)

**Minimal reproduction project:**
[TestPluginError.zip](https://github.com/FlaxEngine/FlaxEngine/files/12841159/TestPluginError.zip)

**Flax version:**
master 5283337

mafiesto4:
Fixed in #1673

PR #1673, opened 2023-10-11
Build C# projects in VS/Rider solution configurations

GoaLitiuM:
Rider's solution wide analysis does not work properly when projects are not included in the active configuration for build.

Also implements experimental MSBuild .targets files for running Flax.Build when building the solution in IDEs, the same way VC++ projects use custom build commands to invoke Flax.Build.

### DIFF
--- a/Source/Tools/Flax.Build/Projects/VisualStudio/CSSDKProjectGenerator.cs
+++ b/Source/Tools/Flax.Build/Projects/VisualStudio/CSSDKProjectGenerator.cs
@@ -91,7 +91,7 @@ namespace Flax.Build.Projects.VisualStudio
             var baseOutputDir = Utilities.MakePathRelativeTo(project.CSharp.OutputPath ?? baseConfiguration.TargetBuildOptions.OutputFolder, projectDirectory);
             var baseIntermediateOutputPath = Utilities.MakePathRelativeTo(project.CSharp.IntermediateOutputPath ?? Path.Combine(baseConfiguration.TargetBuildOptions.IntermediateFolder, "CSharp"), projectDirectory);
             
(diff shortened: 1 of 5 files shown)


GoaLitiuM:
Fixes #1651

Source/Tools/Flax.Build/Projects/VisualStudio/CSSDKProjectGenerator.cs
```diff
@@ -106,14 +111,14 @@ public override void GenerateProject(Project project, string solutionPath)
             csProjectFileContent.AppendLine("    <LangVersion>11.0</LangVersion>");
             csProjectFileContent.AppendLine("    <FileAlignment>512</FileAlignment>");
 
-            // Needed for Hostfxr
-            csProjectFileContent.AppendLine("    <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>");
```
mafiesto4, line 110: Why was it removed?

Source/Tools/Flax.Build/Projects/VisualStudio/CSSDKProjectGenerator.cs
```diff
@@ -106,14 +111,14 @@ public override void GenerateProject(Project project, string solutionPath)
...
-            csProjectFileContent.AppendLine("    <EnableDynamicLoading>true</EnableDynamicLoading>");
```
mafiesto4, line 111: Why was it removed?

GoaLitiuM: These values are not used by IDE, I think these were leftovers from the early prototypes of the .NET 7 support.

PR #1688, opened 2023-10-12
Fix #63

### LINKED ISSUE #63: Small window title bar dragging issue
**Issue description:**
The dockable windows, like the material windows, are a bit odd when dragging. They sometimes jump to a different position when you drag them.

This happens because in the engine core, the mouse position doesn't get updated as soon as the mouse is above a window title bar. 

This piece of code doesn't get triggered. I'd gladly fix it, but am unsure how to best proceed.
https://github.com/FlaxEngine/FlaxEngine/blob/da389eea03ab0775e384d50ff65bf4d4792f3e47/Source/Engine/Platform/Windows/WindowsInput.cpp#L198-L203

**Steps to reproduce:**

1. Open a material window
2. Move your cursor below the title bar, at the very left of the window
3. Move your cursor onto the title bar 
4. Move your cursor all the way to the right on the title bar. Make sure to stay on top of the title bar while moving.
5. Start dragging
6. The window jumps

image
Basically just follow that mouse path ^ and start dragging

**Flax version:**
Flax 1.0

stefnotch:
Seems to fix #63 , but it might have small side-effects if the mouse position tracking is not expected to go outside the client area.

### DIFF
--- a/Source/Engine/Platform/Windows/WindowsInput.cpp
+++ b/Source/Engine/Platform/Windows/WindowsInput.cpp
@@ -200,6 +200,12 @@ bool WindowsMouse::WndProc(Window* window, const UINT msg, WPARAM wParam, LPARAM
         OnMouseMove(mousePos, window);
         result = true;
         break;
+    }    
+    case WM_NCMOUSEMOVE:
+    {
+        OnMouseMove(mousePos, window);
+        result = true;
+        break;
     }
     case WM_MOUSELEAVE:
     {


mafiesto4:
Actually the problem here is that it might trigger mouse move event after mouse left the window (after `WM_MOUSELEAVE` message). Do you have #63 issue on the latest master? Recently, we've got some improvements to dock windows (for macOS0 but it improved Windows experience as well.

GoaLitiuM:
Latest master seems to behave better, but it has some weird issue with the window dragging where it doesn't detect the dragging action if you enter the window title bar from under it and try to drag it. With this PR that quirk doesn't seem to happen.

PR #1720, opened 2023-10-15
Collection Editor UI Change

Tryibion:
Some simple changes to the collection editor UI to better handle grouped properties when  needed. Adds easy move up and down buttons for collections. This also adds an enum to the CollectionAttribute to change the layout type of the editor. This makes the dictionary editor similar (it has a wonky double click to rename, so i just forced it open on double click. it is kind of janky, so I am sure it can be improved).

image


Tryibion:
Closing for redo at some point.

Issue #1694, opened 2023-10-12
Color picker is updated on mouse movement but a click cancels the dialog in Linux


nothingTVatYT:
**Issue description:**
A click on a Color property (e.g. the light color) brings up the color chooser and the magnifier symbol triggers the color picker. Mouse movement is tracked correctly and the widget is updated accordingly. Problem is that on a mouse click the dialog closes and the property is not changed as if the color chooser was cancelled.

There is no error or anything related in the output log when this happens.

**Minimal reproduction project:**
Independent on a project.

**Flax version:**
1.7.6401 (master branch), ArchLinux KDE/X11

nothingTVatYT:
The problem is the "Auto cancel on lost focus" in ColorPickerDialog.OnShow(). If I disable that, the color picker works as expected.

nothingTVatYT:
There is another small issue related to that: Once I got it working temporarily by disabling the cancelling I noticed the picked color has alpha set to 0.
A quick fix:

```
diff --git a/Source/Editor/Utilities/ScreenUtilities.cpp b/Source/Editor/Utilities/ScreenUtilities.cpp
index 44f52350e..943c2dcaf 100644
--- a/Source/Editor/Utilities/ScreenUtilities.cpp
+++ b/Source/Editor/Utilities/ScreenUtilities.cpp
@@ -73,6 +73,7 @@ Color32 ScreenUtilities::GetColorAt(const Float2& pos)
     outputColor.R = color.red / 256;
     outputColor.G = color.green / 256;
     outputColor.B = color.blue / 256;
+    outputColor.A = 255;
     return outputColor;
 }
```

nothingTVatYT:
Fixed in master branch

PR #1722, opened 2023-10-15
fix color picker (signal handler + alpha channel)

nothingTVatYT:
This fixes the signal handling problem on Linux which closes the dialog before the color under the mouse cursor is accepted and the alpha channel of the picked color was always 0.

### DIFF
--- a/Source/Editor/GUI/Dialogs/ColorPickerDialog.cs
+++ b/Source/Editor/GUI/Dialogs/ColorPickerDialog.cs
@@ -218,7 +218,6 @@ namespace FlaxEditor.GUI.Dialogs
 
         private void OnColorPicked(Color32 colorPicked)
         {
-            Editor.Log("OnColorPicked " + colorPicked + ", _active eye dropper=" + _activeEyedropper);
             if (_activeEyedropper)
             {
                 _activeEyedropper = false;
--- a/Source/Editor/GUI/Dialogs/ColorPickerDialog.cs
+++ b/Source/Editor/GUI/Dialogs/ColorPickerDialog.cs
@@ -218,6 +218,7 @@ namespace FlaxEditor.GUI.Dialogs
 
         private void OnColorPicked(Color32 colorPicked)
         {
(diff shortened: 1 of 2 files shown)


nothingTVatYT:
Fixes #1694

Source/Editor/GUI/Dialogs/ColorPickerDialog.cs
```diff
@@ -218,6 +218,7 @@ public ColorPickerDialog(Color initialValue, ColorValueBox.ColorPickerEvent colo
 
         private void OnColorPicked(Color32 colorPicked)
         {
+            Editor.Log("OnColorPicked " + colorPicked + ", _active eye dropper=" + _activeEyedropper);
```
mafiesto4, line 221: This could be removed

Source/Editor/GUI/Dialogs/ColorPickerDialog.cs
```diff
@@ -319,7 +320,7 @@ public override void Draw()
         protected override void OnShow()
         {
             // Auto cancel on lost focus
-            ((WindowRootControl)Root).Window.LostFocus += OnCancel;
+            //((WindowRootControl)Root).Window.LostFocus += OnCancel;
```
mafiesto4, line 323: Wrap it around `#ifdef !PLATFORM_LINUX`/`#endif` instead.

mafiesto4 (changes requested):

Source/Editor/GUI/Dialogs/ColorPickerDialog.cs
nothingTVatYT: Oh, missed that one.

Source/Editor/GUI/Dialogs/ColorPickerDialog.cs
nothingTVatYT: Done.

mafiesto4 (approved):

PR #1725, opened 2023-10-16
Improve documentation for Model vs SkinnedModel vs AnimatedModel vs StaticModel.

Menotdan:
Since it was quite confusing.

### DIFF
--- a/Source/Engine/Content/Assets/Model.h
+++ b/Source/Engine/Content/Assets/Model.h
@@ -9,7 +9,7 @@ class Mesh;
 class StreamModelLODTask;
 
 /// <summary>
-/// Model asset that contains model object made of meshes which can rendered on the GPU.
+/// An asset from disk that stores static model data, which can be rendered on the GPU.
 /// </summary>
 API_CLASS(NoSpawn) class FLAXENGINE_API Model : public ModelBase
 {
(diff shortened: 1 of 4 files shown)


mafiesto4:
What about using `Animated Model`/`Static Model` instead of `AnimatedModel`/`StaticModel` for Editor Display label?

Menotdan:
> What about using `Animated Model`/`Static Model` instead of `AnimatedModel`/`StaticModel` for Editor Display label? 

Maybe, but I was trying to make clear the connection between the class and the display in the editor.

MineBill:
I think this change might make it even more confusing because while there is an actor named "StaticModel", the model asset that it uses will be displayed under a "StaticModel" header in the properties. If this PR is a response to the discussion in the #scripting discord channel yesterday, i think the biggest issue is people assuming(and rightly so) that the panels in the properties window represent components, while in reality they are just groupings of properties the actor has.

Menotdan:
> I think this change might make it even more confusing because while there is an actor named "StaticModel", the model asset that it uses will be displayed under a "StaticModel" header in the properties. If this PR is a response to the discussion in the #scripting discord channel yesterday, i think the biggest issue is people assuming(and rightly so) that the panels in the properties window represent components, while in reality they are just groupings of properties the actor has. This is a bigger issue and requires much more detailed docs explaining that actors use a hierarchy and need the appropriate handling while scripts act more like components.

How about "StaticModel Properties" and "AnimatedModel Properties"? I think a change should take place in the editor display somehow, because I have seen many people get confused by the current naming.

PR #1739, opened 2023-10-18
Fix terrain painting.

### LINKED ISSUE #1654: Terrain painting doesn't work as expected
**Issue description:**
Terrain Paint doesn't work as expected. 

Issues:

1. Layers are not blending properly
2. You can't paint over the top layer. For example in this terrain material my top layer is 4 and I can't paint over that layer. I need to select an empty layer and paint over it to remove it. (I am not sure if this is the intended behaviour)

Example:

Screenshot_357

Used terrain material:
image

Here is a short video:

https://github.com/FlaxEngine/FlaxEngine/assets/47716519/0ca62122-b150-4d85-b333-9fc9251608f4

**Steps to reproduce:**

Create a terrain material with multiple layers and start painting. Layers won't blend properly if it is below the layer you are trying to paint over. And painting with Layer 0 displays a really weird visual bug which can be seen in the attached video.  

**Flax version:**
1.6

MineBill:
Fixes #1654

This is an attempt to fix the issue with terrain painting by making sure the splatmap texture layers are correctly adjusted when one of them is updated by painting. I am a where of a small bug where keeping the mouse button down and painting on one area without moving, will cause the edges of the brush to create a few pixels with weird values.

Old:

https://github.com/FlaxEngine/FlaxEngine/assets/30367251/42b732b6-8ab7-47e8-9139-821196f24c22

New:

https://github.com/FlaxEngine/FlaxEngine/assets/30367251/5c7e059b-d234-46e8-be6b-a87c99365a10

I am also planning to add support for painting the other 4 layers.

### DIFF
--- a/Source/Editor/Tools/Terrain/Paint/Mode.cs
+++ b/Source/Editor/Tools/Terrain/Paint/Mode.cs
@@ -67,11 +67,13 @@ namespace FlaxEditor.Tools.Terrain.Paint
 
             // Prepare
             var splatmapIndex = ActiveSplatmapIndex;
+            var splatmapIndexOther = (splatmapIndex + 1) % 2;
             var chunkSize = terrain.ChunkSize;
             var heightmapSize = chunkSize * FlaxEngine.Terrain.PatchEdgeChunksCount + 1;
             var heightmapLength = heightmapSize * heightmapSize;
             var patchSize = chunkSize * FlaxEngine.Terrain.UnitsPerVertex * FlaxEngine.Terrain.PatchEdgeChunksCount;
-            var tempBuffer = (Color32*)gizmo.GetSplatmapTempBuffer(heightmapLength * Color32.SizeInBytes).ToPointer();
+            var tempBuffer = (Color32*)gizmo.GetSplatmapTempBuffer(heightmapLength * Color32.SizeInBytes, splatmapIndex).ToPointer();
+            var tempBufferOther = (Color32*)gizmo.GetSplatmapTempBuffer(heightmapLength * Color32.SizeInBytes, (splatmapIndex + 1) % 2).ToPointer();
             var unitsPerVertexInv = 1.0f / FlaxEngine.Terrain.UnitsPerVertex;
             ApplyParams p = new ApplyParams
             {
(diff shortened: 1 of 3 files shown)


abrasivetroop:
Great

abrasivetroop:
Lmao sorry the compression made the video horrible


### Issue with the final layer

https://github.com/FlaxEngine/FlaxEngine/assets/47716519/a1d1ada6-ad1a-4e27-a0d0-3730b5478f78



### Blending Issue

Also as you can see blending doesn't happen properly. I was trying to paint over Layer 2 with Layer 1 but at the edge of the brush it reveals Layer 0

image

image

MineBill:
The first issue is because I haven't implemented deleting from the other splatmap yet. I'll probably update this PR to add this feature within the next couple of hours.

The second issue I can't figure out for the life of me. I mentioned in the pr description above.

abrasivetroop:
> The second issue I can't figure out for the life of me. I mentioned in the pr description above.

Ohh sorry! I missed that part. I hope you will be able to fix it. I looked at the code but as a c# noob I didn't understand anything lol

MineBill:
All 8 layers should be able to interact with each other now.

As for the other bug, i know why it's happening. When creating a terrain material we use Lerp nodes and the layer weight value to lerp between textures. If all layers were to be 0 (completely black splatmaps) then due to the nature of the Lerp nodes in the material, the very first texture would be chosen:
image

The way i modify the splatmaps right now when painting a layer works by increasing the color for the current layer, while reducing the color for all the other layers. Because of this, when a channel goes from 1.0 to 0.0 and the others go from 0.0 to 1.0, it's possible that when two colors of the splatmap meet the will both be below 0.5. Here is me inspecting the terrain splatmap i exported from Flax:

https://github.com/FlaxEngine/FlaxEngine/assets/30367251/464a5086-2066-473c-ac4c-85bd2d2a33f5

You can see that when transition from one channel (R) to the other (G), the R and G values **both** drop below 0.5(127.5 because GIMP displays the byte value). Because all the other channels are at 0 and these two are a bit below 0.5, all of the Lerp nodes will either choose their A input or favor it more and this will bubble up and end up favoring the very first texture more than the others:
Untitled

At least, this is what I think the problem is. I'm not sure what would be the course of action. Keep in mind that the alpha channel is used a layer, so we can't simply do alpha blending like you would in gimp.

mafiesto4:
@MineBill is it ready for review and merge?

MineBill:
@mafiesto4 Yes and no? If you read my comments on this thread you'll see that even though my changes fix some stuff, there are some different bugs appearing but it could be merged as is.

However, I'd like to add a few more comments:  I took a look at how various different tools do splatmap painting and I'm not sure if the current solution(reading and writing pixels from one texture) is enough to fix the aforementioned bugs. Some implementations store the layers and the final splatmap independently, which should fix most bugs but it might be harder to implement. I didn't attempt that solution myself. In any case, it's up to you whether to merge it or not.

PR #1740, opened 2023-10-19
Add button to reload scenes

RuanLucasGD:
This is a useful option if you need to edit your scene outside Flax editor, like download commits from gitbug or fixe manually your broken file.
image


mafiesto4:
Maybe move it below `Close scenes` button?

RuanLucasGD:
@mafiesto4 done

Issue #1741, opened 2023-10-19
Editor: Hierarchy Ordering Overhaul


madsfs:
Whatever code is responsible for reordering hierarchies in the Editor needs a do-over. It is incredibly tedious to reorder objects without accidentally making them children of random objects and the order of objects is reversed when you move multiple of them at a time.

See the attached Gifs for examples.

What I propose is that perhaps Key modifiers could be used to help the process. So if you hold down, say Shift while moving an object in the hierarchy then it can only be reordered, it can't be parented. But I'm sure the UI might just be in need of a better implementation as other engines don't seem to have this issue.

https://github.com/FlaxEngine/FlaxEngine/assets/113017777/b361c5e2-3822-4a9f-a28e-28b994ff13f1

https://github.com/FlaxEngine/FlaxEngine/assets/113017777/f446dd09-84b5-4cd0-a55d-e25b1f24d458

madsfs:
Order not being kept seems to be an issue in general. I copied a structure and when I paste it, the order gets very messed up. (see the attached pictures)
![image](https://github.com/FlaxEngine/FlaxEngine/assets/113017777/01f4cc8f-3a9e-49df-ac84-69da7acea668)
![image](https://github.com/FlaxEngine/FlaxEngine/assets/113017777/89340a99-55fb-4c89-b81c-c3fb23ece044)

mafiesto4:
Fixed in c561d684eba1e0cea290b5d07a2b2603132bce72 and ed3ac0af8ea7f2ab4b740f34313a5f38ab4898a8.
Improved visuals of drag&drop in 8de2249ed82d6072641226fb5af7b7027d5817f4.

PR #1744, opened 2023-10-19
Use PrefabObjectID when setting/getting the actor expanded value for prefabs.

### LINKED ISSUE #1742: Prefabs/Editor Hierarchy: Collapsing/Expanding Children
Currently whenever I open a Prefab in the Editor for editing *every* child is expanded of every parent object. This makes for a very messy presentation when all children collapsed from the get-go would be cleaner.

However, even if the default behaviour stays as-is, then at least provide an option to collapse all or expand all children of a parent object. The way Unity did this, which was elegant and easy to use, was to simply read if the Alt key was pressed while clicking the arrow next to a parent.
* If the Alt key is held while clicking Expand, then every child of the parent were also expanded.
* If the Alt key is held while clicking Collapse then every child of the parent were also collapsed.

MineBill:
Fixes #1742.

The ActorTreeNode node now uses the PrefabObjectID when saving to the project cache if the Actor has a prefab link and is not in a scene, indicating that it is in a prefab window. We need to check for the scene explicitly because otherwise, PrefabObjectID will be used inside a scene too and the actors of the prefab will overwrite the project cache and as such collapsing/expanding nodes in the scene will affect the nodes in the prefab window and vice versa.

Also implements the Alt key functionality requested from that issue.

### DIFF
--- a/Source/Editor/GUI/Tree/TreeNode.cs
+++ b/Source/Editor/GUI/Tree/TreeNode.cs
@@ -776,11 +776,20 @@ namespace FlaxEditor.GUI.Tree
                 // Check if mouse hits arrow
                 if (_mouseOverArrow && HasAnyVisibleChild)
                 {
-                    // Toggle open state
-                    if (_opened)
-                        Collapse();
+                    if (ParentTree.Root.GetKey(KeyboardKeys.Alt))
+                    {
+                        if (_opened)
+                            CollapseAll();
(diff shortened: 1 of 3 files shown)


madsfs:
Not sure if this is also relevant, but it is at least adjacent; https://github.com/FlaxEngine/FlaxEngine/issues/1741

PR #1759, opened 2023-10-21
Visject: Comment order improvements

### LINKED ISSUE #1643: comments in graphs have wrong order when added
**Issue**
any new comment is puthed to background without a way to move it to foreground

**Steps to reproduce:**

place comment in material graph.
then place other comment.
move on top of other u should see something like this
image

**Flax version:**

master

Chikinsupu:
This PR:

1. Adds the ability to manually order comments
2. Puts new comments on top of older comments
3. Puts new comments below older comments IF the older comments are children (see feature preview 3)
4. Fixes #1643 

**Important Notes and Testing**
This PR adds a value to the surface comment archetype to serialize their order value.
When opening a project with older comments, it won't break anything. In fact, it even makes them compatible with this system and gives the older comments an order value and then puts them in the back. A warning gets thrown that 4 values are expected, but only 3 are present, other than that nothing else happens.

Opening a project with the newer comments in an older Flax editor version also won't break anything, just throw warnings. The visject serializer already does the heavy lifting in this regard.

### Feature Preview
**1) The big thing - Order menu**
OrderComments
This adds a context menu child menu to the surface comment context menu, which adds the ability to order comment nodes, like in the older office programs e.g. power point

Note 1) I had 2 designs for this:
Design A:
FlaxEditor_yhgWAu2TUF

Design B:
FlaxEditor_SEnDKN9Vg2

Ultimately i went with Design B, since this reflects more how other programs did it back in the past. And because this feature probably won't be used as much.
Of course a design based on the more modern office programs would be a lot nicer. But as of now context menu child menus don't have an exposed click callback. Might be something for a future PR.
image

Note 2) I think this feature could actually be implemented for all nodes, not just the comments. It would be rather easy to do so. But i would like to have some feedback on that (and note 1) first, especially from @mafiesto4 

**2) Comments spawn on top of other comments now**

Before:
CommentOnTopBefore

After:
CommentOnTopAfter

**3) Commenting other comments puts the parent comment in the back of the child comments, but not any further**

Before:
CommentCommentsBefore

After:
CommentCommentsAfter
This one is actually kinda hard to showcase, but trust me, it's better now :)


NoriteSC:
the expanding with
image
it is not good if u can add 8 directions for expanding the coments

Chikinsupu:
> the expanding with image it is not good if u can add 8 directions for expanding the coments

Is that a request?

I agree, that i am also not a fan of this one button in the corner to change the comments size.
I could add a more intuitive way of doing so. But i would do that in another PR, not this one.

NoriteSC:
> Is that a request?

yea
> I agree, that i am also not a fan of this one button in the corner to change the comments size. I could add a more intuitive way of doing so. But i would do that in another PR, not this one.

why not in this pr ?
the changes are small

Chikinsupu:
> why not in this pr ?
the changes are small

Might be, but this PR is about the order of comments. Not general comment improvements. I like to focus on specific things in my PRs. That being said, creating a new PR for better comment resizing, is not a big deal ^^

PR #1760, opened 2023-10-21
Fix particles effect not being able to call play to play again if islooping is false

Tryibion:
I think the better solution is my first commit, but calling stop resets the simulation which then causes the editor (non-play mode) to repeat playing the simulation for non-looping effects.


schmidt-florian (approved):
I tested this PR and can confirm that it fixes my problem. Non-looping particle emitters no longer need an extra stop to start even though they have already finished playing.

Issue #1788, opened 2023-10-24
UI Actor Setup: Odd Workflow


madsfs:
I just started poking around in the Flax UI setup, as in, making UI for your games.
I spawned a UI Canvas Actor and as I exhausted every option, I found nowhere to scale the UI, something fundamental for any UI (how does it scale and relative to what?).

Turns out that you have to add a second actor as a child, which is called a UI Control.
Then you have to set the type of that Actor to a CanvasScaler.

There are some semantic issues here. The CanvasScaler is listed as a UI Control, however when you look at the existing list of UI Controls what you find are:
- AlphaPanel
- BlurPanel
- Border
- Button
- CanvasScaler
- Checkbox
- ContainerController
- ContentSearchWindow
- DropPanel
- Dropdown
- GridPanel
- HorizontalPanel
- Image
- Label
- Panel
- Progressbar
- RenderToTextureControl
- RichTextBox
- SearchResultTreeNode
- Spacer
- TextBox
- TilesPanel
- UniformGridPanel
- VerticalPanel

The CanvasScaler is not so much a UI Control, as it is a set of properties of the UI itself. It would surprise me if 9/10 uses of the UI Canvas does not include using a UI Scaler as well. Which means that my suggestion would be:
Merge the UI Scaler into the UI Canvas or at the very least, when you Create a UI Canvas Actor, just add a CanvasScaler Actor child by default. The few times where someone does not need the scaler they can delete it although merging the Scaler properties into the UI Canvas actor itself would be preferable.

PR #1794, opened 2023-10-24
Move canvas scalar to UICanvas and Canvas Root Control.

Tryibion:
It makes sense for the canvas scalar to be on the canvas and not its own actor. #1788


mafiesto4:
What about simply adding UI Canvas Scaler when adding UI Canvas in Editor via `UICanvasNode.PostSpawn`?

Tryibion:
I mean wither way would be fine, but the scene does look better not having to have everything under UICanvas and a Canvas Scalar.

image

PR #1804, opened 2023-10-25
Fix some Center geometry import issues. Add being able to shift mesh to local origin.

Tryibion:
Center geometry was not scaling or rotating correctly, this should fix those issues (I am sure there are edge cases). This also adds an option called `UseLocalOrigin` which moves the mesh to use it's local origin.

### DIFF
--- a/Source/Engine/Graphics/Models/ModelData.h
+++ b/Source/Engine/Graphics/Models/ModelData.h
@@ -103,7 +103,7 @@ public:
     /// <summary>
     /// Meshes scaling.
     /// </summary>
-    Vector3 Scaling = Vector3(1);
+    Vector3 Scaling = Vector3::One;
 
 public:
     /// <summary>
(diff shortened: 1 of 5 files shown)


Source/Engine/Graphics/Models/ModelData.h
```diff
@@ -90,6 +90,21 @@ class FLAXENGINE_API MeshData
...
+
+    /// <summary>
+    /// Orientation for this mesh at it's local origin.
+    /// </summary>
+    Quaternion OriginOrientation = Quaternion::Identity;
+
+    /// <summary>
+    /// Meshes scaling.
+    /// </summary>
+    Vector3 Scaling = Vector3(1);
```
mafiesto4, line 106: `Vector3::One`

Source/Engine/Tools/ModelTool/ModelTool.Assimp.cpp
```diff
@@ -1,5 +1,6 @@
 // Copyright (c) 2012-2023 Wojciech Figat. All rights reserved.
 
+#include "OpenFBX/ofbx.h"
```
mafiesto4, line 3: Probably unwanted include here inside Assimp impl.

Issue #1840, opened 2023-10-30
Some mouse buttons don't work


RuanLucasGD:
**Issue description:**
Extended1 and Extended2 don't return true when press.
Try this code:
```
Debug.Log(Input.Mouse.GetButtonDown(MouseButton.Extended2));
```
**Flax version:**
1.7 master

mafiesto4:
On Windows it maps to `XButton2` from XInput interface.

RuanLucasGD:
@mafiesto4 i'm using linux

mafiesto4:
Not yet implemented on Linux: https://github.com/FlaxEngine/FlaxEngine/blob/master/Source/Engine/Platform/Linux/LinuxWindow.cpp

PR #1846, opened 2023-10-31
Add extended buttons support for linux

RuanLucasGD:
Fix this problem of inputs on linux: #1840

### DIFF
--- a/Source/Engine/Platform/Linux/LinuxWindow.cpp
+++ b/Source/Engine/Platform/Linux/LinuxWindow.cpp
@@ -594,6 +594,12 @@ void LinuxWindow::OnButtonPress(void* event)
 	case Button3:
 		mouseButton = MouseButton::Right;
 		break;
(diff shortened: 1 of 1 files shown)


Source/Engine/Platform/Linux/LinuxWindow.cpp
```diff
@@ -594,6 +594,12 @@ void LinuxWindow::OnButtonPress(void* event)
 	case Button3:
 		mouseButton = MouseButton::Right;
 		break;
+	case 8:
```
RuanLucasGD, line 597: I don't know why this, but Linux API don't have constants to extended buttons, only Button1, 2, 3, 4, 5 but don't have for 8 and 9.

mafiesto4: Yeah typical of `X11` API :D thx for this improvement

Issue #1787, opened 2023-10-24
Editor: Dynamic Initialization of Lists And Multi-Add


madsfs:
So as you can see in the video below, if you try to drag an element to a list, it will not expand to accomodate and assign the element automatically. You must set a specific size of the list and then drag an element to a separate slot.
In my example I'm only moving 3 elements to this list, so it's not that big a deal to assign them one at a time, however it is still tedious. This would be an issue if you had 10, 20 or 100 elements you needed to add to a list this way (for any reason).

My proposal is that whenever elements are dragged over a list like this, expand the list with the number of slots needed to add the elements and then append them to the end of the list in the order they appeared (as you can see me attempt in the video at the start as well).

This would be a great QoL improvement for the editor.

https://github.com/FlaxEngine/FlaxEngine/assets/113017777/39fa9cda-c612-4ee7-8492-333c372f422f

MineBill:
I have an initial implementation for this and i would like some feedback on it:

https://github.com/FlaxEngine/FlaxEngine/assets/30367251/df6c2b31-85c5-478f-9deb-1e366451721e

Does this look intuitive to you? What if you were a new user?

madsfs:
> I have an initial implementation for this and i would like some feedback on it:
> 2023-10-30.22-35-43.mp4
> 
> Does this look intuitive to you? What if you were a new user?

So if I am not mistaken it seems like when you drag over, the entries get replaced?
I would expect that entries just get appended, not replaced.

mafiesto4:
Added in #1850

PR #1850, opened 2023-10-31
Allow dragging items to lists and arrays in the properties window.

MineBill:
Implements #1787

This PR allows you quickly drag and drop items in lists and arrays:

https://github.com/FlaxEngine/FlaxEngine/assets/30367251/8cb1f3d8-6c1e-4868-822d-b55a9b3932c7

Also supports nested lists:

https://github.com/FlaxEngine/FlaxEngine/assets/30367251/3e980288-d9b3-40e7-b368-f6f09b0d319f

I took most of the item "validation"/"conversion" code from `AssetPicker` and placed it in it's own class `AssetPickerValidator` to reuse it. I'm not a big fan of the name, so feel free to suggest something else.

Dragging items onto a list/array will append them at the end of it. When dragging multiple items, only the correct items (if any) will be added.

Some of the code handles resizing the collection instead of calling the `Resize` method of the `CollectionEditor`. That's because `Resize` uses `SetValue` to update the collection, which delays the actual update until the next frame and that doesn't work for this case.

### DIFF
--- a/Source/Editor/CustomEditors/Editors/AssetRefEditor.cs
+++ b/Source/Editor/CustomEditors/Editors/AssetRefEditor.cs
@@ -71,7 +71,7 @@ namespace FlaxEditor.CustomEditors.Editors
                 {
                     // Generic file picker
                     assetType = ScriptType.Null;
-                    Picker.FileExtension = assetReference.TypeName;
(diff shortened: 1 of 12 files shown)


mads-fs:
Very good work. Thank you!

PR #1883, opened 2023-11-06
Refactor Vehicle System

RuanLucasGD:
This simplify vehicle system removing duplicated code. 
Also adds a property to change gears ratios on WheeledVehicle->GearBox->ForwardGearsRatio


Source/Engine/Physics/PhysX/PhysicsBackendPhysX.cpp
```diff
@@ -3083,6 +3083,147 @@ int32 PhysicsBackend::MoveController(void* controller, void* shape, const Vector
...
+    differential4WData.mFrontRearSplit = settings.FrontRearSplit;
+    differential4WData.mFrontLeftRightSplit = settings.FrontLeftRightSplit;
+    differential4WData.mRearLeftRightSplit = settings.RearLeftRightSplit;
+    differential4WData.mCentreBias = settings.CentreBias;
+    differential4WData.mFrontBias = settings.FrontBias;
+    differential4WData.mRearBias = settings.RearBias;
+    return differential4WData;
+}
+
+PxVehicleDifferentialNWData CreatePxVehicleDifferentialNWData(const WheeledVehicle::DifferentialSettings settings, const Array<WheeledVehicle::Wheel*, FixedAllocation<PX_MAX_NB_WHEELS>> wheels)
```
mafiesto4, line 3099: Pass larger value types such as `DifferentialSettings` via reference - `const DifferentialSettings&`. The same for array to be passed as reference to avoid memory copies.

RuanLucasGD: Can verify again ?

mafiesto4: Missing reference character `&` so the structure will be passed as pointer under the hood without memory copy over the stack.

RuanLucasGD: @mafiesto4 yes, i made this fix on last commit, can check ?

Issue #1585, opened 2023-10-02
Serializing complex generic class into json throws incompatibility error on deserialization


schmidt-florian:
**Issue description:**
I tried to serialize the ``class MySettings``. 

```csharp
public class MySettings : BaseSettings<BaseClass> { }

public class BaseSettings<T>
{
    public List<T> Data = new List<T>();
}

public class BaseClass { }

public class DrivenA : BaseClass { }

public class DrivenB : DrivenA { }
```

On deserialization, this error is thrown:

```
[Warning] Exception has been thrown during FlaxEngine.JsonAsset, 64a5ad30479d1d42ede1b690104b634a, D:\Projects/Flax/Test Blank/Content/My Settings.json. Type specified in JSON 'Game.DrivenB, Game.CSharp, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' is not compatible with 'Game.BaseClass, Game.CSharp, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'. Path 'Data[0].$type', line 4, position 39.
```

BUT this is only thrown if flax had it freshly recompiled. If for example: You save the settings. Close Flax, reopen flax, you can open the json asset editor with no error. If you close the json asset editor, hit recompile, you will be prompted with the error again if you try to open the json asset.

In the class ``Newtonsoft.Json.Serialization.JsonSerializerInternalReader@826`` is the check that failed:
The check ``IsAssignableFrom`` fails and throws this error. If you inspect the assemblies of both types, even they have the same name, a comparison check will return false. It looks like the types have 2 different assemblies, even they should have the same assembly, maybe from recompiling? 

**Steps to reproduce:**

- Open attached project.
- Open Content/My Settings
- Click button to add elements, click save (Inspect Elements under Data)
- Close window
- Open Content/My Settings
- No Error, you can inspect the elements, everything is fine.
- Close window
- File -> Recompile Scripts
- Open Content/My Settings
- The Error should appear
- If you add now elements, save and close and open flax again, you can open the settings without error but only as long you don't recompile the scripts.

**Minimal reproduction project:**
[Test Blank.zip](https://github.com/FlaxEngine/FlaxEngine/files/12785198/Test.Blank.zip)

**Flax version:**
master 5fe135e

Menotdan:
I had a similar bug actually, but it seems to go away for me permanently after re-launch. Recompiling scripts does not re-trigger my issue.

schmidt-florian:
[...]

PR #1888, opened 2023-11-07
Fixed issue involving stale scripting assemblies in FlaxEngine.Json dynamic type resolution

SinnersSum:
Added ExtendedSerializationBinder to handle dynamic type resolution
Added callback to clear serializer cache on scripting assembly reload Added low-cost mechanism to invalidate the SerializerCache after domain reload

https://github.com/FlaxEngine/FlaxEngine/issues/1585 should be solved by this change. In order to avoid old versions of scripting assemblies being loaded, a custom `ISerializationBinder` now preferentially checks known up-to-date scripting assemblies from NativeInterop. I'm not sure if this would be the preferred way to find scripting assemblies, but it is kept up to date, and does appear to contain all relevant assemblies.

Unfortunately. the core problem is likely deeper, as the reason `DefaultSerializationBinder` was resolving out of date assemblies is that the domain never releases old copies of scripting assemblies. I'm not 100% sure but https://github.com/JamesNK/Newtonsoft.Json/issues/2414 seems like it might be related.

### DIFF
--- a/Source/Engine/Engine/NativeInterop.Unmanaged.cs
+++ b/Source/Engine/Engine/NativeInterop.Unmanaged.cs
@@ -1022,6 +1022,8 @@ namespace FlaxEngine.Interop
                 pair.Value.Free();
             classAttributesCacheCollectible.Clear();
 
+            FlaxEngine.Json.JsonSerializer.ResetCache();
+
             // Unload the ALC
             bool unloading = true;
             scriptingAssemblyLoadContext.Unloading += (alc) => { unloading = false; };
--- a/Source/Engine/Engine/NativeInterop.cs
+++ b/Source/Engine/Engine/NativeInterop.cs
@@ -278,44 +278,65 @@ namespace FlaxEngine.Interop
             if (typeCache.TryGetValue(typeName, out Type type))
                 return type;
 
-            type = Type.GetType(typeName, ResolveAssemblyByName, null);
+            type = Type.GetType(typeName, ResolveAssembly, null);
             if (type == null)
-            {
-                foreach (var assembly in scriptingAssemblyLoadContext.Assemblies)
-                {
-                    type = assembly.GetType(typeName);
-                    if (type != null)
-                        break;
-                }
-            }
+                type = ResolveSlow(typeName);
 
             if (type == null)
             {
-                string oldTypeName = typeName;
+                string fullTypeName = typeName;
                 typeName = typeName.Substring(0, typeName.IndexOf(','));
-                type = Type.GetType(typeName, ResolveAssemblyByName, null);
+                type = Type.GetType(typeName, ResolveAssembly, null);
                 if (type == null)
-                {
-                    foreach (var assembly in scriptingAssemblyLoadContext.Assemblies)
-                    {
-                        type = assembly.GetType(typeName);
-                        if (type != null)
-                            break;
-                    }
-                }
-                typeName = oldTypeName;
+                    type = ResolveSlow(typeName);
+
+                typeName = fullTypeName;
             }
 
             typeCache.Add(typeName, type);
 
             return type;
+
+            /// <summary>Resolve the type by manually checking every scripting assembly</summary>
+            static Type ResolveSlow(string typeName) {
+                foreach (var assembly in scriptingAssemblyLoadContext.Assemblies) {
+                    var type = assembly.GetType(typeName);
+                    if (type != null)
+                        return type;
+                }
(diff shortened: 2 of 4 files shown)


SinnersSum:
I can see why this slid under the radar for so long, seems like dynamic type resolution (and thus `ISerializationBinder`) isn't invoked very frequently. Made testing mercifully easy, as really only my error-case and UIBrushes are serialized in this way.

Another thing that concerns me is that this case was broken, but the general case seemingly wasn't and I'm not really sure why. It will probably be necessary eventually to figure out how to flush old assemblies from the context or to otherwise rebuild the context. For now though the custom binder seems to resolve the immediate issue.

mafiesto4:
Thanks for this PR, I'll review and test it. It looks like during Mono -> .NET 7 transition we disabled clearing Json serializer cache upon hot-reload in Editor which might be related to this issue: https://github.com/FlaxEngine/FlaxEngine/commit/96dc279ebdc11a2ddf76634a59357a0ad573889b#diff-4f3de17fc323a9ab118a9f5a4c0f8b0e46a6927fd21b93bbb7756c6879c489ad

mafiesto4:
Test fails on Linux

SinnersSum:
I notice with other PRs that if any test fails, it's usually the linux test, however I notice something weird in the windows logs
```
Testhost process for source(s) 'D:\a\FlaxEngine\FlaxEngine\Binaries\Tests\Flax.Build.Tests.dll' exited with error: A fatal error was encountered. The library 'hostpolicy.dll' required to execute the application was not found in 'D:\a\FlaxEngine\FlaxEngine\Binaries\Tests\'.
Failed to run as a self-contained app.
  - The application was run as a self-contained app because 'D:\a\FlaxEngine\FlaxEngine\Binaries\Tests\testhost.runtimeconfig.json' was not found.
  - If this should be a framework-dependent app, add the 'D:\a\FlaxEngine\FlaxEngine\Binaries\Tests\testhost.runtimeconfig.json' file and specify the appropriate framework.
. Please check the diagnostic logs for more information.
```
Are the windows tests just not even running?

SinnersSum:
> Thanks for this PR, I'll review and test it. It looks like during Mono -> .NET 7 transition we disabled clearing Json serializer cache upon hot-reload in Editor which might be related to this issue: https://github.com/FlaxEngine/FlaxEngine/commit/96dc279ebdc11a2ddf76634a59357a0ad573889b#diff-4f3de17fc323a9ab118a9f5a4c0f8b0e46a6927fd21b93bbb7756c6879c489ad

It doesn't seem like the issue was actually with the cache, but rather that out of date assemblies are remaining in the domain, and so the `DefaultSerializationBinder` sometimes finds those assemblies during dynamic type resolution.

> Test fails on Linux

I don't have a linux machine unfortunately, but I have a suspicion that the windows tests simply aren't being run so I'll try and run the tests myself on my windows machine. Any tips on how to do that by the way? I'm reading the workflow file and trying to follow it, but `FlaxTests.exe` simply isn't being built.

SinnersSum:
Ok, by following the commands in the workflow file I was able to reproduce the error, only problem now is setting up a debugging environment for the test so I can actually see what's happening.

I think the flax tests could probably use an actual target in the generated solution, or even better a proper tests project that can be executed/extended easily. I'm not even sure where the source files for this test that's failing are, so it's hard to diagnose the issue.

SinnersSum:
NVM couldn't get the tests running through VS, I've tried a hundred things, It's just one issue after another. https://github.com/FlaxEngine/FlaxEngine/issues/1892

SinnersSum:
I can't readily debug the tests myself (due to aforementioned issue), but I have been able to reproduce them with a script. Unfortunately —or maybe fortunately— the tests all pass even though this is essentially the same code verbatim. Obviously, the issue is somehow related to the assembly it's in —or perhaps related to the test-harness.

[UndoTest.zip](https://github.com/FlaxEngine/FlaxEngine/files/13289660/UndoTest.zip)
☝🏿 Add this script to a scene, click any of the buttons to run the associated tests.

SinnersSum:
I found the source of the issue: The the scripting assembly load-context wasn't getting initialized in the test harness, so `ResolveScriptingAssemblyByName` was failing internally when it attempted to check the scripting assemblies. Fixing that was relatively simple once I had located it.

I've run the tests locally on my machine, and all have passed. Ready for a second review.

mafiesto4:
Thanks for adding this!

PR #1906, opened 2023-11-09
Rebuild navigation after apply changes in Navigation asset

RuanLucasGD:


mafiesto4:
I would add check if Editor allows updating NavMesh (eg. disabled during specific editor state or via Editor Option to not auto-build navmesh):
```cpp
if (!Editor::IsPlayMode && Editor::Managed->CanAutoBuildNavMesh())
{
...
}
```

RuanLucasGD:
@mafiesto4 fixed

Issue #1329, opened 2023-08-20
Auto create prefab when import models


RuanLucasGD:
I started a game project and one of the things that is bothering me is having to import a lot of models. Flax is unable to preserve the center of model geometry.

That way I need to export all the models separately, and import them into the editor. And then I need to reassemble the whole model.

Please add a way to import the model and generate a prefab with the mesh structure.

related to https://github.com/FlaxEngine/FlaxEngine/issues/1023

**Flax version:**
1.6

mafiesto4:
Added in 63773f2ddf71418cf7fb6b86e1a7e33dcb95c9dc.

PR #1909, opened 2023-11-10
Add creating prefab from models.

Tryibion:
This adds an easy way to create a prefab from 1 or more models and adds the options to use the "before imported" positions and can center the models. I am sure there are plenty of edge cases and I only tested a few... so I am sure it could use more testing. This only really works well with models using local transform and/or being centered. Let me know if there is an easier way besides transferring the data through the model.

Simple create Prefab:

https://github.com/FlaxEngine/FlaxEngine/assets/71274967/36c24142-5aad-4518-827f-f0f303ef4983

Create prefab with original positions:

https://github.com/FlaxEngine/FlaxEngine/assets/71274967/6e8df2c2-7b48-4454-b64c-3086b0af9562

Create prefab with original positions and center models:

https://github.com/FlaxEngine/FlaxEngine/assets/71274967/854aeab8-9d33-4219-8516-36577144bc00

### DIFF
--- a/Source/Editor/Content/Proxy/ModelProxy.cs
+++ b/Source/Editor/Content/Proxy/ModelProxy.cs
@@ -67,10 +67,10 @@ namespace FlaxEditor.Content
             });
             var createPrefabButton = menu.AddButton("Create Prefab");
             createPrefabButton.CloseMenuOnClick = false;
-            createPrefabButton.Clicked += () => OnCreatePrefab(createPrefabButton, item);
+            createPrefabButton.Clicked += () => OnCreatePrefabClicked(createPrefabButton, item);
         }
 
-        private void OnCreatePrefab(ContextMenuButton button, ContentItem item)
+        private void OnCreatePrefabClicked(ContextMenuButton button, ContentItem item)
         {
             var popup = new ContextMenuBase
             {
--- a/Source/Editor/Content/Proxy/ModelProxy.cs
+++ b/Source/Editor/Content/Proxy/ModelProxy.cs
@@ -65,6 +65,146 @@ namespace FlaxEditor.Content
                     collisionDataProxy.CreateCollisionDataFromModel(model);
                 }
             });
+            var createPrefabButton = menu.AddButton("Create Prefab");
+            createPrefabButton.CloseMenuOnClick = false;
+            createPrefabButton.Clicked += () => OnCreatePrefab(createPrefabButton, item);
+        }
+
+        private void OnCreatePrefab(ContextMenuButton button, ContentItem item)
+        {
+            var popup = new ContextMenuBase
+            {
+                Size = new Float2(170, 100),
+                ClipChildren = false,
+                CullChildren = false,
+            };
+            popup.Show(button, new Float2(button.Width, 0));
+
+            var centerModelsLabel = new Label
+            {
+                Parent = popup,
+                AnchorPreset = AnchorPresets.TopLeft,
+                Text = "Center Models",
+                HorizontalAlignment = TextAlignment.Near,
+            };
+            centerModelsLabel.LocalX += 10;
+            centerModelsLabel.LocalY += 10;
+
+            var centerModelsCb = new CheckBox()
+            {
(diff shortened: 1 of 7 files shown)


Menotdan:
Is it possible to have center geometry and use original position on by default? Anyways, thank you for this, I will test it out later and give more feedback and edge case testing since this is definitely something I want in the editor.

mafiesto4:
Very nice! One thing tho: what if we add support for creating prefabs during model import instead? Because when importing model we have all the information from the source asset (like that Original Transform you've exported from it) + we could add support for adding camera/lights/stuff into that prefab to be auto-imported (from `.fbx` file).

RuanLucasGD:
#1329

Tryibion:
> Very nice! One thing tho: what if we add support for creating prefabs during model import instead? Because when importing model we have all the information from the source asset (like that Original Transform you've exported from it) + we could add support for adding camera/lights/stuff into that prefab to be auto-imported (from `.fbx` file).

Is there a good place to do this in the code? I could really find a good spot, so they is why I did it this way, but if there is a recommended spot, I can move it.

Menotdan:
> We could wrap this inside `ModelTool` as it already handles more than model data but also textures, materials, and such.
> 
> So we could:
> 
>     * Add a new `ModelTool::ModelType::Prefab` and new options for prefabs like `Import Lights`/`Import Cameras`
> 
>     * Add `CreateAssetResult::Skip` so native importer could skip creation of binary asset so we could create a prefab file instead manually within `ImportModelFile::Import`
> 
>     * `ModelTool::ImportModel` when called with option `Prefab` should import all data types (Geometry, Nodes, Textures, etc.)
> 
>     * Initially let's focus on static models only - in future we could add support for mixing static and animated models inside the same import flow
> 
>     * Create actor structure with models from file (handle transformation origins, etc.)
>       
>       * Root actor could be a new `ModePrefab` actor (or similar name) and it would store the import options to allow easy reimporting
>       * In future, add lights/cameras/stuff from source model
>       * We could even import animations of those objects as Scene Animation assets
> 
>     * Add option for reimporting the prefab to reflect changes in the source asset (eg. via custom button on that prefab root object)
> 
>     * Update Flax Docs about models importing
> 
>     * Add dedicated tutorial in Flax Docs for importing model files as prefabs
> 
> 
> It might be a little complex task and will take some time to polish it so let's do it on `1.8` branch. @Tryibion what do you think about it? If you need assistance I can write this integration core so you could focus on model transformations solely.

Is there a way we can have the current iteration of this PR on master or something? There are many many people who need functionality like this to work on their projects, one of whom I've been helping on stream.
The only issue I see is that this PR updates the whole model asset object which could be problematic..

Menotdan:
@mafiesto4 How is progress on this, I'm working with another new user who also really would like this feature soon.

Menotdan:
Also, would it be possible to create some feature to have a reimportable asset file that imports all the split meshes without having to go into the main asset and split object index -1 again and then reimport with some iffy behavior... and would be nice to edit all the import settings for all the meshes there as well.

But maybe there is a better way to approach this.. I'm not sure. I just know that many people want this feature and it is a massive block in their workflows from some watching people and asking around and reading suggestions.

mafiesto4:
@Menotdan I want to push `1.7.1` patch update before working on any major features so I'd like to start working on this next week. It should take just a few days to be ready.

Menotdan:
> @Menotdan I want to push `1.7.1` patch update before working on any major features so I'd like to start working on this next week. It should take just a few days to be ready.

Ah, alright. I'm actually working on a bigger issue with a bigger problem list because I sorta feel like the way this functionality is implemented is built on some crumbly parts of model importing, in that they are annoying to use and have a major impact on workflow beyond the problem this addresses.. I still like this idea though.

mafiesto4:
Now, we can create Prefab out of the imported model: 63773f2ddf71418cf7fb6b86e1a7e33dcb95c9dc - all model transformations magic can happen within model imported without a need to modify assets format.

PR #1914, opened 2023-11-10
Don't hard-code appdata path for linux anymore.

MineBill:
This should've been included with #1878 but i completely forgot to update them. I searched the entire repo but didn't find any other hard-coded paths to `.local/share`.


mafiesto4:
Does the Rider integration work fine with this change?

MineBill:
It does. In fact, that's the reason i found out about these paths. With #1878, `localAppDataPath` already contains the appropriate path, either `~/.local/share` if the user didn't set `XDG_DATA_HOME` or the path contained in `XDG_DATA_HOME`, so the paths generated before this pr where something like `/home/minebill/.local/share/.local/share/...` and flax failed to find Rider.

PR #1923, opened 2023-11-12
Fix bounds center for Rigidbodies

RuanLucasGD:
Rigid bodies has a problem like #1365.

I made some tests and works fine, sorry if i forgot something.

### DIFF
--- a/Source/Engine/Physics/Actors/RigidBody.cpp
+++ b/Source/Engine/Physics/Actors/RigidBody.cpp
@@ -327,12 +327,10 @@ void RigidBody::OnColliderChanged(Collider* c)
 
 void RigidBody::UpdateBounds()
 {
-    void* actor = _actor;
-    if (actor && PhysicsBackend::GetRigidActorShapesCount(actor) != 0)
-        PhysicsBackend::GetActorBounds(actor, _box);
-    else
-        _box = BoundingBox(_transform.Translation);
-    BoundingSphere::FromBox(_box, _sphere);
+    const float radius = GetScale().MaxValue();
+    const Vector3 position = _transform.Translation;
+    _sphere = BoundingSphere(position, radius);
+    _box = BoundingBox::FromSphere(_sphere);
 }
 
 void RigidBody::UpdateScale()


MineBill:
With PRs like this, it's a good idea to include some screenshots comparing the before and after, to help mafi merge them faster.

RuanLucasGD:
**Focusing on a Rigidbody before:**
https://github.com/FlaxEngine/FlaxEngine/assets/79365912/111c5f4a-7a90-4aae-8ad3-93098607faca

**Focusing on a Rigidbody now:**
https://github.com/FlaxEngine/FlaxEngine/assets/79365912/17031977-ba53-449c-8d98-42bfde646088

Source/Engine/Physics/Actors/RigidBody.cpp
```diff
@@ -327,12 +327,10 @@ void RigidBody::OnColliderChanged(Collider* c)
 
 void RigidBody::UpdateBounds()
 {
-    void* actor = _actor;
-    if (actor && PhysicsBackend::GetRigidActorShapesCount(actor) != 0)
-        PhysicsBackend::GetActorBounds(actor, _box);
```
NoriteSC, line 332: u know this is getting a BB from PhysicsBackend with is accurate where rb actor is 

new code is ignoring the PhysicsBackend bounding box

Source/Engine/Physics/Actors/RigidBody.cpp
```diff
@@ -327,12 +327,10 @@ void RigidBody::OnColliderChanged(Collider* c)
...
-    else
-        _box = BoundingBox(_transform.Translation);
-    BoundingSphere::FromBox(_box, _sphere);
+    const float radius = GetScale().MaxValue();
```
NoriteSC, line 330: there i a need to sync the BB with PhysicsBackend first

RuanLucasGD: But PhysicsBackend::GetRigidActorShapesCount is making this problem, before CharacterController had this problem with BB too and got fixed removing this and adding this code: https://github.com/FlaxEngine/FlaxEngine/blob/9da686cd525c3ff76fc5b8722b697d26c64688a1/Source/Engine/Physics/Colliders/CharacterController.cpp#L244C32-L244C32

My changes is a "copy" of the CharacterController BB to fix this problem

PR #1929, opened 2023-11-13
Wake up rigidbody when set "IsKinematic" to false

RuanLucasGD:


Source/Engine/Physics/Actors/RigidBody.cpp
```diff
@@ -35,7 +35,11 @@ void RigidBody::SetIsKinematic(const bool value)
         return;
     _isKinematic = value;
     if (_actor)
+    {
         PhysicsBackend::SetRigidDynamicActorFlag(_actor, PhysicsBackend::RigidDynamicFlags::Kinematic, value);
+        if (!value && _isActive)
```
mafiesto4, line 40: Add check for `GetStartAwake()` so we don't wake up rigidbodies that are configured to favor sleeping.

RuanLucasGD: It's right now?

PR #1938, opened 2023-11-14
Avoid crash when try build navmesh with null scene

RuanLucasGD:


Source/Engine/Navigation/NavMeshBuilder.cpp
```diff
@@ -1098,6 +1104,17 @@ void NavMeshBuilder::Build(Scene* scene, float timeoutMs)
 
 void NavMeshBuilder::Build(Scene* scene, const BoundingBox& dirtyBounds, float timeoutMs)
 {
+    if (!scene)
+    {
+        LOG(Warning, "Could not generate navmesh without scene.");
+        return;    
+    }
+    if (!&dirtyBounds)
```
mafiesto4, line 1112: That's always valid so check can be ignored for bounds.

mafiesto4 (approved):

PR #1949, opened 2023-11-17
Added Actor.GetPrefabRoot().

Menotdan:
Maybe this could be added to `SceneObject` instead, but I think this makes the most sense.


mafiesto4:
Yeah, we could move it to `SceneObject` to be usable by Scripts too.

PR #1955, opened 2023-11-18
Spawn actors on SceneTree improviments

RuanLucasGD:
Hello, i made some usability improviments on SceneTree window (scene and prefabs window).

Before you could only create a new actor, for small scenes this is nice, but if you have large hierarchy and want to create a new actor is a bit hard add a new actor without a parent, because when you add a actor (pressing over another actor) the new actor is automatically assigned as child of selected actor and you need to remove the child. Make this often is bored. 

I made a new option to context menu to create a new actor without parent and another to create as child, this fix the problem but i needed make a small change on all actors. Now the ActorContextMenu don't need the `"New/"` on start of the context path because this is auto generated by ContextMenu.

**before:**

https://github.com/FlaxEngine/FlaxEngine/assets/79365912/4e28203c-4575-4eed-837a-98c26dde4309

**After:** 

https://github.com/FlaxEngine/FlaxEngine/assets/79365912/8c1b2931-22c8-49bc-b414-4393eee2dc20

I tried add an option to add new actor as parent of selected actors to make groups easily but don't worked with Undo/Redo.

Dependency:

#1559


RuanLucasGD:
Anyone know how can i remove these unnecessary commits ? ;-;

mafiesto4:
This breaks existing functionality into very different way, I'm not sure if it won't be too confusing. You can right-click on Scene Tree area below with the context menu to add loose actors.

RuanLucasGD:
@mafiesto4 yea, but if you have a hierarchy with many actors the actual way is not very nice, It's just not comfortable but if you want i can close this.

Tryibion:
I think it would be better to have an add button near the search bar that brings up the same menu and will add it to the first scene.

RuanLucasGD:
Yes i think this can be made using a better way, i'll close this.

PR #1961, opened 2023-11-19
Implement SetNodeTransform in company with GetNodeTransform

HydrogenC:


HydrogenC:
And maybe expose the OnAnimatedUpdated function to allow maunal updating after directly modifying NodesPose?

PR #1964, opened 2023-11-19
Add "HasParentTag" to actor

RuanLucasGD:


mafiesto4:
I'm afraid we might come to the point where we have too many APIs related to tags in Actor so the suggested new function is not that essential to be in core engine.

PR #1974, opened 2023-11-21
Remove ability in editor to delete content and source folders.

Tryibion:
This removes the ability inside of the editor to delete the content and source folders. This also limits the CM options on those folder nodes only to ones that make sense (removes delete, duplicate, rename, and copy options).


mafiesto4:
Can you change the naming to `CanDelete` and `CanRename` and make it public? So it matches the `CanHaveAssets` which already exists below.

Tryibion:
> Can you change the naming to `CanDelete` and `CanRename` and make it public? So it matches the `CanHaveAssets` which already exists below.

Done

PR #2002, opened 2023-11-27
UI Improvements

NoriteSC:
Clipping API changes
better control over how the clipping is applied

### DIFF
--- a/Source/Engine/UI/GUI/ContainerControl.cs
+++ b/Source/Engine/UI/GUI/ContainerControl.cs
@@ -29,7 +29,7 @@ namespace FlaxEngine.GUI
         [NoSerialize]
         protected bool _isLayoutLocked;
 
-        private bool _clipChildren = true;
+        private Clipping _clipChildren = Clipping.Inherit;
         private bool _cullChildren = true;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Source/Engine/UI/GUI/ContainerControl.cs
```diff
@@ -98,12 +97,19 @@ public bool IsLayoutLocked
...
+
+        /// <summary>
+        /// Controls how the clipping behavior of this Control.
+        /// Normally content that overflows the bounds of the Control continues rendering.
+        /// Enabling clipping prevents that overflowing content from being seen.
+        /// <para><b>Note:</b></para>
+        /// Controls in different clipping spaces can not be batched together, and so there is a performance cost to clipping.
+        /// Do not enable clipping unless a panel actually needs to prevent content from showing up outside its bounds.
+        /// </summary>
+        [EditorOrder(530), Tooltip("Controls how the clipping behavior of this Control.\nNormally content that overflows the bounds of the Control continues rendering.\nEnabling clipping prevents that overflowing content from being seen.\n\nNote:\nControls in different clipping spaces can not be batched together, and so there is a performance cost to clipping.\nDo not enable clipping unless a panel actually needs to prevent content from showing up outside its bounds.")]
```
Withaust, line 111: Maybe we should leave the original formatting and simply add new lines at the end of the original lines?

NoriteSC: u can't add new line string to a tooltip if u are asking about this
it just is not working

Withaust: I was talking about how the formatting used to look like: https://github.com/FlaxEngine/FlaxEngine/pull/2002/commits/437aa6cd094fdcb6ee1bc2b9e128b59fde0bab4a#diff-2e3b1425cc9e599c0dae72194347c78852075d8a8cff2eff1513fc0ea81db305R112

Source/Engine/UI/GUI/ContainerControl.cs
```diff
@@ -98,12 +97,19 @@ public bool IsLayoutLocked
...
+        public Clipping Clipping;
```
mafiesto4, line 112: Initialize it to the `Clipping.Inherit` so it's more explicit and add DefaultValue(Clipping.Inherit)` attribute so will have hint on modification in Editor.

NoriteSC: ok

Source/Engine/UI/GUI/ContainerControl.cs
NoriteSC: the tool tip is not showing this type of formating because of +

Withaust: Ah, I see, okay then.

Source/Engine/UI/GUI/ContainerControl.cs
NoriteSC: @mafiesto4  done

mafiesto4:
I think we gonna need to change the default clipping to be enabled as previously - otherwise many Editor UI parts have clipping bugs.

PR #2019, opened 2023-12-01
Add font fallback for text rendering

### LINKED ISSUE #1661: The editor font does not contain Chinese or Japanese
**Issue description:**

flax font
The font of the editor does not support languages such as Chinese and Japanese. 
Suggest adding fallback font function.
This is important for the localization of the game

**Flax version:**

master

HydrogenC:
Add font fallback for text rendering, enabling better support for CJK characters. Closes #1661. 

Showcase:  
Comp1
Comp3
Comp2

Work to do: 
- [X] Introduce per-control fallback control (disable font fallback for those controls that may never contain non-ASCII chars for better performance)
- [X] Add xml comments

Note: 
The changes isn't well tested, may expect some bugs. Also, help-wanted to make the code more organized and performant.  
In the pictures I used NotoSans-CJK as the fallback font, however it asks me to auth for the private LFS server when I try to upload it. So I have to add it to .gitignore temporarily to successfully push. 
/Content/Editor/Fonts/NotoSansSC-Regular.flax: [NotoSansSC-Regular.zip](https://github.com/FlaxEngine/FlaxEngine/files/13524270/NotoSansSC-Regular.zip)


HydrogenC:
> That's very cool! What about moving this into `Render2D.DrawText` instead of separate utility

It's a nice idea, I will work on it soon.

HydrogenC:
What is the fastest way to get the current game settings?

mafiesto4:
I simplified some code in https://github.com/FlaxEngine/FlaxEngine/commit/a2b8312fbaed18af9bad302fe15b6d21fd36988f and added docs in https://github.com/FlaxEngine/FlaxDocs/commit/b46cc5deae9d531b80e6573fb9d10696af4c1861.

Most importantly I removed `FontFallbackList` in favor of `FontAsset[] Fonts` inside `Font` class for simplicity. Then I added `Font*` pointer to `FontCharacterEntry` so all existing font codebase can be reused for this. Maybe baseline/height handling won't be that good but I suppose fallback font should not change text layout but fit into the primary font used to design UI.

I removed ability to disable it per Label/TextBox as I'm not sure if it's important feature. We can reimplement it in future when needed.

Thanks again for this feature!

Issue #1859, opened 2023-11-01
Camera: Set Eye Adaptation to 'Mode: None' by default


mads-fs:
I am using the 1.7 branch.

I just made a new scene, removed all lights, made an unlit material, put it on a box and then added that box to the scene. The result in the editor looks like this visual mess:
![image](https://github.com/FlaxEngine/FlaxEngine/assets/146040109/867fb879-b34c-4a35-8c3c-7efc2d23f4be)
An intense and distorted lenseflare effect. I was told that this is a screenspace effect that does not take lights into account at all. While I'm sure that kind of thing makes sense for a Screenspace post processing step, it is *very* unworkable as a default setting and assumes that you want the camera to work as a physically based camera with a lens always.

For comparison, here is an unlit material in a scene with no lights in Unity:
![image](https://github.com/FlaxEngine/FlaxEngine/assets/146040109/f9eec940-6463-423b-82bd-7eec47808f70)

It is not immediately obvious that this is a screenspace Post Processing step nor is it obvious that you'd have to dive into the GraphicsSettings.json file to go into the `Post Process Settings->Eye Adaptation->Mode: None` to get rid of it.

mafiesto4:
I agree those lens flares and bloom make it look bad. Even grid could have better colors since we render it before post-processing now. I wonder if we could make the default PostFX a little better to have better look for this case, otherwise we can just change default as you've suggested.

mafiesto4:
Fixed in #2030 and 1a68a52611c3c0e64111b1549e1833e34f151967. Now we have only bloom around the object which is fine to me. No more lens flares mess. Also, editor grid doesn't affect this. Graphics settings and PostFx settings are a good way to tweak any visual features per-game or per-scene.

![image](https://github.com/FlaxEngine/FlaxEngine/assets/3464853/1c054bc3-7cf6-424e-b3bd-312b966d3a03)

We won't match Unity look of empty scene 1:1 because we want to look better than this potato. That's why Flax uses various Post Fx visuals by default.

PR #2023, opened 2023-12-02
Disable eyeadaptation by default

Withaust:
Fixes https://github.com/FlaxEngine/FlaxEngine/issues/1859


mafiesto4:
This fixes that bug but breaks existing projects. I think we need to improve Eye Adaptation itself or settings for it to prevent that glow.

mafiesto4:
Closing in favor of #2030 and 1a68a52611c3c0e64111b1549e1833e34f151967

PR #2029, opened 2023-12-02
Refactor joints to use auto serialization, excluding D6Joint.

Tryibion:
This refactors joints to use auto serialize. This excludes the D6Joint as that has some special serialization going on. Also fixes missing editor group for Flags in Hinge joint.

### DIFF
--- a/Source/Engine/Physics/Joints/DistanceJoint.cpp
+++ b/Source/Engine/Physics/Joints/DistanceJoint.cpp
@@ -2,7 +2,6 @@
 
 #include "DistanceJoint.h"
 #include "Engine/Physics/PhysicsBackend.h"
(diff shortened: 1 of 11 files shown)


mafiesto4:
Yeah we need to use auto serialization but unfortunately, I made a mistake by merging joint properties into the members of join stream instead of using sub-objects which breaks the deserialization with this PR change.

I'm not sure what should be the best move now. We could refactor serializer to use nested serialization for joint property structures and still load old members from json data with `deprecated` mark to be updated in future - for sake of loading old projects properly. Or we could leave it for now as it is.

Tryibion:
With that in mind it is probably fine to leave it as it is right now. It may be a good one for a version where major breaking changes are allowed.

PR #2030, opened 2023-12-02
Changes to default post process settings.

Tryibion:
I believe these changes to post process settings make the default scene look a little better. I am not sure if the default project will have to have the graphics settings updated to match.


mafiesto4:
I think those changes make sense and help to bring more visual quality by default.

I would only skip changing default values for:
* `VignetteIntensity`+`VignetteColor` - vignette makes the image slightly more cinematic by default and can be easily disabled in Graphics Settings for specific games.

PR #2042, opened 2023-12-08
Fix blending additive animations.

### LINKED ISSUE #1981: Additive animation blending is totally broken.
**Issue description:**
Additive animation should allow us to combine animations cleanly, but atm they are completly broken. 

to test this, i threw two tpose animations with no transforms on any bones, and got this result 

https://github.com/FlaxEngine/FlaxEngine/assets/2074204/8ef49fe4-97f0-4316-8dcf-5cbc10d0dde1

Here is a test case file.
[Additive_pose_broken.zip](https://github.com/FlaxEngine/FlaxEngine/files/13447154/Additive_pose_broken.zip)

There is also a wave and an idle that have the same result, which should cleanly add with each other or with the tpose. 

Reference this video for what should happen. 
https://www.youtube.com/watch?v=TvoASoR7WiY

**Flax version:**
Master

Tryibion:
Sort of Fixes #1981 

This does an okay job. it does have some issues that I am not 100% sure how to mitigate. Like arm rotations and such which seem to be wonky sometime. I have kept the quaternion math in the PR, but it has some weird rotation issues that are fixed by using Matrix math.

### DIFF
--- a/Source/Engine/Animations/Graph/AnimGroup.Animation.cpp
+++ b/Source/Engine/Animations/Graph/AnimGroup.Animation.cpp
@@ -1075,11 +1075,16 @@ void AnimGraphExecutor::ProcessGroupAnimation(Box* boxBase, Node* nodeBase, Valu
                 {
                     tA = nodesA->Nodes[i];
                     tB = nodesB->Nodes[i];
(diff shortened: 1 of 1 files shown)


Tryibion:
This is actually just blending the anim and not doing it additively.

PR #2043, opened 2023-12-08
Add infinite loop to behavior tree 'Loop' decorator.

Menotdan:


NatreexTV:
Maybe not an infinite one, but loop based on bool. While thid condition is, continue, even if the sequence fail or sucess.

Menotdan:
> Maybe not an infinite one, but loop based on bool. While thid condition is, continue, even if the sequence fail or sucess.

You can do this already with knowledge coniditional paired with the loop. Just put a knowledge conidtional decorator, set your condition, put the loop decorator and set to infinite, and then it will end when the condition is false.

Issue #1696, opened 2023-10-12
Editor: Vertex Snapping


madsfs:
As has been discussed here on Discord: https://discord.com/channels/437989205315158016/1153036686356402308/threads/1153248925894066237 it would be extremely beneficial to have Vertex Snapping in the editor for rapid Level Design and Creation purposes. Unity has a really simple solution, at least from a UX standpoint, whereby you can move one object in a scene to another by chosing a vertex on one and dragging it to the vertex of another.

Here is how it is described on their [website](https://docs.unity3d.com/Manual/PositioningGameObjects.html#:~:text=Press%20and%20hold%20the%20V,other%20vertex%20on%20another%20Mesh.):

> **Vertex snapping**
> Use vertex snapping to quickly assemble your Scenes: take any vertex from a given Mesh and place that vertex in the same position as any vertex from any other Mesh you choose. For example, use vertex snapping to align road sections precisely in a racing game, or to position power-up items at the vertices of a Mesh.
> Follow the steps below to use vertex snapping:
> 1. Select the Mesh you want to manipulate and make sure the Move tool is active.
> 2. Press and hold the V key to activate the vertex snapping mode.
> 3. Move your cursor over the vertex on your Mesh that you want to use as the pivot point.
> 4. Hold down the left mouse button once your cursor is over the vertex you want and drag your Mesh next to any other vertex on another Mesh.
> 5. To snap a vertex to a surface on another Mesh, add and hold down the Shift+Ctrl (Windows) or Shift+Command (macOS) key while you move over the surface you want to snap to.
> 6. To snap the pivot to a vertex on another Mesh, add and hold the Ctrl (Windows) or Command (macOS) key while you move the cursor to the vertex you want to snap to.
> 7. Release the mouse button and the V key when you are happy with the results (Shift+V acts as a toggle of this functionality).

There is no reason to reinvent the wheel here. This works and is *very* convenient. It is an essential editor feature (as can be seen in the Discord thread it is highly desirable and agreed should be in there, to the point of being on the road-map).
See the attached Gif for an example of how it looks in Unity.
![Vertex+Snap 1](https://github.com/FlaxEngine/FlaxEngine/assets/113017777/7e97daa7-b931-4476-957e-3c601191ab12)

NoriteSC:
done in 
- https://github.com/FlaxEngine/FlaxEngine/pull/2045

mafiesto4:
Merged #2045.

PR #2045, opened 2023-12-08
vertex snapping

NoriteSC:
alt select vert
alt + right click to deselect the vert

https://github.com/FlaxEngine/FlaxEngine/assets/53096989/65941bbe-601d-4a76-803e-d2ef0056d3a1

# **Done**
- https://github.com/FlaxEngine/FlaxEngine/issues/1696


NoriteSC:
> Cool! A couple of thoughts. Would it bet better instead of using `SceneGraphNode` to create `ModelNode` and add it to the `StaticModel` and `AnimatedModel` types in the `SceneGraphFactory`, this way the code is not applying to all scene graph nodes?

btw the  `AnimatedModel`  is unsuported by vertex snaping
and making the `Model Node` do not makes sense

>  that it could be moved to the debug draw in the new `ModelNode` 
why ? there is close to 0 debug code unless u count the 3 draw class

PR #2051, opened 2023-12-08
Add ability to change default Android screen orientation.

Tryibion:
This is a feature for android where the user can change the default orientation of their application. This is nice because it allows for the user to lock it in portrait or landscape mode without the sensor changing it.

My build for android fails to build because it fails to copy .Net runtime data files. Once that is fixed then I will test this more. This is not a big change and should just work though...

### DIFF
--- a/Source/Editor/Cooker/Platform/Android/AndroidPlatformTools.cpp
+++ b/Source/Editor/Cooker/Platform/Android/AndroidPlatformTools.cpp
@@ -169,6 +169,30 @@ bool AndroidPlatformTools::OnPostProcess(CookingData& data)
         permissions += String::Format(TEXT("\n    <uses-permission android:name=\"{0}\" />"), e.Item);
     }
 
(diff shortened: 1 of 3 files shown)


mafiesto4:
Could you move this enum `AndroidScreenOrientation` to the inside of `AndroidPlatformSettings` class and rename it eg. to `ScreenOrientations`? This would match how a similar feature was implemented in `iOSPlatformSettings::UIInterfaceOrientations` or `UWPPlatformSettings::DisplayOrientations`.

Tryibion:
> Could you move this enum `AndroidScreenOrientation` to the inside of `AndroidPlatformSettings` class and rename it eg. to `ScreenOrientations`? This would match how a similar feature was implemented in `iOSPlatformSettings::UIInterfaceOrientations` or `UWPPlatformSettings::DisplayOrientations`.

Yeah that is good. I have made the change.

PR #2055, opened 2023-12-09
Add the option to deselect all.

Menotdan:
This PR adds the option to deselect all using (`Ctrl+Shift+A`, rebindable) wherever there is the option to select all (`Ctrl+A`, also rebindable). I also refactored most of the UI code to use the options in `InputOptions` for things like Copy, Paste, Cut, etc.

### DIFF
--- a/Source/Editor/Options/InputOptions.cs
+++ b/Source/Editor/Options/InputOptions.cs
@@ -56,6 +56,10 @@ namespace FlaxEditor.Options
         [EditorDisplay("Common"), EditorOrder(190)]
         public InputBinding SelectAll = new InputBinding(KeyboardKeys.A, KeyboardKeys.Control);
 
+        [DefaultValue(typeof(InputBinding), "Ctrl+Shift+A")]
+        [EditorDisplay("Common"), EditorOrder(195)]
+        public InputBinding DeselectAll = new InputBinding(KeyboardKeys.A, KeyboardKeys.Shift, KeyboardKeys.Control);
+
         [DefaultValue(typeof(InputBinding), "F")]
         [EditorDisplay("Common"), EditorOrder(200)]
         public InputBinding FocusSelection = new InputBinding(KeyboardKeys.F);
--- a/Source/Engine/UI/GUI/Common/TextBoxBase.cs
+++ b/Source/Engine/UI/GUI/Common/TextBoxBase.cs
@@ -1,7 +1,9 @@
 // Copyright (c) 2012-2023 Wojciech Figat. All rights reserved.
 
 using System;
+#if FLAX_EDITOR
 using FlaxEditor.Options;
+#endif
 using FlaxEngine.Assertions;
 using FlaxEngine.Utilities;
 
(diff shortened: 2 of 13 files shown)


Source/Engine/UI/GUI/Common/TextBoxBase.cs
```diff
@@ -1296,6 +1299,42 @@ public override bool OnKeyDown(KeyboardKeys key)
             bool ctrDown = window.GetKey(KeyboardKeys.Control);
             KeyDown?.Invoke(key);
 
+            // Handle controls that have bindings
+#if FLAX_EDITOR
+            InputOptions options = FlaxEditor.Editor.Instance.Options.Options.Input;
+            if (options.Copy.Process(this))
```
mafiesto4, line 1305: I'm not sure if we want to modify logic of gameplay component in Editor to have those shortcuts.

Source/Editor/Options/InputOptions.cs
```diff
@@ -56,6 +56,10 @@ public sealed class InputOptions
         [EditorDisplay("Common"), EditorOrder(190)]
         public InputBinding SelectAll = new InputBinding(KeyboardKeys.A, KeyboardKeys.Control);
 
+        [DefaultValue(typeof(InputBinding), "Ctrl+Shift+A")]
```
mafiesto4, line 59: What about `Escape`?

Source/Engine/UI/GUI/Common/TextBoxBase.cs
Menotdan: > I'm not sure if we want to modify logic of gameplay component in Editor to have those shortcuts.

So just drop the deselect all keybind from the gameplay component? or am I misunderstanding?

Source/Editor/Options/InputOptions.cs
Menotdan: There were a couple of discussions about this in discord. You can review them yourself and choose what you see fit.
image
image
image
image

Sorry for having to post it as images here, but the discussion is very very fragmented lol

Source/Engine/UI/GUI/Common/TextBoxBase.cs
mafiesto4: I mean that using Editor shortcuts in this `TextBoxBase` will work in editor but won't work in game so devs can treat it as an error.

Source/Editor/Options/InputOptions.cs
mafiesto4: okay makes sense, then let's leave it as it is with option to rebind by user 😄

Source/Engine/UI/GUI/Common/TextBoxBase.cs
Menotdan: Oh, actually, it isn't there. I was misunderstanding what you are saying.
I modified the editor things just because I wanted to be able to customize keybinds for editor text boxes. Is there a way to get the scope of the control? As in if its part of the editor or part of the game inside the editor?

mafiesto4: We could leave this option for editor keybinds, just need t be aware that editor shortcuts won't work in cooked game. But ofc `TextBoxBase` implements typical `Ctrl+C`/`Ctrl+V`/etc. shortcuts as a base functionality.

Issue #1985, opened 2023-11-24
Add `BehaviorTreeKnowledgeBooleanDecorator` for conditionals based on bool/pointer value


mafiesto4:
When using `Actor` (or other object) references in Behavior Knowledge it would be nice to have an easy way to check if value is not null. Similarly, it could check if a boolean value is set (eg. `bool HasTarget`). Optionally allow inverting condition. Add automatic title based on condition if no custom name is provided.

NatreexTV:
Maybe a while is not or is true or false repeat this could be interesting too.

mafiesto4:
Added in d5f4254a7314a9ef9ac930678d1faa2c04e4983c.

PR #2056, opened 2023-12-09
Enhanced the Knowledge conditional decoration node.

NatreexTV:
So as my first edit ever on GitHub and on any project, I decide to enhanced the AI a little bit because I use it a lot and I needed some more improvement on one node.

image

In this new version, you need to choose between tree type of value. The boolean, the Int or a Float. 

You can now compare any boolean from Value A to the boolean from the setted value.

It was possible with the Knowledge Value conditional decoration node, but if you just want to do it without any other additionnal variable in the Blackboard, that needed to be added.

I comment every field for help anyone with it.


mafiesto4:
For boolean values we have now this: #1985

Issue #2062, opened 2023-12-10
Actor bounding boxes slightly wrong with rotation.


Menotdan:
**Issue description:**
When you rotate an actor, it's bounding box is calculated with extra space that is not needed.

**Minimal reproduction project:**
[ActorBounds.zip](https://github.com/FlaxEngine/FlaxEngine/files/13626075/ActorBounds.zip)

Steps to Reproduce:
1. Open the project, and enter playmode.
2. Ctrl + Shift + Tab to return to the Editor viewport in play mode.
3. Investigate the bounding boxes of the actors.

![image](https://github.com/FlaxEngine/FlaxEngine/assets/32620310/8885c46d-a5f5-4183-8324-0cc7ab0719bd)
(Hard to get a good screenshot)

**Flax version:**
Launcher master branch as 12/09/2023.

mafiesto4:
That's simply how `BoundingBox` works - it represents an axis-aligned bounding box in three-dimensional space.
We have `OrientedBoundingBox` which can represent bounds with rotation but actors have to use an optimized version due to performance (`OBB` uses `Vector3` + `Transform` while `ABB` is 2x`Vector3`).

Menotdan:
> That's simply how `BoundingBox` works - it represents an axis-aligned bounding box in three-dimensional space.
> We have `OrientedBoundingBox` which can represent bounds with rotation but actors have to use an optimized version due to performance (`OBB` uses `Vector3` + `Transform` while `ABB` is 2x`Vector3`).

I know this, but the box is still bigger than the actor's minimum and maximum points after rotation. Maybe that's too expensive to be recalculated after rotating?

Menotdan:
Hm, there also seems to be some odd behavior calculating the bounds for empty actors:
![image](https://github.com/FlaxEngine/FlaxEngine/assets/32620310/21c6e6a9-7a2b-46d5-95ba-3b26759b9d30)
But maybe outside the scope of this issue?

PR #2063, opened 2023-12-10
Implement auto-sizing for box colliders when they are added to the scene.

Menotdan:
This uses the parent model's bounding box, which is currently slightly wrong (#2062), but its better than having to set up your collider from scratch.

### DIFF
--- a/Source/Engine/Level/Actor.cs
+++ b/Source/Engine/Level/Actor.cs
@@ -117,6 +117,15 @@ namespace FlaxEngine
         }
 
         /// <summary>
+        /// Called in-editor when an actor is added to the scene.
+        /// If not in the editor, this function will not be called.
+        /// </summary>
+        public virtual void OnActorSpawned()
+        {
+
+        }
+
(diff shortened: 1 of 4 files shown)


Source/Engine/Level/Actor.cs
```diff
@@ -116,6 +116,15 @@ public void ResetLocalTransform()
             LocalTransform = Transform.Identity;
         }
 
+        /// <summary>
+        /// Called in-editor when an actor is added to the scene.
+        /// If not in the editor, this function will not be called.
+        /// </summary>
+        public virtual void OnActorSpawned()
```
mafiesto4, line 123: We have `ActorNode` system which handles editor-interactions with actors and scene objects.

Menotdan: Alright, I removed `OnActorSpawned` and used BoxColliderNode for this. I did keep the new `BoxCollider.cs` however, as I think that's the best place for the custom editor which adds to manual resizing option.

PR #2079, opened 2023-12-14
implement MoveFileToRecycleBin on Linux

nothingTVatYT:
This adds the Linux equivalent of MoveFileToRecycleBin following the freedesktop.org specs and changes the deletion code in Content.cpp to use it.


nothingTVatYT:
Unfortunately curl is only available in the editor, so this is now using an own encoder.

PR #2083, opened 2023-12-15
fix check in operator for Tag

nothingTVatYT:
operator != used the wrong check


mafiesto4:
Ups sorry, go this committed some time ago just forgot to push

PR #2094, opened 2023-12-16
Change color grading post process settings to use color picker. 

Tryibion:
This PR changes the color grading post process settings to use the color picker. This is a QoL change to allow for different ways of setting the color. This PR also fixes the color editor to work correctly with Float4.

 I can also get rid of the color track ball as I believe that it was only being used in the post process settings... But thought I would leave it in for now unless we should remove it.


mafiesto4:
What about extending trackball color editor to support color picker opening eg. on `right-click`?

Tryibion:
Closing in favor of #2106 after your comments.

PR #2104, opened 2023-12-20
Fix for resetting local transform of newly created prefab child

### LINKED ISSUE #2103: Local transform of an actor created in a prefab does not equal to Identity
**Issue description:**
When you creeate an actor in a prefab it's position is not gonna be zero. Which is the expected behaviour.
 

**Steps to reproduce:**

https://github.com/FlaxEngine/FlaxEngine/assets/47716519/2c385e77-9f85-4030-810e-acc8b41d9609

**Flax version:**
1.7.1

abrasivetroop:
Fixes #2103

MY FIRST EVER PULL REQUEST 😭

If there is a more elegant way to fix it or I did something wrong please let me know!

### DIFF
--- a/Source/Editor/Windows/Assets/PrefabWindow.Hierarchy.cs
+++ b/Source/Editor/Windows/Assets/PrefabWindow.Hierarchy.cs
@@ -363,10 +363,9 @@ namespace FlaxEditor.Windows.Assets
             if (parentActor != null)
             {
                 // Match the parent
-                actor.Transform = parentActor.Transform;
+                actor.LocalTransform = Transform.Identity;
                 actor.StaticFlags = parentActor.StaticFlags;
                 actor.Layer = parentActor.Layer;
-                actor.LocalTransform = Transform.Identity;
 
                 // Rename actor to identify it easily
                 actor.Name = Utilities.Utils.IncrementNameNumber(actor.Name, x => parentActor.GetChild(x) == null);
--- a/Source/Editor/Windows/Assets/PrefabWindow.Hierarchy.cs
+++ b/Source/Editor/Windows/Assets/PrefabWindow.Hierarchy.cs
@@ -366,9 +366,7 @@ namespace FlaxEditor.Windows.Assets
                 actor.Transform = parentActor.Transform;
                 actor.StaticFlags = parentActor.StaticFlags;
                 actor.Layer = parentActor.Layer;
-                actor.LocalPosition = Vector3.Zero;
-                actor.LocalEulerAngles = Vector3.Zero;
-                actor.LocalScale = Vector3.One;
+                actor.LocalTransform = Transform.Identity;
 
                 // Rename actor to identify it easily
                 actor.Name = Utilities.Utils.IncrementNameNumber(actor.Name, x => parentActor.GetChild(x) == null);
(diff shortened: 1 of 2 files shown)


abrasivetroop:
I moved the fix into PrefabWindow.Hierarchy.cs

Source/Editor/Windows/Assets/PrefabWindow.Hierarchy.cs
```diff
@@ -366,6 +366,9 @@ public void Spawn(Actor actor)
                 actor.Transform = parentActor.Transform;
                 actor.StaticFlags = parentActor.StaticFlags;
                 actor.Layer = parentActor.Layer;
+                _actor.LocalPosition = Vector3.Zero;
+                _actor.LocalEulerAngles = Vector3.Zero;
+                _actor.LocalScale = Vector3.One;
```
Tryibion, line 371: `_actor` does not exist here. probably need to change this to `actor`

Source/Editor/Windows/Assets/PrefabWindow.Hierarchy.cs
```diff
@@ -366,6 +366,9 @@ public void Spawn(Actor actor)
                 actor.Transform = parentActor.Transform;
```
Tryibion, line 366: You could simplify this code by just changing this line to `actor.LocalTransform = Transform.Identity;` and removing the 3 times you wrote. No need to set the transform 2 times.

Tryibion: You could replace this line of code with that identity code, or since you added it below, you can remove this line, no need to set the transform 2 times.

abrasivetroop:
Thanks Tryibion!
I guess that should do it right?

Source/Editor/Windows/Assets/PrefabWindow.Hierarchy.cs
```diff
@@ -366,6 +366,7 @@ public void Spawn(Actor actor)
...
                 actor.StaticFlags = parentActor.StaticFlags;
                 actor.Layer = parentActor.Layer;
+                actor.LocalTransform = Transform.Identity;
```
Tryibion, line 369: Because you added this here, you can remove the transform setting above because there is no need to set the transform 2 times. Or you can just use this line of code to replace the other line where the transform is being set (line 366).

Tryibion: Makes more sense to me to remove line 366 and replace it with what you wrote in line 369.

abrasivetroop: But doesn't that line move the actor to where it's parent is?

Tryibion: No, since you are setting the local transform, as long as the actor has a parent set, it will spawn it correctly.

abrasivetroop: Okay I tried your suggestion and you are right!

Tryibion: You can test it and see. It seemed to work for me when I tested it.

abrasivetroop: Yes I made the change

mafiesto4 (approved):

Issue #1957, opened 2023-11-18
[Suggestion] Fullscreen for Game View


Myterian:
I think it's helpful to preview the game in fullscreen in the editor, since this can help adjusting cameras position, UI size and scaling and some post processing effects like vignette. 

Also, a game in fullscreen comes across / feels different than in maximized window and that has to be taken into account during developent as well.

MineBill:
When in play mode, try pressing F11. That should maximize the Game window.

Myterian:
yeah, I'm almost always using that. But I'm talking about proper fullscreen, so you can play the game like a player would do. Actually, borderless window would already be enough

mafiesto4:
Implemented in #2116

Issue #1738, opened 2023-10-18
[Task] fix bad UX design


NoriteSC:
# **Problems** 
the "new" addition of cook and run placement is not the best.
super secret setting (right click on game window tab)
![image](https://github.com/FlaxEngine/FlaxEngine/assets/53096989/0e214e29-6aa7-44e4-b8de-f0dc8b094ff2)
# **Fixes**
moving super secret setting
and cook and run to
a drop down
on the side of play button 

**Refrence** (the standalone mode in unreal is  "cook and run" in flax)
![image](https://github.com/FlaxEngine/FlaxEngine/assets/53096989/f998c5b2-a2fc-4195-8cfb-cb4b6c3c2ee7)

# **Expadbility potencial**
this engine supports the multiplayer it will allow to add the settings in more intuitive place at some point
to play as
  1. client
  2. host
  3. server

other
- play with specyfic resolucion as a presets 
- symulate the scene if u dont acualy need to play with player but u are testing AI for example 
- spawning a instance of the editor for multyplayer

and more

ps. as a player actor will be at some point in this engine because of features will have need a player location

# **Feel free to pick this, happy coding**
* difficulty easy

PR #2116, opened 2023-12-21
Add the ability to quickly change window modes for the game window.

MineBill:
This _kinda_ implements #1957 (explained below).
A new interface option is exposed which allows to quickly change how the "Game" window is displayed upon starting the simulation inside the editor by adding the following choices:
* Docked
  This is the current behavior. It leaves the game window as is and doesn't touch it.
* Popup Window
  This always makes the window popup even if it is docked.
* Maximized Window
  This maximizes the window, same behavior as pressing the F11 key.
* Borderless Window
  This makes the game window borderless, without a titlebar, while also drawing over Windows taskbar.

https://github.com/FlaxEngine/FlaxEngine/assets/30367251/cf43c38d-edc3-43c4-8dcf-18335af9eec8

As you can see in the preview, the Flax drawn titlebar is still visible in borderless, by design. In my initial implementation i also completely removed this too, making it true borderless fullscreen but my concern is how a user would react in the case they forgot to add some functionality to terminate the game (like the ExitOnEsc script). It would probably lead to frustration since there would be no visible way to close the window without pressing Alt-F4, opening the Task Manager or doing something else to kill the window. If you have any more thoughts on this let me know.

@nothingTVatYT and @RuanLucasGD i would appreciate it if you could test this on Linux.

### DIFF
--- a/Source/Editor/Modules/SimulationModule.cs
+++ b/Source/Editor/Modules/SimulationModule.cs
@@ -264,10 +264,13 @@ namespace FlaxEditor.Modules
                 _enterPlayFocusedWindow = gameWin;
 
             // Show Game widow if hidden
-            if (gameWin != null && gameWin.FocusOnPlay)
+            if (gameWin != null)
             {
-                gameWin.FocusGameViewport();
+                if (gameWin.FocusOnPlay)
+                    gameWin.FocusGameViewport();
+                gameWin.SetWindowMode(Editor.Options.Options.Interface.DefaultGameWindowMode);
             }
+
 
             Editor.Log("[PlayMode] Enter");
         }
--- a/Source/Editor/Modules/UIModule.cs
+++ b/Source/Editor/Modules/UIModule.cs
@@ -737,6 +737,17 @@ namespace FlaxEditor.Modules
             playActionGroup.SelectedChanged = SetPlayAction;
             Editor.Options.OptionsChanged += options => { playActionGroup.Selected = options.Interface.PlayButtonAction; };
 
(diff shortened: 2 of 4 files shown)


RuanLucasGD:
Popup:
image

Maximized:
image

I think maximized don't work because "maximize" action of Linux window system isn't implemented yet.
https://github.com/FlaxEngine/FlaxEngine/blob/72f45afa4505e2c4a5df521dbe3b74edf45612ad/Source/Engine/Platform/Linux/LinuxWindow.cpp#L121C39-L121C39

Borderless:
image

NoriteSC:
u made a part of
- https://github.com/FlaxEngine/FlaxEngine/issues/1738 

im happy 👍

MineBill:
Aha, that reminded me. The placement of this menu by right clicking the play button is indeed a bad one. I've been using Flax for months and only found out about it yesterday. I wanted to create an additional options button for these settings (possibly for Cook & Run too) and place it near the Play button but i wanted mafi's approval first. I forgot to mention this on the PR so i added it here.

nothingTVatYT:
Unsurprisingly I get the same result as @RuanLucasGD and yes, it's probably because of the poor implementation of window modes and types on Linux. I'm going to have a look at how we can get fullscreen, maximized and borderless but working just with the plain libX11 feels like programming a tetris clone is assembler.

nothingTVatYT:
[patch0.txt](https://github.com/FlaxEngine/FlaxEngine/files/13746388/patch0.txt)

With this patch the maximized window seems to work but I have problems with the mouse buttons. It could be that all my test projects are broken in some way as I played with plugins recently and experienced lots of crashes.
That means it needs more testing but I think the `true` flag does something we miss currently.

PR #2149, opened 2024-01-02
Add custom editor for buttons that allow listening for them inside the editor.

MineBill:
This adds support for listening and capturing KeyboardKeys, MouseButtons and GamepadButtons.
It makes setting up Action and Axis mappings a bit easier.

https://github.com/FlaxEngine/FlaxEngine/assets/30367251/1b7acfaf-4677-45c4-b7e0-d48b255f3c25

In my initial implementation i made a custom editor for `ActionConfig` but having separate editor for each type of button seems more intuitive, could allow for more specialization in the future and also works anywhere, including user Scripts.


mafiesto4:
Cool! In future, we could change that button into an icon.

PR #2157, opened 2024-01-04
Fix negate node not converting bool to the appropriate float number.

### LINKED ISSUE #2156: Negate node does not work in an AnimationGraph
**Issue description:**
The _Negate_ node can be added to the graph, but it doesn't seem to function at all, but instead always seems to return true.

image

**Flax version:**
1.72 (master as of the time of issue submission)

MineBill:
Fixes #2156.

This makes sure that a false boolean is turned into -1.0f so that when the negate operation is run, it will turn to 1(true). Otherwise, a false boolean would be converted to 0 and stay 0 after the negate operation.

### DIFF
--- a/Source/Engine/Visject/GraphUtilities.cpp
+++ b/Source/Engine/Visject/GraphUtilities.cpp
@@ -297,7 +297,7 @@ void GraphUtilities::ApplySomeMathHere(Variant& v, Variant& a, MathOp1 op)
     switch (a.Type.Type)
     {
     case VariantType::Bool:
-        v.AsBool = op(a.AsBool ? 1.0f : 0.0f) > ZeroTolerance;
+        v.AsBool = op(a.AsBool ? 1.0f : -1.0f) > ZeroTolerance;
         break;
     case VariantType::Int:
         v.AsInt = (int32)op((float)a.AsInt);


mafiesto4:
I'm not sure if it's the best way to fix this. Have you tested other math nodes like `Abs` if they work fine for this? Like `abs(false)=false` and `abs(true)=true`?

MineBill:
You are right, this will break other math nodes. But it got me thinking: Are there well defined math operations on bools? Boolean already have their own nodes but what exactly "should" happen when we try to apply them to math? As demonstrated here, assuming true == 1 and false == 0, doesn't always work but neither does true == 1 and false == -1. Is maybe a better idea to not allow booleans to be used with math nodes?

envision3d:
I honestly wasn't even aware that this node was supposed to return a number as the red result port suggested to me that it would be a bool.  Is there a problem there? Or was that a false assumption on my part?

MineBill:
> I honestly wasn't even aware that this node was supposed to return a number as the red result port suggested to me that it would be a bool. Is there a problem there? Or was that a false assumption on my part?

Negate is a math node and Flax will treat booleans as floats by turning true to 1 and false to 0 before passing them to any math node. In some cases it will work as "expected", in some case not. The problem is that not all math operations will produce the expected boolean result. The bigger problem is that there is no "formal" definition of what all math nodes should do on boolean inputs. What should `atan2(false)` return? It's more a problem of expectations. 

What you experienced on your issue is not a technical bug. Knowing that Flax turns boolean to floats before performing the negation (and then turning them back to booleans), the result makes sense.

nothingTVatYT:
> Is it maybe a better idea to not allow booleans to be used with math nodes?

I think that would be the most clean way. Most math functions simply don't make sense on a boolean unless you know how Flax handles them internally. That knowledge shouldn't be required though. Somebody with less insight in Flax but a basic math background wouldn't and shouldn't expect that e.g. `abs(true)` or `atan2(false)` is anything but undefined.

Issue #817, opened 2022-11-04
Using Smooth Mode/Brush causes chunks to split


CthulhuBaba:
**Issue description:**
When I use the Smooth Mode in the ToolBox to edit the terrain, the chunks are getting splitted/disconnected. I showed all the steps in the video. https://youtu.be/GB-9TSAWNPw

**Flax version:**

1.4.6334
![smoothbrush](https://user-images.githubusercontent.com/31192693/200015211-2d1ca905-0747-4a9d-9d55-69f52db94202.jpg)

AlTheSlacker:
I can see why this happens. In \FlaxEngine\Source\Editor\Tools\Terrain\Sculpt\SmoothMode.cs a bug is introduced where the code attempts to get the average height of the surrounding points, there are some Math.Max functions to prevent the heightmap request going out of range for a given patch, but this results in the values used to calculate the average height for the brush becoming dependent upon which patch is being used: When you brush near the patch boundaries you get inconsistent values. I couldn't figure a quick fix as you need to be able to sample from adjacent patches, which for me at least would significantly change this code. Smoothing is still broken for any terrain more than 1x1 patch in size as a result of this.

AlTheSlacker:
Thanks for the fix. I see in your comments you talk about viewing the terrain height data as a single set of data: If it is of any use to you, this was my approach in https://github.com/AlTheSlacker/FlaxProcTerrain where I dump the entire terrain into a float array. In-editor performance seems good enough, if you want to take any code, help yourself.

PR #2169, opened 2024-01-09
WIP: Fixes terrain smoothing separation issue

envision3d:
Addresses issues in #817 .

### DIFF
--- a/Source/Editor/Tools/Terrain/Sculpt/SmoothMode.cs
+++ b/Source/Editor/Tools/Terrain/Sculpt/SmoothMode.cs
@@ -1,6 +1,7 @@
 // Copyright (c) 2012-2023 Wojciech Figat. All rights reserved.
 
 using System;
+using System.Security.Cryptography.X509Certificates;
 using FlaxEngine;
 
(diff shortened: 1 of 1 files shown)


Source/Editor/Tools/Terrain/Sculpt/SmoothMode.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) 2012-2023 Wojciech Figat. All rights reserved.
 
 using System;
+using System.Security.Cryptography.X509Certificates;
```
Tryibion, line 4: Is this include needed?

envision3d: Ah good catch, thanks.  This probably got included from a bad autocomplete along the way.

I discovered there are still some problems with my fix and am working to solve those; I'll remove this in the next commit.

envision3d:
I discovered there's a problem upstream that will prevent this approach from ever fully working.  The change required upstream will necessitate a different set of changes in this file as well, so I'm going to close this and reopen once the new approach is completed.

PR #2170, opened 2024-01-10
Keep local orientation on collider spawn. 

Tryibion:
This PR keeps the local orientation of the collider when spawned and add option to keep orientation when resizing collider. I found it frustrating when I had a rotated model and added a collider to it and it did not keep the local rotation as I has to rotate again to match the rotation of the model. ~~The issue I have now is that the box is larger because the model is rotated than if the collider is added with the model not rotated.~~


Menotdan:
> The issue I have now is that the box is larger because the model is rotated than if the collider is added with the model not rotated.

This is due to my previous PR only calculating the size of the box that it should be resized to using the AABB that flax has for every actor. I think its calculation is some ways off. See https://github.com/FlaxEngine/FlaxEngine/issues/2062.

Tryibion:
I have a hacky way right now that I may pr, where it just rotates the parent to its identity, grabs it's box, and then rotates it back...

PR #2186, opened 2024-01-17
fix plugin entry layout

### LINKED ISSUE #2185: Plugins Window misses details due to the layout
**Issue description:**
The plugin window is supposed to show the version and whether it's alpha or beta. The version is not visible at all and the alpha label is culled half.

This is independent on the size of the plugin window and most probably a layout problem.

Screenshot_Plugins_2

**Flax version:**
1.7.2

nothingTVatYT:
Fixes #2185


nothingTVatYT:
Before:
Screenshot_Plugins_1

After applying the fix:
Screenshot_Plugins_4

PR #2192, opened 2024-01-21
Vehicle Physics Improviments

RuanLucasGD:
Hello, i made some changes on vehicle system to make more flexible and easy to use.

# Drive Control

image

Able to configure the vehicle control behavior, very useful to improve the maneuvering on curves for example, making the steer more smooth or change throttle sensitive. 

## Steer vs Speed

image

Now is possible set the relationship between steer and velocity, its very good to improve the vehicle control on high speed decreasing the steer angle.

https://github.com/FlaxEngine/FlaxEngine/assets/79365912/dceee1fc-556d-410b-bc8e-b93bbd7e53b1

# Tank Drive Type

image

Added Tank Drive Type to vehicles, tanks can move applying throttle on each side, left or side.
All wheels of a side move together depending of the throttle.

https://github.com/FlaxEngine/FlaxEngine/assets/79365912/58fe7efe-b6cb-4516-9d1d-6ec1989bcf69

[reference](https://nvidia-omniverse.github.io/PhysX/physx/5.1.0/docs/Vehicles.html#pxvehicledrivetank)

### Tank drive mode

image

Tanks can have different drives modes, like Special and Standard.

* Standard can throttle only one track by time, so if you need turn to left for example, the left track will be stoped and right track will move to forward.

* Special apply acceleration in all tracks, so if you need to turn to left for example, the left track will accelerate to backward and right track will accelerate to forward.

# Gears

image

Added Forward Gears Ratio to able configure gears count to move to forward.

# Wheels

image

Removed wheel type from wheels, the unique reason to exist it is sort wheels on start game because DriveW4, DriveNW and TankDrive needs to have all wheels sorted to know the correct wheel position, now the sort is made automatically based on wheel local position on vehicle: 

## Drive W4

| wheels index  | type |
| -- | ------------------------ |
| 0  | forward right      |
| 1  | forward left         |
| 2  | backward right   |
| 3  | backward left      |

## Drive NW

This not need to know the wheel index, but wheels are sorted because anti roll bars needs to know the wheel side.

## Tank Drive

Tank Drive needs to have sorted wheels to know the side to apply throttle correctly on each side.

| wheels index  | type |
| -- | ------------------------ |
| 0  | right                     | 
| 1  | left                        |
| 2  | right                     |
| 3  | left                        |

## Wheel Angle

Wheel steer angle has been changed to -180 180 because this helps to create negative steer. Negative steer is applied automatically on vehicles DriveW4 on back wheels but this not happen on DriveNW vehicles.

Negative wheel angle is very useful to create vehicles with various maneuverable axles, front and back, like large trucks.

https://github.com/FlaxEngine/FlaxEngine/assets/79365912/23f1345d-9ecf-4956-8956-3c6b8d547cb4

## Sprung Mass Multiplier

image

By default, the suspension spring force is based on vehicle mass distribution, for example, if you have 4 wheels and the vehicle have 1000 of mass, the sprung force is 250 for each wheel (only an simple example).

This force is based on center of mass of the vehicle, so if the center mass is on front, the front suspension will have more force than back suspension springs. 

Now is possible edit this force changing the sprung mass multiplier to make the spring more strong or more weak.

# Anti Roll Bars

image
Increase the stability on each vehicle axle (each axle is made using left and right wheel).

Anti roll apply a contrary force to avoid vehicle turn around your our axis, making more maneuvering in curves for example but me personally don't see much difference with it.

Anti roll bars needs to have wheels sorted to know the wheel side and apply the force correctly.

| wheels axle 0 | type |
| -- | ------------------------ |
| 0  | right                     | 
| 1  | left                        |
| wheels axle 1 | type |
| 2  | right                     |
| 3  | left                        |

more info about [anti roll bars](https://nvidia-omniverse.github.io/PhysX/physx/5.1.0/docs/Vehicles.html#pxvehicleantirollbar).

### DIFF
--- a/Source/Engine/Physics/Actors/WheeledVehicle.cpp
+++ b/Source/Engine/Physics/Actors/WheeledVehicle.cpp
@@ -14,7 +14,7 @@
 #include "Engine/Core/Log.h"
 #endif
 
-WheeledVehicle::WheeledVehicle(const SpawnParams& params)
+WheeledVehicle::WheeledVehicle(const SpawnParams &params)
     : RigidBody(params)
 {
     _useCCD = 1;
@@ -33,12 +33,22 @@ void WheeledVehicle::SetDriveType(DriveTypes value)
     Setup();
 }
 
-const Array<WheeledVehicle::Wheel>& WheeledVehicle::GetWheels() const
+void WheeledVehicle::SetDriveMode(DriveModes value)
+{
+    _driveMode = value;
+}
+
+WheeledVehicle::DriveModes WheeledVehicle::GetDriveMode() const
+{
+    return _driveMode;
+}
+
+const Array<WheeledVehicle::Wheel> &WheeledVehicle::GetWheels() const
 {
     return _wheels;
 }
 
-void WheeledVehicle::SetWheels(const Array<Wheel>& value)
+void WheeledVehicle::SetWheels(const Array<Wheel> &value)
 {
 #if WITH_VEHICLE
     // Don't recreate whole vehicle when some wheel properties are only changed (eg. suspension)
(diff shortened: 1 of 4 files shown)


RuanLucasGD:
Example project:

[VehiclesTemplate.zip](https://github.com/FlaxEngine/FlaxEngine/files/14001740/VehiclesTemplate.zip)

PR #2206, opened 2024-01-26
Fixing codegen skipping "if" and writing "else if" first

z1dev:
Structs, that have static and other fields before "normal" fields generated c++ code like this example:
```
    static void GetField(void* ptr, const String& name, Variant& value)
    {
        else if (name == TEXT("_tok"))
            value = Variant(((FudgetToken*)ptr)->_tok);
    }
```


NoriteSC:
i will prefer this way to fix it 
image

z1dev:
My fix follows the same pattern seen elsewhere in that same file. Both versions fixes it, I don't care which one is used.

PR #2208, opened 2024-01-26
Collection UI adjustments

Tryibion:
This is another version of #1720 but with most of the dictionary editor changes removed and the moving arrows removed.

image


ElevenArt:
That's great. When I research Flax, this was really lacking. I have a question. Would it be possible to make it easier to remove item from lists? For the current method, you need to click RMB over Element 0,1,2, etc. and select the delete option from the context menu. This is very confusing behavior and it took me 2 months to find this. Thanks :)

nothingTVatYT:
Has anyone used FreeCAD? They introduced a spreadsheet-like editor to keep values in a list form.
Although I don't like their implementation particularly because it is mostly disconnected from the task where you use the values, the general idea of a spreadsheet-like table has potential if it's an integrated part rather than a kind of external add-on like it feels like in FreeCAD.
Tables are something people are used to and it can optimize the space you need to organize values in a two-dimensional grid. I imagine some well-known and accepted UI features like selecting, inserting, deleting rows, copy and paste of values, fill cells, etc. and it could become a solution for quite some applications.

PR #2213, opened 2024-01-28
add units support in float, double and Float3 input

nothingTVatYT:
~~One remaining issue: The syntax check on Rider marks the attribute as an error although it compiles and works just fine.~~ Fixed.


mafiesto4:
Very cool stuff!

@nothingTVatYT Maybe we rename `NumberCategoryAttribute` to `ValueCategoryAttribute` to match `Utils.ValueCategory` enum?

nothingTVatYT:
> Very cool stuff!
> 
> @nothingTVatYT Maybe we rename `NumberCategoryAttribute` to `ValueCategoryAttribute` to match `Utils.ValueCategory` enum?

I intentionally renamed it because I had some conflicts because of the two ValueCategory types. I don't remember the exact error message but Rider did not like that so I had to come up with a different name.

PR #2241, opened 2024-02-14
added IAssetFactory::Get().ContainsKey cheak

NoriteSC:
gives more info what went wrong instead of empty call stack and crash
image

### DIFF
--- a/Source/Engine/Content/Factories/JsonAssetFactory.h
+++ b/Source/Engine/Content/Factories/JsonAssetFactory.h
@@ -49,7 +49,20 @@ protected:
 	class CONCAT_MACROS(Factory, type) : public JsonAssetFactory<type> \
 	{ \
 		public: \
-		CONCAT_MACROS(Factory, type)() { IAssetFactory::Get().Add(type::TypeName, this); } \
+		CONCAT_MACROS(Factory, type)() \
+        { \
+            if (IAssetFactory::Get().ContainsKey(type::TypeName)) \
+            { \
+                if (Platform::IsDebuggerPresent()) \
+                { \
+                    PLATFORM_DEBUG_BREAK; \
+                } \
+                Platform::Assert("[IAssetFactory] Can't Register "#type \
+                " it is already registered", __FILE__, __LINE__); \
+                return; \
+            } \
+            IAssetFactory::Get().Add(type::TypeName, this); \
+        } \
 		~CONCAT_MACROS(Factory, type)() { IAssetFactory::Get().Remove(type::TypeName); } \
 		bool SupportsVirtualAssets() const override { return supportsVirtualAssets; } \
 	}; \


Source/Engine/Content/Factories/IAssetFactory.h
```diff
@@ -67,3 +67,32 @@ class FLAXENGINE_API IAssetFactory
         return nullptr;
     }
 };
+
+#if NDEBUG
```
mafiesto4, line 71: `#if !BUILD_RELEASE`

Source/Engine/Content/Factories/IAssetFactory.h
```diff
@@ -67,3 +67,32 @@ class FLAXENGINE_API IAssetFactory
...
+#define INTERNAL_FACTORY_REGISTER_CHEAK(type)\
+if (IAssetFactory::Get().ContainsKey(type::TypeName)) \
```
mafiesto4, line 73: `CHECK(!IAssetFactory::Get().ContainsKey(type::TypeName))`

Source/Engine/Content/Factories/IAssetFactory.h
```diff
@@ -67,3 +67,32 @@ class FLAXENGINE_API IAssetFactory
...
+    if (Platform::IsDebuggerPresent()) \
+    { \
+    PLATFORM_DEBUG_BREAK; \
+    } \
+    Platform::Assert("[IAssetFactory] Can't register "#type \
+        " it is already registered", __FILE__, __LINE__); \
+        return; \
+}
+#define INTERNAL_FACTORY_UNREGISTER_CHEAK(type)\
+if (!IAssetFactory::Get().ContainsKey(type::TypeName)) \
```
mafiesto4, line 84: CHECK(IAssetFactory::Get().ContainsKey(type::TypeName))

mafiesto4 (changes requested):

Source/Engine/Content/Factories/IAssetFactory.h
NoriteSC: custom massage better suits this case 
giving better information was is wrong

Source/Engine/Content/Factories/IAssetFactory.h
NoriteSC: custom massage better suits this case 
giving better information was is wrong

Source/Engine/Content/Factories/IAssetFactory.h
NoriteSC: will do

Source/Engine/Content/Factories/IAssetFactory.h
mafiesto4: Then at least I would skip using `#type` to reduce unique strings in compiled binary (smaller engine size).

PR #2246, opened 2024-02-14
Rubberband-style selection in a ContentView

nothingTVatYT:
This adds a rubberband-style selection option to the content view. Shift key adds to the selection otherwise it's replaced.


MineBill:
Nice! You could draw the rectangle at the end of `Draw()` after the `base.Draw` call. This way, it will be rendered above the content items, which are children. Also, how about additionally adding a `FillRectangle` with an alpha multiplied color? That should make it look more like existing file managers.

nothingTVatYT:
Yeah, you're right. Now it looks like visject is doing it.

PR #2250, opened 2024-02-15
Fix NextUnitVector2

rkrahn:
Fixed NextUnitVector2 method not producing a vector within a unit circle.

### DIFF
--- a/Source/Engine/Utilities/Extensions.cs
+++ b/Source/Engine/Utilities/Extensions.cs
@@ -298,8 +298,8 @@ namespace FlaxEngine.Utilities
         public static Vector2 NextUnitVector2(this Random random, float radius = 1.0f)
         {
             float magnitude = (float)random.NextDouble() * radius;
-            double randomDegree = random.NextDouble() * 360;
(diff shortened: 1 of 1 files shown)


Source/Engine/Utilities/Extensions.cs
```diff
@@ -297,8 +297,9 @@ public static Quaternion NextQuaternion(this Random random, bool randomRoll = fa
         /// <returns>A random <see cref="Vector2"/>.</returns>
         public static Vector2 NextUnitVector2(this Random random, float radius = 1.0f)
         {
-            var randomRadius = (float)random.NextDouble() * radius;
-            return new Vector2((float)Math.Cos(random.NextDouble()) * randomRadius, (float)Math.Sin(random.NextDouble()) * randomRadius);
+            float magnitude = (float)random.NextDouble() * radius;
+            double randomDegree = random.NextDouble() * 360;
```
mafiesto4, line 301: I think this should be: ` double randomDegree = random.NextDouble() * Mathf.RevolutionsToRadians;` so cos/sin will get proper value in radians.

rkrahn: You're right. I assumed it was degrees since the result was still what was desired. Fixed in new commit.

PR #2251, opened 2024-02-15
Add NextUnitCircleVector2 extension method

rkrahn:
Adds the NextUnitCircleVector2 extension method which produces a Vector2 on a unit circle with a magnitude equal to the radius.

### DIFF
--- a/Source/Engine/Utilities/Extensions.cs
+++ b/Source/Engine/Utilities/Extensions.cs
@@ -309,8 +309,8 @@ namespace FlaxEngine.Utilities
         /// <returns>A random <see cref="Vector2"/>.</returns>
         public static Vector2 NextUnitCircleVector2(this Random random, float radius = 1.0f)
         {
-            double randomDegree = random.NextDouble() * 360;
-            return new Vector2((float)Math.Cos(randomDegree) * radius, (float)Math.Sin(randomDegree) * radius);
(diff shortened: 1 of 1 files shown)


Source/Engine/Utilities/Extensions.cs
```diff
@@ -300,6 +300,18 @@ public static Vector2 NextUnitVector2(this Random random, float radius = 1.0f)
...
+        
+        /// <summary>
+        /// Generates a random <see cref="Vector2"/> point on a circle of a given radius.
+        /// </summary>
+        /// <param name="random">An instance of <see cref="Random"/>.</param>
+        /// <param name="radius">Radius of circle. Default 1.0f./>.</param>
+        /// <returns>A random <see cref="Vector2"/>.</returns>
+        public static Vector2 NextUnitCircleVector2(this Random random, float radius = 1.0f)
+        {
+            double randomDegree = random.NextDouble() * 360;
```
mafiesto4, line 312: I think this should be: `double randomDegree = random.NextDouble() * Mathf.RevolutionsToRadians;` so cos/sin will get proper value in radians.

rkrahn: You're right. I assumed it was degrees since the result was still what was desired. Fixed in new commit.

PR #2270, opened 2024-02-22
Attempt to fix `IAssetFactory::Get()` in plugins on Linux.

Menotdan:
Previously, on Linux, any use of `IAssetFactory::Get()` in another module, such as a plugin, created a new `Dictionary` which prevents external asset factories from being registered. This fixes that by moving `static Collection Factories;` into the class itself instead of in a function, 
and adding a `Dictionary<StringView, IAssetFactory*> IAssetFactory::Factories;` in `Content.cpp`, but maybe there is a better place for it, feel free to move it.

Currently this breaks for `CollisionData` assets, but I'm not sure why as seemingly every other data type works perfectly.

### DIFF
--- a/Source/Engine/Content/Content.cpp
+++ b/Source/Engine/Content/Content.cpp
@@ -912,6 +912,8 @@ bool Content::IsAssetTypeIdInvalid(const ScriptingTypeHandle& type, const Script
     return true;
 }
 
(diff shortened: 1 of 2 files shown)


Menotdan:
Ok, apparently this breaks the factory for `FlaxEngine.CollisionData`. I'm not sure what the correct solution is, I just know this fixes the plugin's asset registration.

PR #2292, opened 2024-03-01
Re-add important inputs to GameWindow

Tryibion:


GoaLitiuM:
I would also add hotkeys for starting/stopping the profiler and also clearing the profiler results from `SetupCommonInputActions`

Issue #1796, opened 2023-10-24
Transform Scaling applied wrong to Children


schmidt-florian:
**Issue description:**
The scaling to the children is applied wrong. Maybe related to: #1350

https://github.com/FlaxEngine/FlaxEngine/assets/14238126/8dc9bf89-c609-41b1-a250-4b4a1494754e

**Steps to reproduce:**
Open attached project.
Open `Content/Scene`
Change scale from Floor on X from 20 to like 50

**Minimal reproduction project:**
[BugTest1.zip](https://github.com/FlaxEngine/FlaxEngine/files/13125217/BugTest1.zip)

**Flax version:**
master 806590d

mafiesto4:
When using:
```cpp
// Transform::LocalToWorld
(GetWorld() * other.GetWorld()).Decompose(result);
```
Fixes the scaling but if there are additional transformations (eg. that whole object has another parent with some rotation) then we end up with weird issues:

https://github.com/FlaxEngine/FlaxEngine/assets/3464853/a555c533-32fc-4f12-93d0-581ac9489214

We need to find a better solution to this problem.

mafiesto4:
It looks like Unity is also able to break visuals when we have different rotations/scales combined in:
![image](https://github.com/FlaxEngine/FlaxEngine/assets/3464853/55d7b6fd-7fd0-48b0-a293-60e02c4e38fb)

mafiesto4:
Fixed in 9a21cfd092621fdb5bc8bfe31e6d968c8095094a.

Issue #1350, opened 2023-08-27
Incorrect transform on change model scale


RuanLucasGD:
**Issue description:**

If i change actor scale, the model child don't change your local scale and local position correcly.

Unity:

[Desktop 2023.08.27 - 09.53.15.01.webm](https://github.com/FlaxEngine/FlaxEngine/assets/79365912/cd5e98c2-137c-4698-af83-00f2b923c71f)

Flax:

[Development 2023.08.27 - 09.54.31.02.webm](https://github.com/FlaxEngine/FlaxEngine/assets/79365912/5d727d21-c53e-4253-ad55-556f6f1ef55e)

**Minimal reproduction project:**
[Transform Test.zip](https://github.com/FlaxEngine/FlaxEngine/files/12448122/Transform.Test.zip)

**Flax version:**
1.6

mafiesto4:
That's because Flax inverts the normal vectors (inversed culling) when object's scale is negative:

![image](https://github.com/FlaxEngine/FlaxEngine/assets/3464853/a07777f7-4f23-4604-adb6-e3e674f37e7c)

This allows to easily build environments with objects being inverted. But I guess we could invert only along the negative axis. You can overcome this now by using two-sided material.

Menotdan:
> That's because Flax inverts the normal vectors (inversed culling) when object's scale is negative:
> 
> ![image](https://user-images.githubusercontent.com/3464853/267034423-a07777f7-4f23-4604-adb6-e3e674f37e7c.png)
> 
> This allows to easily build environments with objects being inverted. But I guess we could invert only along the negative axis. You can overcome this now by using two-sided material.

I don't believe this is the only issue being pointed out, I think they are also pointing out the difference in how the actual object scales. I'm not sure how to make it work properly though.

mafiesto4:
That's the same problem as #1796. We probably need a different of calculating Matrix for rendering.

mafiesto4:
Fixed in 9a21cfd092621fdb5bc8bfe31e6d968c8095094a.

PR #2295, opened 2024-03-02
Fix `Transform::LocalToWorld` not applying rotation to scale properly

GoaLitiuM:
Alternative fix for #1796 #1350


GoaLitiuM:
The math tests seems to fail, not sure if the tests itself are correct and doesn't account for the rotation and scale in children correctly.

mafiesto4:
hmm it doesn't fix it properly to me.

Before:
image

After:
image

GoaLitiuM:
Hmm, it seems to only work when the parent rotation is zeroed.

GoaLitiuM:
Closing in favor of #2301

PR #2299, opened 2024-03-04
vertex snapping refactor corrections

cNori:
fixed selecting other object the selected transform tool was staying in the same place

removed
public bool CanVertexSnap was calling too many times to heavy function (wrong implementacion)

removed `Find the closest object in selection (in case ray didn't hit anything)`
part because it is an edge case and has weird behavior with selection

move comment to correct place 
changed some stuff for a future
added to do in [StaticModelNode.cs](https://github.com/FlaxEngine/FlaxEngine/pull/2299/files#diff-e4f80873cfa59fec67cdaf216e3f9d844ff60b5c6c5f259eefb484e8fa13d87c) because it will cause the problem if is not fixed at some point


mafiesto4:
Thanks!

PR #2301, opened 2024-03-04
Update world transform of child objects once on change

GoaLitiuM:
Alternative fix for #1796 #1350

### DIFF
--- a/Source/Editor/Viewport/MainEditorGizmoViewport.cs
+++ b/Source/Editor/Viewport/MainEditorGizmoViewport.cs
@@ -782,7 +782,13 @@ namespace FlaxEditor.Viewport
                     trans.SetRotation(ref world);
                     trans.Translation += world.TranslationVector;
                 }
-                obj.Transform = trans;
+                if (obj is ActorNode actorNode)
+                {
+                    actorNode.Actor.Position = trans.Translation;
+                    actorNode.Actor.Orientation = trans.Orientation;
+                }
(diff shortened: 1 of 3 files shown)


mafiesto4:
It makes the mesh look fine but editing with gizmo is not accurate. Also, with current fix it supports skewing the meshes from multiple transformations being applied which is a nice feature. Ofc this fix has nice perf benefit.

Before:

https://github.com/FlaxEngine/FlaxEngine/assets/3464853/caf403b8-5c6f-4304-90af-0f54c181ef80

After;

https://github.com/FlaxEngine/FlaxEngine/assets/3464853/9462203f-50b1-48f6-b85f-748d2df5c1db

GoaLitiuM:
The performance was my main concern with the recent changes. I can't reproduce the issue in the second video. Maybe the gizmo rotation shouldn't change the whole transform of the actor? I can see all values including scale changing whenever I rotate the actor.

GoaLitiuM:
@mafiesto4 Can you check if these Gizmo related changes fixes the unstable rotation?

GoaLitiuM:
Closing this one, the performance concerns would go away if the cached actor `_transform` is used instead of recalculating the transform during rendering of the actors (in `StaticModel::Draw` and probably other places as well). The Gizmo change might be worth salvaging as it feels more stable and doesn't cause all transform values to change when rotating the actor along one axis.

PR #2304, opened 2024-03-06
Fix transform when re-parenting actor.

### LINKED ISSUE #2297: Funky behavior with scales and rotation when changing actor hierarchy.
**Issue description:**
I have a rotated and non-uniformly scaled table, which I would like to child some objects to. However, when I do so, their position and scaling is messed up and becomes unusable.
https://streamable.com/ijnjet

**Steps to reproduce:**
1. Create an empty actor.
2. Rotate it around the Y axis by some small amount, maybe 15 degrees.
3. Scale the actor by some non-uniform set of numbers.
4. Create a cube actor and rotate it a bit.
5. Add the cube as a child of the empty actor, the transformation will break.

[break scale.zip](https://github.com/FlaxEngine/FlaxEngine/files/14479600/break.scale.zip)
(under "Repro" in the main scene, take the empty actor with children and child it to the empty actor without children)

**Flax version:**
master e68c7366071df0615d1d8f6f1fc70dc5e9f7a6ad

Tryibion:
Fix #2297 . Not sure if this causes other problems in different areas of the engine, but it does fix the issue... it seems like the code was trying to keep the exact world transform and not keep the local transform. it seems unintuitive to me and I would expect the code to keep the local transform when reparenting which is what this PR does. The issue did not seem to say what the desired outcome would be, so this would be my desired outcome. But it is interesting because Unity does something similar to the original code, so maybe I just want something different...

### DIFF
--- a/Source/Editor/Windows/SceneTreeWindow.cs
+++ b/Source/Editor/Windows/SceneTreeWindow.cs
@@ -170,7 +170,7 @@ namespace FlaxEditor.Windows
             if (parentActor != null)
             {
                 // Use the same location
-                actor.Transform = parentActor.Transform;
+                actor.LocalTransform = Transform.Identity;
 
                 // Rename actor to identify it easily
                 actor.Name = Utilities.Utils.IncrementNameNumber(type.Name, x => parentActor.GetChild(x) == null);
--- a/Source/Engine/Level/Actor.cpp
+++ b/Source/Engine/Level/Actor.cpp
@@ -317,12 +317,7 @@ void Actor::SetParent(Actor* value, bool worldPositionsStays, bool canBreakPrefa
 
     // Update the transform
     if (worldPositionsStays)
-    {
-        if (_parent)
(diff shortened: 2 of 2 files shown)


Menotdan:
Hm, I still seem to be getting some issues when doing the same thing, but it's a different behavior.

https://github.com/FlaxEngine/FlaxEngine/assets/32620310/d2fd7ea8-6846-454c-964e-1217e3af4971

Tryibion:
yeah, I think my desired behavior of keeping the local transform on reparenting is different than what others would want (different than other game engines)... so I guess I will close this PR...

Menotdan:
Ah, ok. Didn't realize that's what that was doing. I wonder what the actual problem is..

Menotdan:
Must be a regression from 9a21cfd092621fdb5bc8bfe31e6d968c8095094a I guess.

Tryibion:
Was thinking of this differently...

PR #2314, opened 2024-03-10
Store Debug Log view options in Editor options

GoaLitiuM:
Save and persist the Debug Log options when Editor is closed and opened again. Also adds the same options in the view to Editor options Interface-tab.


GoaLitiuM:
Rebased and updated the order of the new bindings

Issue #2311, opened 2024-03-08
undefined behavior of SetEnabled(bool value) when seting it vita editor in play mode while using properties in c# 


cNori:
**Issue description:**

i was traing to do this 
```c#
public bool Release 
{ 
        get => m_Release;
        set 
        {
            m_Release = value; 
            if (m_Release && IsDuringPlay) Missle.GetScript<MislleControler>().ReliseFromRack(); 
        }
}
public void ReliseFromRack()
{
        Scripting.InvokeOnUpdate(() =>
        {
            ParticleEffect.IsLooping = true;
            ParticleEffect.IsActive = true;
            rb = (RigidBody)Actor;
            rb.IsKinematic = false;
            //Enabled = true;
            this.Actor.SetParent(Scene, true, false);
        });
}
```
but it keeps crashing 
looking at the code the SetEnabled(bool value) is failing to do it job
there is 100% a way to recover from it instead of having a assert it this place 

![image](https://github.com/FlaxEngine/FlaxEngine/assets/162097313/f4ee3db2-36d6-4971-939e-ccd53a120308)

**How to trigger it**
- press play
- select 1 script containing property
- press on the field to set it 
- now go to the secend script and try to enable it 

it will hit the assert

**Flax version:**

master around this point https://github.com/FlaxEngine/FlaxEngine/commit/d4e0023925f2f4df79038d2ae1391ae2292baa04

cNori:
i think i found a problem
if OnStart contains Enabled = false; and object get enabled but it was never enabled at start it will crash 
call order
SetEnabled(true); calls Start and Enable
Start (but start calls SetEnabled(false);)
look like race condition

PR #2316, opened 2024-03-10
fix for #2311

cNori:
~~replaced ASSERT with Platform::CheckFailed for now, the https://github.com/FlaxEngine/FlaxEngine/issues/2311 can be moved to 1.9~~

added a check and a massage 
image
ps.
there is not real way to fix it in the sake what state is should be selected

### DIFF
--- a/Source/Engine/Scripting/Script.cpp
+++ b/Source/Engine/Scripting/Script.cpp
@@ -66,13 +66,13 @@ void Script::SetEnabled(bool value)
                     {
                         //let the API user know abaut this
                         //silence is the worst
-                        LOG_STR(Warning, 
+                        LOG_STR(Warning,TEXT(
                             "Can't modify Enabled state in OnStart() because it was modyfaied before object got a chanse to complite the last funcion call.\n"
                             "This might cause unexpected behavior in youre scripts.\n"
                             "See:\n"
                             "https://docs.flaxengine.com/manual/scripting/events.html?tabs=code-csharp\n"
                             "Note:\n"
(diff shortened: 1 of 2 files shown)


Menotdan (changes requested):
Consider changing the message to be something like:
```
"Disabling the enabled state of an object in OnStart() is not supported.\n"
"OnStart() is called just before OnEnabled(), which expects the enabled state to be true.\n"
"See:\n"
"https://docs.flaxengine.com/manual/scripting/events.html?tabs=code-csharp\n"
"If you want to force the object to be disabled when the game is started, consider using OnAwake()."
```

cNori:
i think now it is stable

Menotdan:
```
"Modification of the Enabled state failed, state was modified before the object got a chance to call Enable().\n"
"This may cause unexpected behavior in your scripts.\n"
"See:\n"
"https://docs.flaxengine.com/manual/scripting/events.html?tabs=code-csharp\n"
"Note:\n"
"If you were trying to disable the object when the game starts, consider using OnAwake()."
```
Maybe a better message now? I tried to modify it less and keep the language more readable.

cNori:
> ```
> "Modification of the Enabled state failed, state was modified before the object got a chance to call Enable().\n"
> "This may cause unexpected behavior in your scripts.\n"
> "See:\n"
> "https://docs.flaxengine.com/manual/scripting/events.html?tabs=code-csharp\n"
> "Note:\n"
> "If you were trying to disable the object when the game starts, consider using OnAwake()."
> ```
> 
> Maybe a better message now? I tried to modify it less and keep the language more readable.

look ok

PR #2320, opened 2024-03-14
initialize Quaternion to Quaternion::Identity by default

cNori:


mafiesto4:
This affects performance thus we don't initialize structures such as `Quaternion`, `Transform`, `Vector3` and so on.
Alternatively, there is a common practise to provide special `ctor` that accepts enum for identity initialization like PhysX does:

https://github.com/NVIDIA-Omniverse/PhysX/blob/439b4767167e1753d335b613b059140d4d1b4757/physx/include/foundation/PxQuat.h#L52-L62

cNori:
values 0,0,0,0 is a invalid Quaternion for any kind of math 
it is basically a undefined rotacion
with can produse a inf , - inf , nan or other invalid Quaternion
the length needs to be neer 1 for a valid quaternion

PR #2330, opened 2024-03-17
LODIndexFence

cNori:
fixes 2 cases
case 1 index was invalid
case 2 model was not ready for use
this this case the code with wait it will print a information about this
```
[ 00:00:26.749 ]: [Warning] Can't get Model LODs, Model is not initialized yet. awaiting...
[ 00:00:26.751 ]: [Info] Model is now initialized... took ~1ms
```
it not always triggered looks like it is something to do with async loading of the asset its 90% time triggering when model is first loaded and is not existing the the scene

awaiting time is dependent on model complexity

### DIFF
--- a/Source/Engine/Content/Assets/Model.cpp
+++ b/Source/Engine/Content/Assets/Model.cpp
@@ -782,7 +782,7 @@ Check_Again:
         else
         {
             int i = 0;
-            LOG_STR(Warning, TEXT("Can't get LOD Model is not initialized awaiting..."));
+            LOG_STR(Warning, TEXT("Can't get Model LODs, Model is not initialized yet. awaiting..."));
             while (!IsInitialized()) //Wait for a model
             {
                 Platform::Sleep(1);
--- a/Source/Engine/Content/Assets/Model.cpp
+++ b/Source/Engine/Content/Assets/Model.cpp
@@ -781,18 +781,40 @@ Check_Again:
         }
         else
         {
+            
             int i = 0;
             LOG_STR(Warning, TEXT("Can't get Model LODs, Model is not initialized yet. awaiting..."));
-            while (!IsInitialized()) //Wait for a model
+            const double startTime = Platform::GetTimeSeconds();
+            if (!WaitForLoaded())
             {
-                Platform::Sleep(1);
-                i++;
-                if (i == 1000) {//give it 1000 try's
-                    LOG(Error, "Timeout Model is now initialized in expected time of 1000ms ""\nStackTrace:\n{ 0 }", DebugLog::GetStackTrace());
+                auto timespan = (Platform::GetTimeSeconds() - startTime);
+
+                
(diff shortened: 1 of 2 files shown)


Menotdan:
What exactly are you doing to cause the issue this is trying to fix? This doesn't seem like a good solution..

cNori:
> What exactly are you doing to cause the issue this is trying to fix? This doesn't seem like a good solution..

model.Model.GetBox(Transform.Identity); when is called in OnAwake is has chanse to crash the engine even if index provided is valid but model is not ready yet 
to read the LODs 

if provided LOD index is out of range it will also crash

cNori:
> The correct way to access any asset data is to ensure it has been loaded - Content load task might modify the asset before so code that doesn't check might fail due to memory/threading issues. Use `asset.WaitForLoaded()` that returns `true` if asset is not loaded properly.

i will try it if it fails i will follow up

Menotdan:
I'm still not sure what the point of the `LODIndexFence` function is, you should just be able to use `WaitForLoaded()` directly in the problematic code, and then once it's loaded (if it succeeds), you can check if the index is valid.

cNori:
> I'm still not sure what the point of the `LODIndexFence` function is, you should just be able to use `WaitForLoaded()` directly in the problematic code, and then once it's loaded (if it succeeds), you can check if the index is valid.

there is not cheaks acualty saing u need to use WaitForLoaded unless u know why this is happening remember some devs are beginers and might just keep crashin the engine

one of the rules of software engineering is don't trust user input
2.
code should not allows to access invalid index in first place it makes the code kinda explosive in case asserts are disabled the os will slap u insted of flax errors

Menotdan:
> there is not cheeks casualty saying u need to use WaitFor Loaded unless u know why this is happening remember some devs are beginners and might just keep crashing the engine
> 
> one of the rules of software engineering is don't trust user input 2. code should not allows to access invalid index in first place it makes the code kinda explosive in case asserts are disabled the os will slap u instead of flax errors

I'm not sure how using an engine provided function to handle waiting for engine provided asset loading is trusting user input lol (unless I'm misunderstanding something you said)

And yes you shouldn't allow access to an invalid index, that's what calling the `IsValidIndex()` function is for.

cNori:
> I'm not sure how using an engine provided function to handle waiting for engine provided asset loading is trusting user input lol (unless I'm misunderstanding something you said)

in this case it will say it cant load it so asset missing or something else is wrong

other aprouthe is just to throw a error and, let the user deal with it

Menotdan:
> in this case it will say it cant load it so asset missing or something else is wrong
> 
> other aprouthe is just to throw a error and, let the user deal with it

Yes, WaitForLoaded() will complain if it can't load the asset. This is normal behavior and should be standardized across the engine. If you don't like how that error is reported, you can change it. The asset failing to load is independent of handling the model LOD indices. I suggest simply adding `WaitForLoaded()` on the functions which are currently causing crashes, and then once you've called `WaitForLoaded()`, check if the LOD index is valid.

PR #2354, opened 2024-03-22
Allow multiple collision actors to be created from context menu.

Tryibion:
A nice QoL addition that allows for multiple collision actors to be created if multiple staticmodels are selected and the context menu is used to add colliders.

### DIFF
--- a/Source/Editor/SceneGraph/Actors/StaticModelNode.cs
+++ b/Source/Editor/SceneGraph/Actors/StaticModelNode.cs
@@ -89,71 +89,87 @@ namespace FlaxEditor.SceneGraph.Actors
 
         private void OnAddMeshCollider()
         {
-            var model = ((StaticModel)Actor).Model;
-            if (!model)
-                return;
+            // Allow collider to be added to evey static model selection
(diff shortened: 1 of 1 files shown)


Source/Editor/SceneGraph/Actors/StaticModelNode.cs
```diff
@@ -89,71 +89,87 @@ public override void OnContextMenu(ContextMenu contextMenu, EditorWindow window)
...
         {
-            var model = ((StaticModel)Actor).Model;
-            if (!model)
-                return;
-
-            // Special case for in-built Editor models that can use analytical collision
-            var modelPath = model.Path;
-            if (modelPath.EndsWith("/Primitives/Cube.flax", StringComparison.Ordinal))
+            // Allow collider to be added to evey static model selection
+            var selection = Editor.Instance.SceneEditing.Selection.ToArray();
```
mafiesto4, line 93: Can you pass `EditorWindow window` from `OnContextMenu` here and handle using it in Prefab window?

Tryibion: yes.

PR #2357, opened 2024-03-23
Prevent creating new project into non-empty folder

GoaLitiuM:
Error is now thrown if the new project folder is not empty. This prevents a case where user tries to create project following way: `FlaxEditor -new <path_to_project_folder>`

The command is interpret as `FlaxEditor -new` because project path is expected to be specified with `-project <path_to_project_folder>` in addition to the `-new` switch, and the project is thus accidentally created on the current working directory.


mafiesto4:
I think it would be nice to leave the logic working in case of opening project with `-new` that already exists like it used to do now.

PR #2368, opened 2024-03-27
Add support for log contexts, and add logging of the loading object's GUID when finding an object by GUID failed.

Menotdan:
This PR adds support for logging contexts, which  are a simple way to pass information about the caller of a function when it passes through a large call stack and when parameters would be unnecessary. This adds a lot of value to the log in several situations.

For example, if a script on a prefab is referencing an actor that is in a scene, and the developer uses the prefab in another scene, the script's reference to the actor will be broken, and a log will be shown. In the current version of the log, it will simply say what object failed to be loaded. This means you can't search for it using the content finder (Ctrl+O) because the object doesn't exist. This PR adds an extra piece of information to the log: the GUID of the script which tried to load the object, allowing you to find the script which is missing a reference much more easily.

This system can further be expanded with more context information (the name of the field that failed to load, other kinds of contexts, etc).

If this PR gets merged, I will make a PR which adds more places with log contexts and expands on the system more.

This PR also adds functionality for the GUID of a script to be searched using Ctrl+O, although currently it will just highlight the script's parent actor the same way simply searching for an actor's GUID does. This feature can also be expanded upon.

### DIFF
--- a/Source/Editor/Modules/ContentFindingModule.cs
+++ b/Source/Editor/Modules/ContentFindingModule.cs
@@ -228,12 +228,22 @@ namespace FlaxEditor.Modules
                         new SearchResult { Name = item.ShortName, Type = assetItem.TypeName, Item = item }
                     };
                 }
-                var actor = FlaxEngine.Object.Find<Actor>(ref id);
+                var actor = FlaxEngine.Object.Find<Actor>(ref id, true);
                 if (actor != null)
                 {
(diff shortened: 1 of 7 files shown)


mafiesto4:
I did some tweaks to simplify the code and moved that first scope from Script to generic SceneObject so it works for Actors too. Now we can move forward and track more stuff to show the errors properly in Editor.

PR #2388, opened 2024-04-01
.NET8: Move generated marshallers into separate namespace

GoaLitiuM:
Avoid polluting the `FlaxEngine` namespace with interop related marshallers, move those to nested namespace called `Interop` where most of the common marshallers are placed already. Also includes few minor refactoring in the touched areas.


mafiesto4:
Merged.

PR #2391, opened 2024-04-03
Extended math lib

cNori:
Added
C# and c++

**Quaternion.GetRotationFromNormal**
A special function for converting normal from ray cast to the rotation
When normal is axis-aligned, the Quaternion. Direction is not enough

**Transform. AlignRotationToNormalAndSnapToGrid**
Advanced function for snapping to grid with is in local space and aligning to normal vector

**Vector3. SnapToRotatedGrid**
function for snapping to grid with is in local space

**Vector3. SnapToRotatedGridWithOffset**
function for snapping to grid with is in local space with offset in local space

added missing
C++ was missing snap to grid function

### DIFF
--- a/Source/Engine/Core/Math/Quaternion.cpp
+++ b/Source/Engine/Core/Math/Quaternion.cpp
@@ -534,13 +534,13 @@ void Quaternion::RotationYawPitchRoll(float yaw, float pitch, float roll, Quater
     result.Z = cosYawOver2 * cosPitchOver2 * sinRollOver2 - sinYawOver2 * sinPitchOver2 * cosRollOver2;
 }
 
-Quaternion Quaternion::GetRotacionFromNormal(const Vector3& InNormal, const Transform& InRefrenceTransform)
+Quaternion Quaternion::GetRotationFromNormal(const Vector3& InNormal, const Transform& InReferenceTransform)
 {
-    Float3 up = InRefrenceTransform.GetUp();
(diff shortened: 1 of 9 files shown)


cNori:
> Thanks!
> 
> removed `Vector3.GetClosest` function as it doesn't make sense (not a good location for it plus it doesn't return closest point but only the one at actual location from input list).

yes this is a point of the function to return closest point from a list to the vector
with is defined as a self in the funcion 


also geting the closest point to a vector from a list is a common operation

usage example
```c#
    public virtual void Snap(RayCastHit rayCastHit,PartControler Other)
    {
        ComputeBounds();
        var offset = ((ModelBoundingBox.Size.Z) * 0.5f * Actor.Scale.Z) - ModelBoundingBox.Center.Z;
        Transform = SnapToGridXZ(rayCastHit.Point, rayCastHit.Normal, offset, rayCastHit.Collider.Transform, Transform.Scale,new Vector3(5f));
        if (Other != null)
        {
            Vector3 v = Other.Transform.WorldToLocal(rayCastHit.Point);
            var id = v.GetClosest(ref Other.SnapPoints, 5f);
            if (id != -1)
            {
                Actor.Position = Other.Transform.LocalToWorld(Other.SnapPoints[id]) + (Vector3.Forward * offset) * Transform.Orientation;
            }
        }
    }
```
i added it to avoid doing the 
```c#
GameUtils.GetClosest(ref v,ref Other.SnapPoints, 5f);
```

Issue #1132, opened 2023-05-29
Content improvements


PhyresiCompany:
- Add a tag to the filter
![Captura de Tela (162)](https://github.com/FlaxEngine/FlaxEngine/assets/101017436/3e2ac162-9f07-4ed1-a38a-b7ef7fc57625)

- Add an option to remove all filters

**Create an asset from another asset as is done in the material instance from the material**
![Captura de Tela (163)](https://github.com/FlaxEngine/FlaxEngine/assets/101017436/e4b9376c-d7ea-4112-bce3-dbc283c27055)

- Create a `Particle System` from a `Particle Emitter`
- Create an `Animation Graph` from a `Skinned Model`
- Create a `Prefab Variant` from a `Prefab`

_If anyone has any more ideas for another asset, please leave a comment._

PR #2415, opened 2024-04-11
New content creation options and features.

Menotdan:
This adds several new features to the content creation workflow.

1. Creating particle systems from particle emitters.  This simply creates a new particle system asset with a track containing the particle emitter asset.
2. Creating animation graphs from skinned models. This simply sets the base model of the animation graph to the skinned model asset the graph was created from. This feature is currently a bit hacky because visject, it opens an animation graph editor window to set the base model and then save that, this avoids reusing code but the behavior may seem weird to users, maybe leave the window open after the operation.
3. Picking the root asset type of a prefab:
image
4. Picking the root control type (or canvas) for a widget:
image
image

This also adds showing prefabs which have a `UICanvas` or `UIControl` as the root as "Widget" instead of "Prefab", to match how the UI Editor works. Currently this is cached because its potentially slow as I load the prefab into memory and create a default instance of it, but that could break when people change whether or not a prefab is a "widget" or not, since it won't update.
image

This PR is based off of #1132.

### DIFF
--- a/Source/Editor/Content/Create/PrefabCreateEntry.cs
+++ b/Source/Editor/Content/Create/PrefabCreateEntry.cs
@@ -3,12 +3,13 @@
 using System;
 using FlaxEditor.Scripting;
 using FlaxEngine;
+using FlaxEngine.GUI;
 using Object = FlaxEngine.Object;
 
 namespace FlaxEditor.Content.Create
 {
     /// <summary>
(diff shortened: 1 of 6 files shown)


Menotdan:
@mafiesto4 could you add a TODO to make a better system for differentiating widgets and regular prefabs? Loading the whole prefab could cause issues in some projects sometimes, but I'm not sure there is a better option atm... Maybe if there was a seperate asset for widgets but yea idk

PR #2418, opened 2024-04-11
Fix particle emitter creation from templates.

Menotdan:
This PR fixes a bug with the content cloning system which caused new particle emitters created from the editor-provided templates to break until you restarted the editor, or even to fail to clone at all. The issue was due to `Content::CloneAssetFile()` not running on the main thread, so I moved it to the main thread using a `MainThreadTask`. I haven't done extensive testing but it appears to fix the particle emitter issue.


mafiesto4:
Wow! thanks

PR #2423, opened 2024-04-12
Added CAS sharpening to SMAA antialiasing. 

Muzz:
Just to explain motivation here. 

I'm not a fan of the ghosting of TAA, but i think most people would say it looks best in static screenshots because of the sharpening inside of it. 

So i wanted something that had that crispy look, but more configurable, so I wrote an implementation based on https://gpuopen.com/fidelityfx-cas/

Currently it's always on when SMAA is active, so you may want to throw in a switch for it. 

give it a shot, i think it looks quite nice. 

examples:

NO AA

image

TAA

image

SMAA with CAS

image

### DIFF
--- a/Content/Shaders/SMAA.flax
+++ b/Content/Shaders/SMAA.flax
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:65a272fec450ab68086794cb7db73956a8bc9fbcd17c4820564d23fe352ab3a6
-size 46449
+oid sha256:4d9f21478bf454bd713b147a776b6febfb38c7e38845eef7974e16886c8cfe7a
+size 48460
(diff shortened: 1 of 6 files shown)


Muzz:
Not sure correct way to push a modified .flax shader because of the gitlab LFS btw, so i just pushed with no verify. I know it's a generated file so i think that's ok?

Menotdan:
Did you change something between the first two and the last screenshot? the dark cracks on the rock texture are significantly darker on the last image.

Muzz:
> Did you change something between the first two and the last screenshot? the dark cracks on the rock texture are significantly darker on the last image.

That's just the results of a different sharpening algorithm.

cNori:
> Not sure correct way to push a modified .flax shader because of the gitlab LFS btw, so i just pushed with no verify. I know it's a generated file so i think that's ok?

if u need to submit modified .flax use the zip file for it

mafiesto4:
That's fine I'll update shader binary, shader source is only needed for this.

mafiesto4:
Thanks! I added it in 09737023a13fcbbb9b2156f77a00c37f808f70f9 as general pass for SMAA/FXAA/NoAA.

Dummy docs: 1b5f91ce4fa75c2ff22e2c904d431b04b19d18b4

Issue #1842, opened 2023-10-31
Update tracy version to v0.10


MineBill:
Tracy recently released version 0.10 and the protocol is incompatible.
![image](https://github.com/FlaxEngine/FlaxEngine/assets/30367251/2069c16d-f88c-4141-8c1d-f7a44ae56708)

mafiesto4:
Yeah Tracy was updated recently so we gonna migrate to the latest version soon. Most likely for 1.8. Meanwhiel you can use previous Tracy version for profiling.

mafiesto4:
Added in 7dc645c1147e61895ef39500c6c7b451d74b3a35.

PR #2433, opened 2024-04-13
Integrate latest Tracy v0.10.0

yhyu13:
Simply updating tracy to v0.10.0 latest

- no new feature added to flax engine

Three changes kept when upatding
- v0.10.0 has removed ZoneScope::Begin/End, we added them back as they are required for profiling event begin/end
- Add Flax Engine Char adaption in tracy
- Using Flax Engine LZ4 instead of tracy lz4
-Compile flags kept


GoaLitiuM:
Isn't latest version of Tracy already supported? https://github.com/FlaxEngine/FlaxEngine/issues/1842

GoaLitiuM:
This looks like changeset of v0.11.0 or future version of Tracy

yhyu13:
@GoaLitiuM I am on the latest commit of tracy https://github.com/yhyu13/tracy/commit/b5b985d1d09e2c757338a51f8a82115626bf81ab

which is still v0.10.0 on the server UI.

mafiesto4:
I think it would be better to wait for the official `v0.11.0` version release with that new protocol version.

PR #2462, opened 2024-04-20
Optimize TreeNode rendering

GoaLitiuM:
Similar style of optimization than with the RichTextBox line rendering, drawing should try to not go through every single children in the tree if the children are sorted from top to bottom.

### DIFF
--- a/Source/Editor/GUI/Tree/TreeNode.cs
+++ b/Source/Editor/GUI/Tree/TreeNode.cs
@@ -713,6 +713,71 @@ namespace FlaxEditor.GUI.Tree
         }
 
         /// <inheritdoc />
+        protected override void DrawChildren()
+        {
+            // Draw all visible child controls
+            var children = _children;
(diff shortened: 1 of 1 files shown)


Source/Engine/Level/Actor.h
```diff
@@ -382,6 +382,15 @@ API_CLASS(Abstract) class FLAXENGINE_API Actor : public SceneObject
         return _isActiveInHierarchy != 0;
     }
 
+    /// <summary>
+    /// Gets value indicating if actor is in a scene.
+    /// </summary>
+    API_PROPERTY(Attributes="HideInEditor, NoSerialize")
+    FORCE_INLINE bool IsInScene() const
```
mafiesto4, line 389: I would use name `HasScene` to match naming convention in Actor: `HasParent`/`HasTag`/`HasPrefabLink`/..

mafiesto4 (changes requested):

PR #2468, opened 2024-04-21
Release Prefab assets when unloading Scripting service

GoaLitiuM:
Fixes a crash caused by release of Prefabs containing references to already unloaded scripting assemblies assets such as Scripts when closing the engine.


mafiesto4:
Have you looked into `Prefab.cpp`? It already destroys default prefab instance and local cache on `Scripting::ScriptsReloading` and `Scripting::ScriptsUnload`. Maybe we need to clear some additional cached on C# side or in Editor. If we unload `Prefab` objects then the asset rrefs might be missing in Level/Scripts.

GoaLitiuM:
Fixed by using the `Scripting::ScriptsUnload` events instead.

PR #2532, opened 2024-05-01
Fix looping audio sources not looping seamlessly

GoaLitiuM:


GoaLitiuM:
I think the original code tried to queue an additional buffer to the audio engine, but with short audio clips having only one buffer, the queue becomes empty for a brief time after the audio finishes looping.

GoaLitiuM:
I found the non-looping audio effects are sometimes having audio glitches with these changes

PR #2554, opened 2024-05-05
Fix odd offset between 1st property and property label.

Tryibion:
Before:
image

After:
image


Tryibion:
Hmm this does break inline with more than 1 property... need to figure that out.

PR #2624, opened 2024-05-19
Add support for Windows on ARM

GoaLitiuM:
This PR adds runtime and tooling support for building ARM64 binaries for Windows (Windows on ARM).

For building ARM64 binaries on MSVC, the v143 build tools for ARM64/ARM64EC are needed in Visual Studio components in order to cross-compile the binaries from x86 system (or build natively on ARM64 system).

I also updated few of the dependencies to newer versions as some of them had added proper support for WoA, and some of the dependency scripts were updated to use CMake for building the libraries for easily target ARM64. NvCloth and PhysX fixes for WoA were already merged earlier into Flax-specific forks. Some DLL-dependencies needs to be updated manually for ARM64, namely hostfxr.dll from the ARM64 .NET runtime package and the `Microsoft.VisualStudio.Setup.Configuration.Native` package.


mafiesto4:
Thanks! I updated binaries fo the updated deps. I've got to disable change from `DeployDataStep` related to fxr host copying as it was not working properly. We can come back to it later,

PR #2635, opened 2024-05-22
Add SDF sky light shadows (help wanted)

### LINKED ISSUE #1189: Add shadowed sky light
Add support for using `Global SDF` to render shadow for Sky Light. This should reduce sky-light leaking into interiors and improve ambient lighting.

Global SDF usage in shaders: https://docs.flaxengine.com/manual/graphics/models/sdf.html#shaders

Example of raytracing in Screen Space Reflections shader:

https://github.com/FlaxEngine/FlaxEngine/blob/228ef4e1306d8d38468d2e41d578154bdd90046b/Source/Shaders/SSR.shader#L136-L150

HydrogenC:
A very primitive implementation of the SDF sky shadow (traces along the normal direction).  Closes #1189.  

ad4289a89ea403b6a9820fcc4cd7f23
Showcase rendered with only skylight enabled.  
Work to do: Do tracing along a random direction on the hemisphere around the normal and do temporal blending to achieve a good-looking soft shadow (I don't know exactly how to implement that on the current codebase, for example, how to use the sampling functions in `MonteCarlo.hlsl` since there are no samples, so help wanted).


mafiesto4:
Here is a project that uses Flax's `GlobalDistanceField` and `GlobalSurfaceAtlas` and `SDF-DDGI` algos but it also implements SDF-based ambient occlusion. I think sth similar can be used to trace sky light shadowing: https://github.com/flwmxd/LuxGI/blob/main/Code/Maple/src/Shaders/SDF/SDFAO.comp

HydrogenC:
> Here is a project that uses Flax's `GlobalDistanceField` and `GlobalSurfaceAtlas` and `SDF-DDGI` algos but it also implements SDF-based ambient occlusion. I think sth similar can be used to trace sky light shadowing: https://github.com/flwmxd/LuxGI/blob/main/Code/Maple/src/Shaders/SDF/SDFAO.comp

Thanks for that. I will take it for reference.

HydrogenC:
Not working on this recently, closing.

PR #2639, opened 2024-05-23
Add slot spacing to uniform grid panel.

Tryibion:
Using this makes everything look a little better in the editor when using the uniform grid and lets you control the x and y spacing between the grid items. This adds slot spacing which only affects the control size by controlling the space between the controls in the grid instead of just the control's size like the padding does.

Before:
image
image

After:
image
image


mafiesto4:
Cool! But can we make the horizontal margins between position/rotation fields similar to the vertical margins?

Tryibion:
> Cool! But can we make the horizontal margins between position/rotation fields similar to the vertical margins?

I had some wrong scaling in the spacing calculations. Fixing those, it looks like this with similar spacing.
image

Tryibion:
I also made the color value box be consistent with the others.
image

mafiesto4:
Yeah, now it looks sweet!

PR #2640, opened 2024-05-23
Fix a few edge cases for cursor showing/not showing

Tryibion:


mafiesto4:
Maybe add a safety counter for max iteration to prevent deadlock in that loop? Like max `100`?

Tryibion:
> Maybe add a safety counter for max iteration to prevent deadlock in that loop? Like max `100`?

Added this to the loop.

PR #2663, opened 2024-05-29
Infinite Grid

Swiggies:
An update to make the grid look much nicer. No longer uses debug lines and instead a shader on a large quad that follows the camera around. Configurable color and viewable distance for the main grid,

How it looks by default:
image

Options configurable in Viewport options:
image

Notes;
- Wasn't sure if Viewport options was the best place to put the settings for this but it also seemed like the most relevant.
- The default distance is 2500, this can feel a bit small and still considering if doubling it to 5000 would be better.
- The X and Z axis lines can be a bit difficult to see when far away. 
- At a later point I could also make it so you can configure the Y position of the grid through the View dropdown menu on the Editor window.

Despite those notes I think it's still a rather large improvement.

### DIFF
--- a/Content/Shaders/Grid.flax
+++ b/Content/Shaders/Grid.flax
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:67be5a9207994cdb858b65d9c06d763e0cb51f532e8ef37c18b24c17c216c0cd
-size 4575
+oid sha256:ff8ecc01befe7149ce3ad7e5de37503b7acc0ef25a8ba13a316aa4047b8ff6d1
+size 4882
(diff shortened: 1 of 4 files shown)


mafiesto4:
1. Can you delete ` Content/Shaders/Grid.flax` file from that fork branch so automated tests won't fail on Git LFS checkout?
2. Have you used some external reference for grid shader or are you an author?

Swiggies:
> 1. Can you delete ` Content/Shaders/Grid.flax` file from that fork branch so automated tests won't fail on Git LFS checkout?
> 
>     2. Have you used some external reference for grid shader or are you an author?

1. Done in the latest commit (I hope). Wasn't sure why it was failing so that's good to know for the future
2. Yes, I loosely used [this](https://bgolus.medium.com/the-best-darn-grid-shader-yet-727f9278b9d8#3e73) for the grid pattern and added credit in the shader file

mafiesto4:
Merged in 2a4a5d25190584bae12a79208060fac9c4c371b6.

PR #2670, opened 2024-05-31
Added reordering lists and visject parameters.

Menotdan:
This PR adds the ability to reorder lists and graph parameters in visject graphs. I think the code is a bit messy so any suggestions or changes are appreciated.


Tryibion:
Looks like LFS is failing because a .flax file has been included.

Menotdan:
> Looks like LFS is failing because a .flax file has been included in the pr.

Yeah I think the problem was introduced when I merged some change from master a while back, not sure how to fix it though.

mafiesto4:
Maybe merge `master` into this branch and revert that file changes to be from `master`? Then push this branch? Or remake PR as a fallback option.

mafiesto4:
Or you can delete that file from the branch. I'll ensure it's not removed when mering to `master`.

mafiesto4:
Merged in 892ab47b7a48ac8fde7e399853e0984d73ece936.

PR #2675, opened 2024-06-01
Visject: Spawn parameter setter node when holding ALT during drag and drop

Chikinsupu:
This PR adds the ability to easily drag and drop a parameter setter node while holding the alt key. This mimics the functionality of other visual scripting tools like Bolt. Requested in the discord server, no related issue found.

Note: This only works for visual scripting, since the visual scripting graph is the only tool that has a setter node defined.

In both gifs i hold the ALT key for the 2nd node.

Before:
ParametersAltDragSetterBefore

After:
ParametersAltDragSetterAfter


ElevenArt:
Mentioned here: https://discord.com/channels/437989205315158016/1239886455011020800/1242441424721543258

PR #2680, opened 2024-06-02
Use `PixelFormat::MAX` in `RenderToolsVulkan.cpp`

tecnessino:


mafiesto4:
That manually specified value si there on purpose so when adding new pxiel formats the code won't compile and will trigger fixing by the programmer.

tecnessino:
ok, thanks for explanation

PR #2685, opened 2024-06-03
Draw impulse/ execution lines slightly thicker than data ones

xxSeys1:
_This is my first pr so if there are any issue with code quality or stuff like that let me know :)_

Slightly increases the thickness of impulse/ execution lines for node editors to make it easier for the user to follow execution flow (and indirectly also data flow). This is the same thing Unreal Engine does.

Before:
image

After:
image

### DIFF
--- a/Source/Editor/Surface/Elements/OutputBox.cs
+++ b/Source/Editor/Surface/Elements/OutputBox.cs
@@ -198,15 +198,11 @@ namespace FlaxEditor.Surface.Elements
 
                 // TODO: Figure out how to only draw the topmost connection
                 if (IntersectsConnection(ref startPos, ref endPos, ref mousePosition, mouseOverDistance))
-                {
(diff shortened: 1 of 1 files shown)


Source/Editor/Surface/Elements/OutputBox.cs
```diff
@@ -163,6 +163,12 @@ public void DrawConnections(ref Float2 mousePosition)
                     highlight += 0.5f;
                 }
 
+                // Increase thickness on impulse/ execution lines
+                if (targetBox.CurrentType.IsVoid)
+                {
+                    highlight += 1.25f;
+                }
```
Tryibion, line 170: Only suggestion is that for single line if statements, Mafi likes them to be like this
```
if (targetBox.CurrentType.IsVoid)
     highlight += 1.25f;
```

Issue #2310, opened 2024-03-08
Problem compiling on Linux Mint 21.3


justawhitecat:
**Issue description:**
Upon trying to "Run without debugging" in VSCODE, many errors generate.

**Steps to reproduce:**
Follow the steps available in the github repo on Linux Mint 21.3

The terminal outputs the following (or, at least, the last bits of it):
[output.txt](https://github.com/FlaxEngine/FlaxEngine/files/14541340/output.txt)

mafiesto4:
`fatal error: 'new' file not found` - missing C++ stdlib header file.

Compiler command line:
```
/usr/bin/clang++-14 -fvisibility-inlines-hidden -fvisibility-ms-compat -target x86_64-unknown-linux-gnu --sysroot="/" -mssse3 -fPIC -c -pipe -x c++ -std=c++14 -Wdelete-non-virtual-dtor -fno-math-errno -fdiagnostics-format=msvc -frtti -glldb -pthread -O2 -fexceptions -D "PLATFORM_UNIX" -D "_GNU_SOURCE" -D "PLATFORM_LINUX" -D "_LINUX64" -D "UNICODE" -D "_UNICODE" -D "USE_EDITOR" -D "BUILD_DEVELOPMENT" -D "NDEBUG" -D "FLAXENGINE_API=__attribute__((__visibility__(\"default\")))" -D "COMPILE_WITH_PROFILER" -D "TRACY_ENABLE" -D "TRACY_ON_DEMAND" -I"/usr/include" -I"/usr/lib/clang/14/include" -I"/home/whitecat/.github/FlaxEngine/Source/ThirdParty" -I"/home/whitecat/.github/FlaxEngine/Source" -o "/home/whitecat/.github/FlaxEngine/Cache/Intermediate/FlaxEditor/Linux/x64/Development/Core/Vector3.o" "/home/whitecat/.github/FlaxEngine/Source/Engine/Core/Math/Vector3.cpp"
```

Probably some packages are missing.

You could verify files in folders:
* `/usr/include`
* `/usr/lib/clang/14/include`

if there are any stdlib headers. maybe we need to pick another directory on Linux.

justawhitecat:
> `fatal error: 'new' file not found` - missing C++ stdlib header file.
> 
> Compiler command line:
> 
> ```
[...]

PR #2701, opened 2024-06-11
update readme to include required linux dependency

envision3d:
I'm running Zorin 17.1 which is based on Ubuntu 22.04.  While `sudo apt-get install clang lldb lld` seemed to trigger an install of libstdc++-11, it didn't seem to suffice as I was getting compilation issues until adding 13.  I conferred with an arch user and it doesn't seem a similar addition will be necessary for arch users.

Example of issues seen:
```
FlaxEngine/Source/ThirdParty/recastnavigation/DetourNavMeshQuery.cpp(28,10): fatal error: 'new' file not found
#include <new>
         ^~~~~

...

FlaxEngine/Source/ThirdParty/fmt/format.h(38,10): fatal error: 'limits' file not found
#include <limits>
         ^~~~~~~~
```

### DIFF
--- a/Source/Editor/Content/Proxy/JsonAssetProxy.cs
+++ b/Source/Editor/Content/Proxy/JsonAssetProxy.cs
@@ -168,11 +168,18 @@ namespace FlaxEditor.Content
 
         private SpriteHandle _thumbnail;
 
+        /// <summary>
+        /// Default Constructor.
+        /// </summary>
         public SpawnableJsonAssetProxy()
         {
             _thumbnail = SpriteHandle.Invalid;
         }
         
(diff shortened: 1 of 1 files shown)


Tryibion:
When I had a similar issue I have to do this <https://github.com/FlaxEngine/FlaxEngine/issues/2310#issuecomment-2014277765>.

envision3d:
Installing g++-13 also worked for me.  Should we maybe include a section below the instructions with troubleshooting information and suggested solution steps?

Also, would g++ be preferred over manually including libstdc++?  It looks like g++ installs that and more.

PR #2705, opened 2024-06-14
add default audio listener to newly created scenes

envision3d:


xxSeys1:
Are u gonna add a warning for that with this pr? Otherwise I gonna make one for that.

envision3d:
I just quickly took the first step here based on feedback on discord.

I still think there's value in the warning, though.  Go for it!

PR #2722, opened 2024-06-19
Absolute snapping

Zode:
sorry for re-opening, apparently renaming a branch causes it to be closed :weary: 

Adds a toggle for absolute snapping, always silently disabled during local space transformations.
When enabled: adjusts transformation to the closest snap point as if the object started perfectly on the snap grid when snap hotkey is pressed for a smoother user experience.


Zode:
@mafiesto4 any news for this? :eyes: been a while and haven't heard if its OK or not at all heh

PR #2730, opened 2024-06-22
More 'moves'

mtszkarbowiak:
1. I'm going to work more on moves. 2. I'd love to be corrected.

### DIFF
--- a/Source/Engine/Level/Actor.h
+++ b/Source/Engine/Level/Actor.h
@@ -182,6 +182,11 @@ public:
     }
 
     /// <summary>
+    /// Sets the actor name without copying the string.
+    /// </summary>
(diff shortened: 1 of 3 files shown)


Source/Engine/Level/Actor.h
```diff
@@ -181,6 +181,11 @@ API_CLASS(Abstract) class FLAXENGINE_API Actor : public SceneObject
         return _name;
     }
 
+    /// <summary>
+    /// Sets the actor name without copying the string.
+    /// </summary>
+    void SetNameNoCopy(String&& value);
```
mafiesto4, line 187: Rename to `SetName`.

mafiesto4 (changes requested):

PR #2752, opened 2024-07-10
Call `Actor::OnBeginPlay` by collecting all actors instead of hierarchical by children

### LINKED ISSUE #2623: Reparenting the camera through scripting makes a single object disappear in play mode
**Issue description:**

When reparenting the camera through scripting, a single object below the camera's parents gets hidden. All objects above the parents and below the actor to get hidden stay visible.

https://github.com/FlaxEngine/FlaxEngine/assets/83252371/011427b1-933a-4861-9b82-e7333c29dba0

**Steps to reproduce:**

In any script's "OnStart", "OnEnable", or "OnAwake" method, simply reparent the camera to a child of whatever actor the script is on.

[ParentingTest.zip](https://github.com/FlaxEngine/FlaxEngine/files/15370560/ParentingTest.zip)

**Flax version:**

1.8.1 release

### LINKED ISSUE #2751: [Bug] Changing an Actor's parent in OnStart prevents the next Actor's OnStart from being called
Set a few actors up like this:
image

On each of the children,  have this simple script that reassigns their parent:
image

It'll look like this after:
image
image

Only half of them have their `OnStart` method called.  This happens if you reassign to something other than the Scene as well.  I think something in `SetParent` is causing a side effect, maybe in the locking?

Tryibion:
This allows for parents to be changed on start in a Script.

This does run slower than the previous solution, but it doesn't rely on the parent/child actor relationship to initialize the actor.

Fix #2751
Fix #2623

### DIFF
--- a/Source/Engine/Level/Actor.cpp
+++ b/Source/Engine/Level/Actor.cpp
@@ -875,14 +875,6 @@ void Actor::BeginPlay(SceneBeginData* data)
             e->BeginPlay(data);
     }
 
-    // Update children
-    for (int32 i = 0; i < Children.Count(); i++)
-    {
-        auto e = Children.Get()[i];
-        if (!e->IsDuringPlay())
-            e->BeginPlay(data);
-    }
-
     // Fire events for scripting
     if (IsActiveInHierarchy() && GetScene() && !_isEnabled)
     {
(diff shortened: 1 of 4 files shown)


Tryibion:
Taking a look at this some more. More logic will need to be added around spawning actors.

Tryibion:
Closing in favor of #2940

PR #2767, opened 2024-07-16
Fix issues with hiding and showing cursor.

### LINKED ISSUE #2765: [Bug] Cursor Visibility
**Issue description:**
Setting `Screen.CursorVisible = false` during the `OnEnable` event will prevent changing the visbility state during play. After attempting to change the visibility state during play, setting `Screen.CursorVisible = true` during the `OnDisable` for the editor will also not work.

**Steps to reproduce:**
1. Create a script, where you set the cursor visibility to false in `OnEnable`
2. Create a Method to set `Screen.CursorVisible` to `true`
3. During play mode or during game runtime, call the created method to set visibility to true
4. Cursor will not show
5. (In Editor) Cursor will not show after exiting play mode,even when `Screen.CursorVisible` is set to `true` in `OnDisable`. This only happens when the method is called. If we don't call the method and only `OnEnable` and `OnDisable` are called, cursor visibility works
 

Base

**Flax version:**
1.8.6512.2

Tryibion:
Fix #2765

### DIFF
--- a/Source/Engine/Platform/Windows/WindowsWindow.cpp
+++ b/Source/Engine/Platform/Windows/WindowsWindow.cpp
@@ -775,37 +775,32 @@ void WindowsWindow::UpdateCursor()
     // Don't hide cursor when window is not focused
     if (_cursor == CursorType::Hidden && _focused)
     {
-        if (!_lastCursorHidden)
-        {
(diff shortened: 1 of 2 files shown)


Tryibion:
Never mind, this doesnt fix that issue quite yet.

PR #2785, opened 2024-07-20
Improve engine responsiveness when not focused

GoaLitiuM:
This is particularly noticeable over Linux where copying from clipboard becomes super slow when editor is focused and trying to paste text from Flax Editor to other applications.

### DIFF
--- a/Source/Engine/Engine/Engine.cpp
+++ b/Source/Engine/Engine/Engine.cpp
@@ -195,6 +195,12 @@ int32 Engine::Main(const Char* cmdLine)
         // Use the same time for all ticks to improve synchronization
         const double time = Platform::GetTimeSeconds();
 
+        // Update application (will gather data and other platform related events)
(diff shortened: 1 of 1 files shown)


mafiesto4:
Do you think this can have some negative effect on the system events processing? For example, on Windows this will process input and window messages at different stage (outside the Update) which could result in some missing data (eg. `Time::DetlaTime` is different).

If it's a Linux-only issue maybe that platform could implement `PlatformBase::GetIsPaused` and feed `X11` with `SelectionRequest` for clipboard data at higher responsivness?

GoaLitiuM:
I don't see where the `Time::DeltaTime` is using during `Platform::Tick`, the input events are flushed during `InputService::Update`, and the input queue content would be identical in both cases.

The editor isn't paused when it's not focused, so I don't see how that would help with this either.

PR #2788, opened 2024-07-20
Add game window options into toolstrip in window for better visibility.

Tryibion:
Several users have discussed how the game window options being hidden in the context menu of the game tab is hard to find at first. This PR moves the options into a toolstrip in the window for better visibility.

New toolstrip.
image


RomanZhu:
Could it be based on the settings option?
I like to have more space for my 3d view

xxSeys1:
I mean this is not taking up thaaaat much space and I think it's important to have to options accessible in a few mouse clicks.
Maybe tuck them away in a hamburger menu?

RomanZhu:
image
For example, add here an option to draw this panel in the Game View or not

xxSeys1:
@Tryibion Hey, I made a pr into your branch that updates this branch with the newest changes to the right click menu from 1.10 (amongst other things):

https://github.com/Tryibion/FlaxEngine/pull/1

I also noticed two problems, for that see the pr.

PR #2800, opened 2024-07-25
Implement SDL3 platform, windowing and input handling

GoaLitiuM:
Implements support for using SDL3 as backend for common platform functionality, windowing and handling input. Windows and X11 under Linux are fully supported, but support for some Wayland-specific functionality like drag-and-drop, clipboard and color picker has not been implemented yet.

Some few notable changes also implemented in this branch required for SDL to properly function:

- Support for raw input mouse ("mouse relative mode") and keyboard handling was added, mostly due to how poorly the mouse input handling works in non-raw input modes. This is automatically enabled when the cursor is locked and hidden, or when cursor is hidden while mouse tracking is enabled. GUI `Control` now also has a method `OnMouseMoveRelative` which is called when mouse is moved when relative mode is enabled, providing only accurate mouse deltas instead of absolute cursor position.
- Added new command-line switches `-x11` and `-wayland` to force-enable specific SDL driver instead of automatically detecting one (Wayland is preferred over X11 at the moment).
- Enabled modern looking dialog boxes under Windows.
- Native windowing mode settings in Editor are now also available under Linux, and enabled by default.
- `CreateWindowSettings.IsRegularWindow` is now deprecated and is replaced by setting the `WindowType` `Type` to a specific window type instead. This is required as popup-windows behave differently under Linux platform, and the use-case can't be determined from the other window settings alone.
- Added new engine configuration setting `UseSDL` (command-line `-useSdl`) to enable SDL backend support, this is enabled by default. When disabled, Windows and Linux platforms will use the previous native implementation over SDL implementation.

The current version of SDL3 is not considered stable yet, so the unstable version will receive regular updates until SDL 3.2 is released.


GoaLitiuM:
The SDL library dependency can be built by running `Flax.Build -log -BuildDeps -depsToBuild=SDL,libportal,Wayland -verbose` (may fail during first run but works after second run). If you run into issues while building dependencies after branch update, try removing the `Cache\Intermediate\Deps\SDL` folder before trying again.

Here are the precompiled platform files (extract to your FlaxEngine folder) for **SDL 3.4.0**:
- Windows x64: [FlaxEngine-SDL3.4.0-only-windows-x64.zip](https://github.com/user-attachments/files/24443181/FlaxEngine-SDL3.4.0-only-windows-x64.zip)
- Windows ARM64: [FlaxEngine-SDL3.4.0-only-windows-ARM64.zip](https://github.com/user-attachments/files/24443182/FlaxEngine-SDL3.4.0-only-windows-ARM64.zip)

Tryibion:
Initial bugs found:

- [ ] Rotation in the transform gizmo does not work
- [ ] Windows: drag and drop seems to not happen on the first click and drag, it sometimes takes 2 tries to start the drag.

GoaLitiuM:
Pre-built SDL library files for Windows (extract to your FlaxEngine folder): [FlaxEngine-SDL3.2.0-only-windows-x64.zip](https://github.com/user-attachments/files/18578901/FlaxEngine-SDL3.2.0-only-windows-x64.zip)

GoaLitiuM:
> Initial bugs found:
> 
> * [ ]  Rotation in the transform gizmo does not work
> * [ ]  Windows: drags seems to not happen on the first click and drag, it sometimes takes 2 tries to start the drag.
> * [ ]  Windows: I am guessing this is the known focus issue, but it seems like some mouse clicks don't register and I have to click again for something to be selected.

A lot of changes have been made, but the fix for gizmo rotation was included in latest batch of updates.

GoaLitiuM:
New window and tab dragging system was implemented (supported on Windows and Linux X11/Wayland), which should be much stable to work with on Linux from previous system.

https://github.com/user-attachments/assets/c180a301-b69c-4b2f-8891-c1ae298953f5

Tryibion:
> > Initial bugs found:
> > 
> > * [x]   Rotation in the transform gizmo does not work
> > * [x]   Windows: drags seems to not happen on the first click and drag, it sometimes takes 2 tries to start the drag.
> > * [x]   Windows: I am guessing this is the known focus issue, but it seems like some mouse clicks don't register and I have to click again for something to be selected.
> 
> A lot of changes have been made, but the fix for gizmo rotation was included in latest batch of updates.

Yeah, looks like the drag drop issues and the focus/clicking issues are resolved. 

Additional things I noticed after the recent changes.
- [x] Double clicking on any docked window's title bar changes whether the main window is maximized or windowed.
- [ ] If a subwindow is maximized, when I drag it to un-maximize it, I can not dock it into the main window. I have to stop dragging once the window is windowed and then initiate the drag again for the docking to work.

GoaLitiuM:
Updated SDL3 to 3.2.4, which seems to fix a crash when trying to create tooltip window over context menus.

GoaLitiuM:
> * [ ]  Double clicking on any docked window's title bar changes whether the main window is maximized or windowed.
> * [ ]  If a subwindow is maximized, when I drag it to un-maximize it, I can not dock it into the main window. I have to stop dragging once the window is windowed and then initiate the drag again for the docking to work.

These should be fixed now.

Source/Engine/Platform/SDL/SDLPlatform.Windows.cpp
```diff
@@ -146,34 +124,40 @@ bool SDLPlatform::EventFilterCallback(void* userdata, SDL_Event* event)
...
-        Float2 mousePos = WinImpl::DraggedWindowMousePosition;
+        if (window)
+            window->HandleEvent(*event);
+
+        if (WinImpl::DraggedWindowSize != window->GetClientSize())
+        {
+            // The window size changed while dragging, most likely due to maximized window restoring back to previous size.
+            WinImpl::DraggedWindowStartPosition = window->GetClientPosition();
+            WinImpl::DraggedWindowSize = window->GetClientSize();
+            WinImpl::DraggedWindowMousePosition = WinImpl::DraggedWindowStartPosition + WinImpl::DraggedWindowMousePosition - window->GetClientPosition();
```
Tryibion, line 139: Remove the `WinImpl::DraggedWindowStartPosition` and this will work fine.

GoaLitiuM:
Rebased to 1.10 and updated the precompiled SDL3 windows binaries to 3.2.10.

alsed:
Just a minor problem detected (running on linux), the panel to copy/paste the game parameters does not respond:

Captura desde 2025-05-13 08-17-05

GoaLitiuM:
> Just a minor problem detected (running on linux), the panel to copy/paste the game settings does not respond:
> 
I believe this was fixed when copying non-text data was implemented in this branch recently.

GoaLitiuM:
Added libportal library to help with XDG Desktop Portal integration required for picking colors on screen on Wayland, this also needs to be built as an additional dependency along with SDL using the following command:

`Flax.Build -log -BuildDeps -depsToBuild=SDL,libportal -verbose`

ThePhantomMask:
When were the checks carried out?

GoaLitiuM:
Updated to latest master/1.11 and SDL3 3.2.24. Precompiled binaries updated and added for Windows on ARM.

GoaLitiuM:
Updated SDL to 3.4.0 release version, this should be the last major update until final review and merge.

mafiesto4:
I can start merging this but I'd like to target `1.2` branch instead, which will be merged back to `master` when we're closer to release.

mafiesto4:
I started merging but it will take a while to test all the changes (esp. input, dock panel, high-dpi support...)

mafiesto4:
Merged! Thanks for this, you're build that SDL layer in a very smart way int othe existing codebase! ❤️

mafiesto4:
I took the liberty of making some changes:
* 019a9f6089518702314838e4a16726ff1cb2db17 - got rid of `ScreenUtilities` as it's not a big thing and moved it to `Platform` + made it Editor-only
* a3492e59eff91e6933cdc3c7784093ea755b3858 - changed some Windows APIs to use old impl that was tested and used for now
* 8e7dc2a91ea0578a9e349975c41b1a020323453a / d4a7b3074e9be9694c9269e319ecb2ce99274ae2 - bugfix to ensure that when building with SDL disabled it works fine
* 3c5c6f988340e54f451e57864e3da24d2ccbaf67 - fixed dock window hints placement to match the old layout more, made them larger
* c4d20f06ee4fea0079e71e02846d9d914ea62294 / d68969dbe2a462e8085350bc8eff56f5191eb43e - made a way to make platform-specific scripting APIs cleaner to impl without a need to put stuff into base class (used it for some Android/PS4/Xbox/PS5/Switch specific things too that were C++-only until now 😄 )
* 2a6e38e020e28ee4a3f27647a65e2f32105978ac - fixed floating windows on Windows to still link into parent window but exist on tasbar by hacking underlying native window style

mafiesto4:
Feedback for future work:
* Need to build libs for Mac/Linux and ensure all build fine and work
* Every time editor viewport starts controling mouse to move around there is a new SDL Raw Input thread created, probably fine for Editor as in-game it would last longer (eg. only except pause/main menu or when not using mouse lock)
* test for some time and esnrue it works msoothly once we merge to `master` for update.

mafiesto4:
It works quite well on Mac. Docking and popups/tooltips/menus are more functional! But:
* drag&drop is broken (or I was unable to use it with a touchpad)
* cannot dock single floating window by the window title bar (I think some code for movement via caption is windows ot linxu only)
* sometimes camera jumps when using editor viewport and mouse is far from center (we move mouse there but it breaks the rotation), but I was unable to repro it 100%

GoaLitiuM:
macOS Drag and drop was one of the last things I was fixing but forgot to finalize for some reason. Linux Wayland also had some unfinished DnD fixes which also got left behind IIRC.

PR #2801, opened 2024-07-25
Force rebuild on proxy changes if rebuild is true

Tryibion:
This saves an extra step when adding proxies.

I have to use this code right now, other wise when I change back the editor and have an actor selected property references will be cleared.
 ```
Editor.ContentDatabase.AddProxy(new SpawnableJsonAssetProxy<TestingJsonStuff>(Editor.Icons.SwitchSettings128), true);
Editor.ContentDatabase.Rebuild(true);
```

This just simplifies the code to only have to be
`Editor.ContentDatabase.AddProxy(new SpawnableJsonAssetProxy<TestingJsonStuff>(Editor.Icons.SwitchSettings128), true);`

### DIFF
--- a/Source/Editor/Modules/ContentDatabaseModule.cs
+++ b/Source/Editor/Modules/ContentDatabaseModule.cs
@@ -1275,6 +1275,12 @@ namespace FlaxEditor.Modules
                 }
                 _dirtyNodes.Clear();
             }
+
+            // Lazy-rebuilds
+            if (_rebuildFlag)
+            {
+                RebuildInternal();
+            }
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: Source/Editor/Modules/ContentDatabaseModule.cs [hunk 1]
```diff
@@ -1275,10 +1275,6 @@ public override void OnUpdate()
                 }
                 _dirtyNodes.Clear();
             }
-
-            // Lazy-rebuilds
-            if (_rebuildFlag)
```
mafiesto4, line 1280: This still might be important for lazy-rebuilds - esp. if plugin adds multiple proxy types so it triggers only one rebuild on the next update.

Tryibion: I forgot I removed this when I was doing some testing. I will add it back in.

PR #2803, opened 2024-07-26
Editor UI Improvements

Tryibion:
Changes splitter colors, tab colors, toolbar colors, and some spacing.

Before:
image

After:
image

### DIFF
--- a/Source/Editor/CustomEditors/Elements/Container/GroupElement.cs
+++ b/Source/Editor/CustomEditors/Elements/Container/GroupElement.cs
@@ -19,7 +19,7 @@ namespace FlaxEditor.CustomEditors.Elements
             ArrowImageClosed = new SpriteBrush(Style.Current.ArrowRight),
             ArrowImageOpened = new SpriteBrush(Style.Current.ArrowDown),
             EnableDropDownIcon = true,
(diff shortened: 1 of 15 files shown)


xxSeys1:
This looks awesome and I can't wait to use this editor, but one thing that I immediately noticed was that there is a black separator line over every tab group *except* for the "Texture" and "Import" tabs in the image editor of the icon atlas.

Would it be possible to add that? Would make the whole thing look even more better and consistent.

Tryibion:
> This looks awesome and I can't wait to use this editor, but one thing that I immediately noticed was that there is a black separator line over every tab group _except_ for the "Texture" and "Import" tabs in the image editor of the icon atlas.
> 
> Would it be possible to add that? Would make the whole thing look even more better and consistent.

I dont know if it adds much...
image

xxSeys1:
I think it looks good but that's just my opinion

kitsuniru:
image
x_x

Tryibion:
> image x_x

That one should not have the black line above it so, I removed that. As far as the little line, not much I can do about that due to drawing order.

GasimoCodes:
Any updates on this PR?

PR #2807, opened 2024-07-27
Add options menu to json assets

Tryibion:
Adds useful utilities to Json Assets via an options button.
image

A better options image could be used, but this uses what is available...

### DIFF
--- a/Source/Editor/Windows/Assets/JsonAssetWindow.cs
+++ b/Source/Editor/Windows/Assets/JsonAssetWindow.cs
@@ -4,6 +4,7 @@ using System;
 using FlaxEditor.Content;
 using FlaxEditor.CustomEditors;
 using FlaxEditor.GUI;
+using FlaxEditor.GUI.ContextMenu;
 using FlaxEngine;
 using FlaxEngine.GUI;
 
@@ -24,6 +25,7 @@ namespace FlaxEditor.Windows.Assets
         private object _object;
         private bool _isRegisteredForScriptsReload;
         private Label _typeText;
+        private ToolStripButton _optionsButton;
 
         /// <summary>
         /// Gets the instance of the Json asset object that is being edited.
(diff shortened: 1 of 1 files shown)


RomanZhu:
Where is "Paste Asset data"?

Tryibion:
To much work

RomanZhu:
What's the reason for copying then?

Tryibion:
What's the reason for copying or pasting when you can just duplicate the item in the content window?

Tryibion:
Someone else can implement it if they want it.

Source/Editor/Windows/Assets/JsonAssetWindow.cs
```diff
@@ -16,6 +19,53 @@ namespace FlaxEditor.Windows.Assets
     /// <seealso cref="FlaxEditor.Windows.Assets.AssetEditorWindow" />
     public sealed class JsonAssetWindow : AssetEditorWindowBase<JsonAsset>
     {
+        private class ObjectPasteUndo : IUndoAction
+        {
+            /// <inheritdoc />
+            public string ActionString => "Object Paste Undo";
+
+            private JsonAssetWindow _window;
+            private object _oldObject;
```
mafiesto4, line 28: `_oldObject` and `_newObject` should be stored as `string` and use Json serialization to store in memory so it will properly work when code is hot-reloaded in Editor.

Source/Editor/Windows/Assets/JsonAssetWindow.cs
```diff
@@ -139,6 +191,29 @@ protected override void OnAssetLoaded()
...
+                {
+                    buttonTag = type.ContentItem;
+                    break;
+                }
+            }
+            
+            _optionsButton = new ToolStripButton(_toolstrip.ItemsHeight, ref Editor.Icons.Settings12)
+            {
+                AnchorPreset = AnchorPresets.TopRight,
+                Tag = buttonTag,
```
mafiesto4, line 210: `Size = new Float2(18),` would make the that icon look lees pixalated

Source/Editor/Windows/Assets/JsonAssetWindow.cs
```diff
@@ -165,6 +239,62 @@ protected override void OnAssetLoaded()
...
         }
 
+        private void OpenOptionsContextMenu()
+        {
+            if (_optionsCM != null && _optionsCM.ContainsFocus)
+                return;
+            
+            _optionsCM = new ContextMenu();
+            _optionsCM.AddButton("Copy type name", () => Clipboard.Text = Asset.DataTypeName);
+            _optionsCM.AddButton("Copy Asset data", () => Clipboard.Text = Asset.Data);
```
mafiesto4, line 249: Let's change `Asset` to `asset` lowercase to maintain the case in the context menu (I'm aware it could be already made that way in other Editor parts)

mafiesto4 (changes requested):

Source/Editor/Windows/Assets/JsonAssetWindow.cs
Tryibion: Doing this does make it smaller than the other toolstrip icons. I guess because it is off in the corner, it does not look bad being smaller.

mafiesto4 (approved):

PR #2834, opened 2024-08-07
Fix audio state for non-streaming audio

### LINKED ISSUE #1658: IsActuallyPlayingSth should return False if finished
If "Disable Streaming" is enabled on a sound file IsActuallyPlayingSth will always return True even after the audio is finished playing. It will return false if "Disable Streaming" is disabled.

I think it should return false with "Disable Streaming" enabled as well since that would be the expected behaviour. 

image

Tryibion:
Fix #1658 

I am sure there is a better way to fix this, but this worked fine for me.

### DIFF
--- a/Source/Engine/Audio/AudioSource.cpp
+++ b/Source/Engine/Audio/AudioSource.cpp
@@ -432,6 +432,8 @@ void AudioSource::Update()
             Stop();
         }
     }
+
+    // Reset starting to play value
     if (_startingToPlay)
     {
         _startingToPlay = false;
--- a/Source/Engine/Audio/AudioSource.cpp
+++ b/Source/Engine/Audio/AudioSource.cpp
@@ -420,7 +420,15 @@ void AudioSource::Update()
         AudioBackend::Source::VelocityChanged(this);
     }
 
-    if (!UseStreaming() && Math::NearEqual(GetTime(), GetStartTime()) && _isActuallyPlayingSth && !_startingToPlay)
+    // Reset starting to play value once time is greater than zero
+    if (_startingToPlay && GetTime() > 0.0f)
+    {
+        _startingToPlay = false;
+    }
+
+    int32 queuedBuffers;
+    AudioBackend::Source::GetQueuedBuffersCount(this, queuedBuffers);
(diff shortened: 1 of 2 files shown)


artzn:
Thank you for working on the issue, I tried these changes and they do not work repeatedly when I tested them with the project from #1658. Both of these videos were done from a fresh build of a newly opened project each time (remember to unmute).

**Here is the standard run, works first with both sound and reporting, but then does not work afterwards, then randomly works (26s), then does not work:**

https://github.com/user-attachments/assets/748ff59a-f12f-4dcc-aa4d-8d78f2fc131d

**Here is what happens when I adjust the Start Time on the AudioSource to 0.5f. Does not work the first time in the reporting, does not play afterwards, (47s) plays but the reporting does not work:**

https://github.com/user-attachments/assets/36ac7d4a-f8c3-4d1c-97ef-0de8e69da40f

Speculation: When [AudioBackend] finishes playing a clip it always sets its time to 0.0f, irregardless of the start time, the finished clip will report 0.0f so the GetTime sorta== GetStartTime might never be true in some parameters of adjusted AudioSource start time. _Unknown if that is the same behaviour for all AudioBackends_.

Changing line 423 from:
```cpp
if (!UseStreaming() && Math::NearEqual(GetTime(), GetStartTime()) && _isActuallyPlayingSth && !_startingToPlay)
```
to
```cpp
float const backend_may_have_completed_time_set_to_zero = 0.0f;
if (!UseStreaming() && GetTime() == backend_may_have_completed_time_set_to_zero && _isActuallyPlayingSth && !_startingToPlay)
```
Only solves the second case of a start time set to 0.5f, now working in both sound and reporting consistently.
It does not fix all the issues with first case where the start time is at the default of 0.0f: The sound and reporting play correctly on the first run, sometimes the sound does not play, if the sound does play it works in both sound and reporting but I only got it to work by triple clicking the play button so the game starts (not working), stops, starts (plays correctly).

Tryibion:
Yeah it looks like in some instances the Time will remain 0 for multiple update ticks and will cause the code to run stopping the audio... Can you test with the most recent commit?

artzn:
I confirm the sound plays correctly and the reporting is also correct. This fixes the project within the referenced issue.

**Video showing everything working (remember to unmute):**

https://github.com/user-attachments/assets/f32984bf-dba8-4473-922e-8c21a013d1eb

PR #2900, opened 2024-08-30
Fix calling null actor in prefab debug draw.

### LINKED ISSUE #2898: [Crash] Editor crashes when switching tabs
**Issue description:**

Sometimes (haven't found a way to repro it yet at all) the editor crashes when switching to another tab that's docked in the main editor window.

I think this may happen when switching to a tab of an asset that has been open for a longer time but of which the tab I haven't used recently.

I now have this happen every time I tab back into the editor from vscode. The editor crashes when scripts have finished compiling. I think this may be the same issue as described here.

Had this happen maybe 5 times.

Also there does not seem to be a crash folder.

**Steps to reproduce:**

I don't know :(

none

**Flax version:**

latest launcher master as of 8/30/2024
First noticed ~three days ago (~8/27/2024)

Tryibion:
Fix #2898

### DIFF
--- a/Source/Editor/Viewport/PrefabWindowViewport.cs
+++ b/Source/Editor/Viewport/PrefabWindowViewport.cs
@@ -635,6 +635,8 @@ namespace FlaxEditor.Viewport
                 if (child is not ActorNode actorNode)
                     continue;
                 var actor = actorNode.Actor;
+                if (!actor)
+                    continue;
                 if (collectActors)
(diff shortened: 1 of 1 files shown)


Tryibion:
Ah, this has already been fixed in the 1.9 branch. Maybe that fix can be back ported.

mafiesto4:
Backported in 25aa5bee85bebfe934e01ce87e74386f37078714

PR #2929, opened 2024-09-13
Improve output log scroll

xxSeys1:
This pr aims to make improvements to output log scrolling

### TODO
- [X] Implement scroll to bottom on new line
- [X] Add scroll to bottom on enter play
- [x] Fix scroll after resizing output log


xxSeys1:
One problem this pr might be causing is that the output log will now scroll to the bottom for every new entry.
The editor setting was enabled before by default but not implemented. I'd imagine that most people prefer the old behavior without auto scroll because they've gotten used to it by now.

Maybe mention this fact in the release notes to avoid confusion?

Also I was considering to disable this option by default, but I kept it enabled for now cause that's how it has been so far.

PR #2931, opened 2024-09-14
Make it possible to pan editor when rmb down on node

xxSeys1:
Currently you have to start the right click in an empty space in the node editor.

This pr makes it possible to also start right click on a node by checking if the mouse moved.

### DIFF
--- a/Source/Editor/Surface/SurfaceNode.cs
+++ b/Source/Editor/Surface/SurfaceNode.cs
@@ -146,6 +146,8 @@ namespace FlaxEditor.Surface
         /// </summary>
         protected virtual Color FooterColor => GroupArchetype.Color;
 
+        private Float2 mouseDownMousePosition;
+
         /// <summary>
         /// Calculates the size of the node including header, footer, and margins.
         /// </summary>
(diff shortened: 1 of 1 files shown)


xxSeys1:
Dunno why the Linux check failed, seems to have something to do with nuget

mafiesto4:
You could use `Float2 location` instead `Input.Mouse.Position` and apply some safe margin (like `4` pixels) when checking if mouse moved.

xxSeys1:
> You could use `Float2 location` instead `Input.Mouse.Position` and apply some safe margin (like `4` pixels) when checking if mouse moved.

I tried using `Float2 location` but it wasn't working anymore, so I'm back to `Input.Mouse.Position` now.

xxSeys1:
@mafiesto4 Can this be merged for 1.11?

There are a lot new Visject features and fixes, so I think a lot of people who will give Flax a try will try out Visject based tools.

Not being able to pan the editor when starting the mouse drag on a node is a major pain point and feels unpolished in my opinion.
The changes made are also very small. It was also on the 1.10 milestone before.

I have been using my custom branch, with this pr merged in, for about half a year now and I never noticed any issues with it.

PR #2939, opened 2024-09-23
Alternative begin play call fix for scene actor initialization.

Tryibion:
Alternative fix to #2752. Take which ever one you like better.
In the small amount of testing that I did, it seems to work fine.


Tryibion:
Taking a look at this some more. More logic will need to be added around spawning actors.

Tryibion:
Closing in favor of #2940

PR #2940, opened 2024-09-23
Alternative and better fix for calling BeginPlay on actors.

Tryibion:
Better fix than #2939 and #2752 for #2751 and #2623


mafiesto4:
Here is a bigger fix that doesm't allocate dynamic memory and works for other script events that could potentially mess with hierarchy: ca6544204b0bb413083f322ee4a80f077a2141ad

PR #2945, opened 2024-09-25
Fix ARM64 compiler support detection with MSVC v140 toolset

GoaLitiuM:
Possible fix for build tool attempting to generate Windows ARM64 configurations on systems with older MSVC toolsets installed.


WolfInABowl:
Can confirm that this fixed the issue that I was having

PR #2954, opened 2024-09-28
Render transform gizmo on top of everything.

Tryibion:
I am sure there is a better way to do this, but I am not that good at this rendering stuff... so I created a new post process effect only for the transform gizmo.

Before:
image

After:
image


mafiesto4:
I used ordering draw calls to fix this issue instead of 2nd pass. d9ae41ed108ae6c892e55f0b7c4fef29aab75f17

PR #2977, opened 2024-10-09
Add model import option to only create material slots

xxSeys1:
This adds an import option to only create material slots in a model but not import the actual materials or textures.

image

Not quite happy with the name yet, if anyone has a better idea, please let me know.


xxSeys1:
@mafiesto4 Hi,
can this PR be added to the 1.12 milestone?

PR #2979, opened 2024-10-09
Make "Edit Prefab" no longer change content panel file path

xxSeys1:
It's always bothered me that the "Edit Prefab" button changes the directory the Content Panel has open.
My primary use case for that button is to quickly open a prefab, when I have an asset selected in the Content Panel that I want to drag into the prefab.

If someone still wants to change the Content panel to the file path of the prefab, they can use the "Select Prefab" button (or I'll make an editor setting for that if there is demand).


xxSeys1:
@mafiesto4 Can this be merged in the next batch of merges? It's a major pain point of the editor for me atm (I'm on launcher master)

PR #2982, opened 2024-10-11
Coroutines

mtszkarbowiak:
Observations: 

* Enumerator methods may not be used, because they can't yield to C++.
* C++ 20 coroutines may not be used, because they are C++ 20 after all.

This quickly suggest homebrew Flax-specific solution. Unfortunately, Flax API limits require using chain of methods to set up stages of the coroutine. Thus, they are not as sexy as Unity enumerator method-based coroutines, but share the same API in both managed and native scripts. **CRs are welcome!**

> Oh please, and since now on, please stop suggesting people to use TPL. Using other threads just for what are coroutines is ridiculous. (Implementing our own single-threaded scheduler for async tasks does not count.)


mafiesto4 (changes requested):

Source/Engine/Scripting/Coroutines/CoroutineSuspendPoint.h
```diff
@@ -0,0 +1,35 @@
...
+/// </summary>
+/// <remarks>
+/// If loop stages are not present in this enum, the coroutines may not be suspended at these points.
+/// </remarks>
+API_ENUM() enum class CoroutineSuspendPoint : uint8
+{
+    /// <summary>
+    /// Suspension point during or after the OnUpdate event.
+    /// </summary>
+	Update          = 0,
```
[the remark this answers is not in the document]
mtszkarbowiak: Gonna reformat it back!

mafiesto4 (approved):

mtszkarbowiak:
https://github.com/FlaxEngine/FlaxDocs/pull/165

Source/Engine/Scripting/Coroutines/CoroutineSequence.h
```diff
@@ -0,0 +1,165 @@
...
+#include "Engine/Core/Types/Variant.h"
+#include "Engine/Scripting/ScriptingObject.h"
+#include "Engine/Scripting/ScriptingObjectReference.h"
+
+
+/// <summary>
+/// Stores code to be executed once previous step is completed.
+/// </summary>
+/// <remarks>
+/// This class is used due to limitations of Flax API.
```
mafiesto4, line 17: If function takes `Function<void()>` as parameter then it will be passed as C# delegate to be invoked from C++. So this `CoroutineRunnable` is not needed.

Source/Engine/Scripting/Coroutines/CoroutineSequence.h
```diff
@@ -0,0 +1,165 @@
...
+        explicit Step(float secondsDelay);
+        explicit Step(PredicateReference&& predicate);
+
+        Step(const Step& other);
+        Step(Step&& other) noexcept;
+
+    private:
+        void Clear();
+        void EmplaceCopy(const Step& other);
+        void EmplaceMove(Step&& other) noexcept;
```
mafiesto4, line 138: I think memory management for such structure is overenginnering. It can be data-only (maybe with frames delta, seconds delay and suspension point as union). Only `CoroutineSequence` and `CoroutineExecutor` use it.

Source/Engine/Scripting/Coroutines/CoroutineSequence.h
```diff
@@ -0,0 +1,165 @@
...
+    /// <summary> Suspends the coroutine until the predicate is true. </summary>
+    API_FUNCTION()
+    ScriptingObjectReference<CoroutineSequence> ThenWaitUntil(ScriptingObjectReference<CoroutinePredicate> predicate);
+
+
+    /// <summary> QOL wrapper for running runnable action in native scripts. </summary>
+    ScriptingObjectReference<CoroutineSequence> ThenRunFunc(const Function<void()>& runnable);
+
+    /// <summary> QOL wrapper for waiting until predicate in native scripts. </summary>
+    ScriptingObjectReference<CoroutineSequence> ThenWaitUntilFunc(const Function<void(bool&)>& predicate);
```
mafiesto4, line 93: It can be `Function<bool()>` and `API_FUNCTION` so C# can pass `Action` directly.

Source/Engine/Scripting/Coroutines/CoroutineExecutor.cpp
```diff
@@ -0,0 +1,313 @@
...
+    ScriptingObjectReference<CoroutineSequence> sequence,
+    const CoroutineSuspendPoint accumulationPoint
+)
+{
+    const ExecutionID id = _uuidGenerator.Generate();
+    Execution execution{ MoveTemp(sequence), accumulationPoint, id };
+    execution.ContinueCoroutine(CoroutineSuspendPoint::Update, Delta{ 0.0f, 0 });
+    _executions.Add(MoveTemp(execution));
+
+    ScriptingObjectReference<CoroutineHandle> handle = NewObject<CoroutineHandle>();
```
mafiesto4, line 18: `CoroutineHandle` as struct would not leak memory and be faster to use.

mafiesto4 (commented):
Sorry for late review. I've got some feedback. I can do some changes while merging.

Source/Engine/Scripting/Coroutines/CoroutineExecutor.cpp
mtszkarbowiak: As I remember I thought API structs must not have `ScriptableObjectReference`. If it is supported, then I can change it.

Source/Engine/Scripting/Coroutines/CoroutineSequence.h
mtszkarbowiak: Last time I tried it didn't work. Gonna try it again.

PR #2983, opened 2024-10-12
Implement Ground Truth Ambient Occlusion (GTAO)

### LINKED ISSUE #2021: Suggestion: Add GTAO in addition to ASSAO
**Issue description:**
Intel has marked the [ASSAO](https://github.com/GameTechDev/ASSAO) repository that Flax refers to as obsolete and recommended their [GTAO] implementation as a faster and better-looking alternative. 

**Steps to reproduce:**
N/A

N/A

**Flax version:**
1.7.6404

HydrogenC:
Implements GTAO and closes #2021.  
Since I'm a newbie in computer graphics, there might be some potential code defects that may cause artifacts.  
The scene looks overall slightly darker. I suppose this is a characteristic of the GTAO algorithm itself, but perhaps it is caused by code defects in my implementation.  
~~Update: After removing a normalization factor `2/PI`, the result looks as bright as ASSAO. However, Unreal and a couple of other implementations multiply the out `ao` term by this normalization. But here in Flax, the brightness becomes too low compared with ASSAO after multiplying this factor, whereas it looks as bright as ASSAO without this factor. (Maybe it's already handled elsewhere so we don't need it)~~
Update: Unreal engine multiplied the ao value by `2/PI` in the compute pass and later multiplied a `PI/2` in the filter pass. So it's same to not multiplying anything (which my code currently does). 

Updated images (without `2/PI` ), it looks as bright as ASSAO: 

Comp

Note: default parameters (e.g. for multiplier, pow) may have to be further tuned for GTAO to get better results

## Implementation points
 - Temporal filter not included
   Temporal filter is optional for GTAO and it GTAO works well either with or without temporal pass. For example, the [O3DE Draft for GTAO](https://github.com/o3de/sig-graphics-audio/blob/main/rfcs/GTAOIntegration/Proposed%20RFC%20Feature%20GTAO%20(Ground-Truth%20based%20Ambient%20Occlusion).md) also doesn't plan adding temporal pass. 
 - Spatial blurring for ASSAO reused in GTAO
   In the original paper, the implementation of the spatial filter is similar to the existing depth-aware blurring function for ASSAO. So the same blurring pass are used for GTAO as ASSAO.  

## Other code changes
Code changes irrelevant to GTAO implementation are explained below.


Source/Engine/Graphics/PostProcessSettings.cpp
```diff
@@ -17,17 +17,20 @@ void AmbientOcclusionSettings::BlendWith(AmbientOcclusionSettings& other, float
...
     BLEND_FLOAT(FadeDistance);
+    BLEND_FLOAT(GTAORadius);
+    BLEND_FLOAT(GTAOThickness);
 }
 
 void GlobalIlluminationSettings::BlendWith(GlobalIlluminationSettings& other, float weight)
 {
     const bool isHalf = weight >= 0.5f;
-    BLEND_BOOL(Mode);
+    BLEND_ENUM(Mode);
```
HydrogenC, line 33: ENUM should be used instead for `Mode`

Source/Engine/Renderer/AmbientOcclusionPass.cpp
```diff
@@ -348,7 +351,7 @@ void AmbientOcclusionPass::UpdateCB(const RenderContext& renderContext, GPUConte
     _constantsBufferData.Viewport2xPixelSize = Float2(_constantsBufferData.ViewportPixelSize.X * 2.0f, _constantsBufferData.ViewportPixelSize.Y * 2.0f);
     _constantsBufferData.Viewport2xPixelSize_x_025 = Float2(_constantsBufferData.Viewport2xPixelSize.X * 0.25f, _constantsBufferData.Viewport2xPixelSize.Y * 0.25f);
 
-    const float tanHalfFOVY = 1.0f / proj.Values[1][1];
+    const float invTanHalfFOVY = view.IsPerspectiveProjection() ? proj.Values[1][1] : 1.0f;
```
HydrogenC, line 354: Use inv instead for better performance.

Source/Engine/Renderer/AmbientOcclusionPass.cpp
```diff
@@ -269,10 +273,9 @@ void AmbientOcclusionPass::Render(RenderContext& renderContext)
     // Request temporary buffers
     InitRTs(renderContext);
 
-    // Update and bind constant buffer
+    // Update and bind SSAO constant buffer
     UpdateCB(renderContext, context, 0);
-    context->BindCB(SSAO_CONSTANTS_BUFFER_SLOT, _shader->GetShader()->GetCB(SSAO_CONSTANTS_BUFFER_SLOT));
```
HydrogenC, line 274: Remove redundant call: `BindCB` is already called inside `UpdateCB` so this call is redundant.

mafiesto4:
I think this needs some improvements or maybe using code from [XeGTAO](https://github.com/GameTechDev/XeGTAO).

ASSAO:

Screenshot_2025_03_10_22_35_18

GTAO:

Screenshot_2025_03_10_22_35_14

HydrogenC:
It's the blurring pass. The logic of the main pass is the same 

> I think this needs some improvements or maybe using code from [XeGTAO](https://github.com/GameTechDev/XeGTAO).
> 
> ASSAO:
> 
> Screenshot_2025_03_10_22_35_18
> 
> GTAO:
> 
> Screenshot_2025_03_10_22_35_14

This may be partially because that the intensity parameters of these two methods isn't of the same scale. A value appropriate for ASSAO may be too dark for GTAO. 
As for quality issues, It's the blurring pass. The logic of the main pass is the same (if I hadn't wrote anything wrong). I'll look into it when I have time. Meanwhile, I will appreciate it if any contributor with better expertise in computer graphics than me may help.

HydrogenC:
> 这是有道理的，谢谢。我稍后会回到这个 PR，因为我们需要这个功能。

Thanks. The multiplier for the intensity still have to be tested and adjusted, and maybe we shall also improve the blurring a bit (since [XeGTAO](https://github.com/GameTechDev/XeGTAO) uses a more complicated blurring pass than this) to achieve better results.

TheIronLefty:
Nice to see an evolution of ASSAO, I wish to add that while original GTAO paper is really good, there has been iteration of it using solid angle and visibility bitmask to make it even better, ReShade's MXAO from iMMERSE contains these changes for demonstration purposes.

Here's the paper for visbit mask version https://arxiv.org/abs/2301.11376

HydrogenC:
> Nice to see an evolution of ASSAO, I wish to add that while original GTAO paper is really good, there has been iteration of it using solid angle and visibility bitmask to make it even better, ReShade's MXAO from iMMERSE contains these changes for demonstration purposes.
> 
> Here's the paper for visbit mask version: https://arxiv.org/abs/2301.11376 Link for comparision images by Kostas Anagnostou: https://x.com/KostasAAA/status/1637135245739802626

It indeed looks better, but it isn't widely adopted in current generation game/rendering engines, so there are few reference implementations, so I currently don't plan to implement this. However, any contribution or assistance is welcomed!

PR #2988, opened 2024-10-15
Add debug draw clear method

### LINKED ISSUE #2958: Add DebugDraw::Clear function
Currently, there is no dedicated function to clear shapes drawn via `DebugDraw`.
The desired behavior can be achieved via:
```cs
DebugDraw.UpdateContext(IntPtr.Zero, float.MaxValue);
```
```cpp
DebugDraw::UpdateContext(nullptr, MAX_float);
```
However, I wouldn't expect someone to figure this out without looking at the source code.
This is especially true for the C# method, as usage of IntPtr is typically rare for user-facing APIs.

The Clear function would work the same as the above, except omitting the deltaTime parameter.
Another overload would also omit the context parameter
```cpp
DebugDraw::Clear(fooContext);
DebugDraw::Clear();
```
Internally, these would just call `UpdateContext`

xxSeys1:
Adds a method to clear all debug draw on screen to the C# & C++ api.

Closes #2958

### DIFF
--- a/Source/Engine/Debug/DebugDraw.cpp
+++ b/Source/Engine/Debug/DebugDraw.cpp
@@ -2147,6 +2147,6 @@ void DebugDraw::DrawText(const StringView& text, const Transform& transform, con
 
 void DebugDraw::Clear(void* context)
 {
-    DebugDraw::UpdateContext(nullptr, MAX_float);
+    DebugDraw::UpdateContext(context, MAX_float);
 }
 #endif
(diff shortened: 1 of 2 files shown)


WolfInABowl:
There seems to be a mistake in the implementation - the `context` parameter isn't being passed to the `UpdateContext` call. 
```cpp
void DebugDraw::Clear(void* context)
{
    DebugDraw::UpdateContext(nullptr, MAX_float);
}
```

PR #2989, opened 2024-10-15
Add debug draw clear to game and editor panel

xxSeys1:
This pr adds a button to clear all debug draw to the _Game Panel_ rmb menu as well as to the _View_ menu.

Has happened a lot of times that I wanted to get rid of some old debug draw with the click of a button.
Requires #2988 to be merged (which it is).

image
image


mafiesto4:
I added check in 47b22b305d084cf9a25bab563dc365b8378d9d65 to hide this button if no Debug Draw contents is there.

PR #2991, opened 2024-10-15
Add tooltip text alignment editor options

### LINKED ISSUE #2946: Tooltip Justification
**Issue description:**
Enhancement request

It would be useful for the readability of large tooltips if there was an option to left justify them (currently they are centred).

Example:
[Tooltip("A long and winding tool tip, that may have more than one paragraph. \n Starts to look a bit weird with only centre justification. \n It's not really that easy to read.")]

I know this probably sounds like a crazy request, but I find using the tooltips to provide an "always on" documentation system is quite effective and long tips can be useful. 

**Flax version:**
1.8.2

xxSeys1:
If anyone has a better alternative for the `switch` statement, please leave a comment :)

Closes #2946

### DIFF
--- a/Source/Editor/Options/InterfaceOptions.cs
+++ b/Source/Editor/Options/InterfaceOptions.cs
@@ -218,6 +218,15 @@ namespace FlaxEditor.Options
         public bool ShowTreeLines { get; set; } = true;
 
         /// <summary>
+        /// Gets or sets tooltip text alignment.
+        /// </summary>
+        [DefaultValue(TextAlignment.Center)]
+        [EditorDisplay("Interface"), EditorOrder(321)]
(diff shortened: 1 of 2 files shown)


WolfInABowl:
You could use a switch expression to condense it from:
```cs
float textX = 0;
switch (Editor.Instance.Options.Options.Interface.TooltipTextAlignment)
{
    case TextAlignment.Near:
        textX = 15;
        break;
    case TextAlignment.Center:
        textX = 5;
        break;
    case TextAlignment.Far:
        textX = -5;
        break;
}
```
To:
```cs
float textX = Editor.Instance.Options.Options.Interface.TooltipTextAlignment switch
{
    TextAlignment.Near => 15,
    TextAlignment.Center => 5,
    TextAlignment.Far => -5,
};
```

PR #2993, opened 2024-10-16
Add focus Output or Debug Log panel on Begin Play editor option

xxSeys1:
image

Does what the title says.
Will focus the previously focused panel on end play if it's still in the same container than the selected log panel.

### DIFF
--- a/Source/Editor/Options/InterfaceOptions.cs
+++ b/Source/Editor/Options/InterfaceOptions.cs
@@ -138,7 +138,7 @@ namespace FlaxEditor.Options
         }
 
         /// <summary>
-        /// Avaliable options for focusing a log panel on Begin Play.
+        /// Avaliable options for showing a log panel on Begin Play.
         /// </summary>
         public enum LogWindowType
         {
--- a/Source/Editor/Options/InterfaceOptions.cs
+++ b/Source/Editor/Options/InterfaceOptions.cs
@@ -2,7 +2,6 @@
 
 using System.ComponentModel;
 using FlaxEditor.GUI.Docking;
-using FlaxEditor.Utilities;
 using FlaxEngine;
 
 namespace FlaxEditor.Options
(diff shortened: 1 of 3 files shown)


Source/Editor/Options/InterfaceOptions.cs
```diff
@@ -340,6 +340,13 @@ public FontReference OutputLogTextFont
         [EditorDisplay("Output Log", "Scroll to bottom"), EditorOrder(470), Tooltip("Scroll the output log view to bottom automatically after new lines are added.")]
         public bool OutputLogScrollToBottom { get; set; } = true;
 
+        /// <summary>
+        /// Gets or set a value indicating wether the outpug log panel should be focussed on begin play.
+        /// </summary>
+        [DefaultValue(false)]
+        [EditorDisplay("Output Log", "Focus Output Log Panel on Begin Play"), EditorOrder(471), Tooltip("Focus the Output Log panel on Begin Play. Focuses the previously selected panel on End Play.")]
+        public bool FocusOutputLogOnBeginPlay { get; set; } = false;
```
mafiesto4, line 348: I think it would be better to use `Show` instead of `Focus`  as there already is `FocusOnPlayMode` option enum.

Source/Editor/Windows/OutputLogWindow.cs
```diff
@@ -210,12 +212,33 @@ public OutputLogWindow(Editor editor)
...
+                return;
+
+            _previousWindow = this.ParentDockPanel.SelectedTab;
+
+            this.ParentDockPanel.SelectTab(this);
+        }
+
+        private void OnPlayModeEnd()
+        {
+            if (!Editor.Instance.Options.Options.Interface.FocusOutputLogOnBeginPlay || !this.ParentDockPanel.ContainsTab(_previousWindow))
```
mafiesto4, line 236: Add check for `!_previousWindow.IsDisposing` too.

mafiesto4 (changes requested):

Source/Editor/Windows/OutputLogWindow.cs
xxSeys1: Added in https://github.com/FlaxEngine/FlaxEngine/pull/2993/commits/3e50e2cbe20f1e7a9db94b7e6a7247fb54c6bb44 .

Thanks for the review!

xxSeys1: I removed those checks in https://github.com/FlaxEngine/FlaxEngine/pull/2993/commits/1e4fba2a6a077b14d08c8cc40780b413ce58d18e because they broke showing the previous panel.

PR #3003, opened 2024-10-20
Memory enhancements

mtszkarbowiak:


tecnessino:
Afaik flax has own type traits. `Core/Templates.h`

mtszkarbowiak:
I haven't seen an equivalent of integral type and suspected those are compiler-intrinsic. But apparently MSVC implements it by manually listing all types. Gonna add that. 

https://github.com/microsoft/STL/blob/e0ad9c36380c2fa78042ac556eb8c2100fe2d22e/stl/inc/xtr1common#L188

mtszkarbowiak:
@mafiesto4 I have couple of questions:

1. Fancy allocators require context. This leads to: a) using heavy-templated code and manual context injection b) using only static context. I wanted to make sure you don't have anything about the a. For now I have `BumpFastAlloc` without thread safety mechanisms, but I see `BumpSafeAlloc` for many threads and shared global `FrameAlloc`. Notably, context is also necessary for pooling allocators, etc.

2. Allocators API is a little bit inconsistent on one thing: It allocates, deallocates and relocates, including non-trivial relocation with reconstruction (moves). Maybe I should move construction and destruction into allocation too? It would become much more consistent.

3. I personally would love to see an equivalent of `std::unique_ptr`. To make it simpler, I'd add nullability trait to allocators, if you don't mind.

3. Do you find naming correct? Some suggestions?

mtszkarbowiak:
I have issues with separation of concerns. I need to make some radical changes here.

Source/Engine/Core/Memory/BumpFastAllocation.h
```diff
@@ -0,0 +1,256 @@
...
+#include "Engine/Core/Collections/HashSet.h"
+#endif
+
+/// <summary>
+/// Memory allocation policy with the fastest possible allocation and deallocation.
+/// It uses a simple bump allocator that allocates memory in a linear fashion with alignment support.
+/// It does not synchronize access to the memory block, so it is not thread-safe.
+/// Context must be reset manually between uses.
+/// </summary>
+class BumpFastAllocation
```
mafiesto4, line 22: I think `BumpAllocation` name would make more sense.

Source/Engine/Core/Memory/BumpFastAllocation.h
```diff
@@ -0,0 +1,256 @@
+// Copyright (c) 2012-2024 Wojciech Figat. All rights reserved.
+
+#pragma once
+
+#include "Allocation.h"
+#include "Engine/Core/Log.h"
+#include "Engine/Core/Collections/BitArray.h"
```
mafiesto4, line 7: Most likely not needed header here.

Source/Engine/Core/Memory/BumpFastAllocation.h
```diff
@@ -0,0 +1,256 @@
+// Copyright (c) 2012-2024 Wojciech Figat. All rights reserved.
+
+#pragma once
+
+#include "Allocation.h"
+#include "Engine/Core/Log.h"
```
mafiesto4, line 6: Include `Log.h` below within `#if ASSERT_SAFE_BUMP_ALLOC` so header is ligher for compilation.

Source/Engine/Core/UPtr.h
```diff
@@ -0,0 +1,158 @@
...
+
+/// <summary>
+/// Uniquely owning pointer that manages the lifetime of an object.
+/// </summary>
+/// <remarks>
+/// This class is similar to std::unique_ptr but tailored for the engine's memory management system.
+/// Short name is used to make it easier to type and read.
+/// </remarks>
+template<typename T, typename Alloc = HeapAllocation>
+class UPtr
```
mafiesto4, line 15: I think `UniquePtr` name would make more sense here.

mafiesto4 (changes requested):

mafiesto4:
@mtszkarbowiak  Reply to those questions:

1) I like they way you went here with that context. Sometime swe want static/global context (eg. `RendererAllocation`) where whole code shares the same allocation, while sometimes it makes sense to have per-function or per-tick allocations sharing.

2) Allocators API was meant to be simplistic so it only took care of memory but it might make some sense to support ability to handle actual value data operations (like move). We already kid of inform allocator about data type (eg. `FixedAllocation::Data<T>` where `T` is value type). So if you think it would be good to move more logic there then I'm fine. Maybe adding base impl for allocator data class that would define default logic could reduce code size (esp. impl move op for every new allocator).

3) Engine was built without smart pointers with more low-level approach where raw pointers should be handled properly. But of course it's a good thing to app shared/unqiue pointers to codebase for users to access. Even if engine would not utilize them that much.

Source/Engine/Core/Memory/BumpFastAllocation.h
mtszkarbowiak: I wanted to have 2 bump allocators: with and without safe threading. Thus two of them: `BumpFactAlloc` and `BumpSafeAlloc`.

mtszkarbowiak:
@mafiesto4 I want a permission to make a massive API change. I was refactoring the Dictionary to use the new type-unaware allocations (so allocations do allocations only) but then I realized the bucket has massive issue - It is possible to change bucket key after it was added to dictionary and that ruins data integrity of entire collection. The key is retrieved by a field directly in many places in code, so the change is 'very breaking'. Can I change the API?

mafiesto4:
@mtszkarbowiak yeah key should be `const` so I'm fine with this breaking change (it will be merged to `1.10` branch first)

Source/Engine/Core/Collections/Array.h
```diff
@@ -323,7 +312,7 @@ API_CLASS(InBuild) class Array
     FORCE_INLINE const T& operator[](int32 index) const
     {
         ASSERT(index >= 0 && index < _count);
-        return _allocation.Get()[index];
+        return Get()[index];
```
mafiesto4, line 315: I'm aware that changing `_allocation.Get()` into `Get()` makes code cleaner and more readable but it comes at cost of performance - esp. in Debug builds where MSVC compiler doesn't inline that call.

mtszkarbowiak: Actually, I'm going to reverse a lot of those changes. I guess including this one. I got a little bit of tunnel vision trying to get rid of the type template argument from allocation data. On the positive note, now I understand the internals of the collections significantly better. I'm definitely splitting the PR into more digestible portions both for me, cr'ers and the users.

mtszkarbowiak:
I wanted to accomplished unmanageable amount of things here. I'm closing it.

PR #3007, opened 2024-10-22
Add default actor names

Tryibion:


mafiesto4:
Was it requested for gameplay code? This adds significant memory allocations count when loading game thus we have editor-side of naming spawned things.

Tryibion:
Seems like creating new actors from the c++ side does not give them a name. 

So something like this:
```
StaticModel* model = New<StaticModel>();
model.SetParent(GetActor());
```
produces an actor with no name.

Issue #2237, opened 2024-02-12
[Suggestion] Tokenizer error messages improvements


mtszkarbowiak:
The suggestion is to improve documentation and in the future error messages of the following scenarios:

- [X] Field is initialized universally/etc.: `API_FIELD() int32{ 3.2f };`, `API_FIELD() int32 = int32{ 3.2f }`, ...
- [X] Method uses trailing return type: `API_FUNC() auto DoSth(int32) -> int32`
- [ ] Method argument name is not given: `API_FUNC() int32 DoSth(int32)`

PR #3008, opened 2024-10-22
Handle C++ curly brace field default values for code gen to C#

Tryibion:
Handles the code gen for field default values if they are not an array and have curly braces.

```
API_FIELD() float TestFloat = {};
API_FIELD() float TestFloat = { };
API_FIELD() float TestFloat = {1.0f};

API_FIELD() float TestFloat = float{};
API_FIELD() float TestFloat = float{ };
API_FIELD() float TestFloat = float{1.0f};

API_FIELD() float TestFloat {};
API_FIELD() float TestFloat {1.0f};
```

Part of #2237

### DIFF
--- a/Source/Engine/Tests/TestScripting.h
+++ b/Source/Engine/Tests/TestScripting.h
@@ -8,6 +8,35 @@
 #include "Engine/Scripting/ScriptingObject.h"
 #include "Engine/Scripting/SerializableScriptingObject.h"
 
+// Test default values init on fields.
+API_STRUCT(NoDefault) struct TestDefaultValues
+{
+    DECLARE_SCRIPTING_TYPE_MINIMAL(TestDefaultValues);
+
+    // Default value case 1
+    API_FIELD() float TestFloat1 = {};
+    // Default value case 2
+    API_FIELD() float TestFloat2 = { };
+    // Default value case 3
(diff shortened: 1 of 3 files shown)


mafiesto4:
Cool! Could you add this sample code to `TestScripting.h` so we can have an automated test for this feature?

```cpp
// Test default values init on fields.
API_STRUCT(NoDefault) struct TestDefaultValues
{
    DECLARE_SCRIPTING_TYPE_MINIMAL(TestDefaultValues);

    // Default value case 1
    API_FIELD() float TestFloat1 = {};
    // Default value case 2
    API_FIELD() float TestFloat2 = { };
    // Default value case 3
    API_FIELD() float TestFloat3 = {1.0f};
    // Default value case 4
    API_FIELD() float TestFloat4 = float{};
    // Default value case 5
    API_FIELD() float TestFloat5 = float{ };
    // Default value case 6
    API_FIELD() float TestFloat6 = float{1.0f};
    // Default value case 7
    API_FIELD() float TestFloat7 {};
    // Default value case 8
    API_FIELD() float TestFloat8 {1.0f};
};
```

Tryibion:
Should we add standard cases to this test as well?

mafiesto4:
I don't think it's necessary but you can if you want. Lots of cases are already covered in engine code.

PR #3010, opened 2024-10-22
Add warning when used param is being deleted

### LINKED ISSUE #3002: Add alert when is deleting parameter from visject
**Issue description:**
Sometimes i want to delete unused parameters from visject but if i have a complex graph it's possible delete accidentally important parameters breaking the system. Pls, add an alert (Like a popup) if the parameter is used for some node to avoid mistakes.
image

**Flax version:**
1.9

xxSeys1:
image

Closes #3002

"Used" in this case means in graph and connected.

TODO:
- [x] Editor setting to disable prompt
- [x] Test with other graphs than material


Source/Editor/Surface/VisjectSurfaceWindow.cs
```diff
@@ -775,7 +775,34 @@ private void EditAttributesParameter(int index, Control label)
...
+
+            int connectedParameterNodeCount = 0;
+
+            List<SurfaceNode> nodes = window.VisjectSurface.Nodes;
+
+            for (int i = 0; i < window.VisjectSurface.Nodes.Count; i++)
+            {
+                if (nodes[i] is IParametersDependantNode node)
+                {
+                    if (displayWarning && (Guid)nodes[i].Values[0] == param.ID && nodes[i].GetBoxes().Any(b => b.Connections.Count > 0))
```
mafiesto4, line 791: I think it would be better to add method like `bool IsParamReferenced(SurfaceParameter param)` to `IParametersDependantNode ` so we could have more generic way of checking this parameter reference.

mafiesto4 (changes requested):

Source/Editor/Surface/VisjectSurfaceWindow.cs
xxSeys1: Implemented that here: https://github.com/FlaxEngine/FlaxEngine/pull/3010/commits/57b4e9f2955b21d93239d718bc6e8fcaaf811906

Source/Editor/Surface/VisjectSurfaceWindow.cs
```diff
@@ -775,7 +775,34 @@ private void EditAttributesParameter(int index, Control label)
...
+            bool displayWarning = Editor.Instance.Options.Options.Interface.WarnOnDeletingUsedVisjectParameter;
+
             var window = (IVisjectSurfaceWindow)Values[0];
+            SurfaceParameter param = window.VisjectSurface.Parameters[index];
+
+            int connectedParameterNodeCount = 0;
+
+            List<SurfaceNode> nodes = window.VisjectSurface.Nodes;
+
+            for (int i = 0; i < window.VisjectSurface.Nodes.Count; i++)
```
mafiesto4, line 787: I would simplify this and just make `bool isParamUsed = window.VisjectSurface.IsParamreferenced(param )`, then ignore plural version of message.

Source/Editor/Surface/IParametersDependantNode.cs
```diff
@@ -33,5 +33,13 @@ public interface IParametersDependantNode
...
         /// <param name="param">The parameter.</param>
         void OnParamDeleted(SurfaceParameter param);
+
+        /// <summary>
+        /// Get if the parameter is referenced in a graph. Referenced in this case means in a graph and at least one node in-/output connected to another node.
+        /// </summary>
+        /// <param name="param">The parameter.</param>
+        /// <param name="surface">The visject surface.</param>
+        /// <returns>If the parameter is referenced.</returns>
+        bool IsParamreferenced(SurfaceParameter param, VisjectSurface surface);
```
mafiesto4, line 43: Drop `VisjectSurface surface` parameter as node itself knows about internals and can detect whether uses that param.

Source/Editor/Surface/Archetypes/Parameters.cs
```diff
@@ -425,6 +426,20 @@ public void OnParamDeleted(SurfaceParameter param)
                 UpdateCombo();
             }
 
+            /// <inheritdoc />
+            public bool IsParamreferenced(SurfaceParameter param, VisjectSurface surface)
+            {
+                for (int i = 0; i < surface.Nodes.Count; i++)
```
mafiesto4, line 432: Simplify this code into `return (Guid)Values[0] == param.ID;`

Source/Editor/Surface/Archetypes/Parameters.cs
```diff
@@ -937,13 +952,25 @@ public void OnParamDeleted(SurfaceParameter param)
...
                     _combobox.SelectedIndex = -1;
-                }
 
                 UpdateCombo();
             }
 
+            /// <inheritdoc />
+            public bool IsParamreferenced(SurfaceParameter param, VisjectSurface surface)
+            {
+                for (int i = 0; i < surface.Nodes.Count; i++)
```
mafiesto4, line 963: Simplify this code into `return (Guid)Values[0] == param.ID;`.

Source/Editor/Surface/VisjectSurfaceWindow.cs
mafiesto4: Or even better would be to use `VisjectSurfaceContext` to check any nested graphs: eg. if Anim Graph has state machine and that param is used in nested graph but I can take care of that - it will require a little bit more refactor.

mafiesto4 (changes requested):

Source/Editor/Surface/Archetypes/Parameters.cs
xxSeys1: Added in https://github.com/FlaxEngine/FlaxEngine/pull/3010/commits/b055117144d42df15435a98207e41299161185c6. Thanks for the review

Source/Editor/Surface/Archetypes/Parameters.cs
xxSeys1: Added in https://github.com/FlaxEngine/FlaxEngine/pull/3010/commits/b055117144d42df15435a98207e41299161185c6

Source/Editor/Surface/IParametersDependantNode.cs
xxSeys1: Added in https://github.com/FlaxEngine/FlaxEngine/pull/3010/commits/b055117144d42df15435a98207e41299161185c6

Source/Editor/Surface/VisjectSurfaceWindow.cs
xxSeys1: It'd be very cool if you could implement that.

About dropping the plural version:
My thought was, that if a parameter is being used multiple times in a graph but only one *Get {ParamName}* node is visible, the user may not know that the parameter is being used in an important way a second time.

So then the "being used x times" makes them go "Ohh, wait, I thought this parameter is only being used in the place(s) I'm currently seeing, I should check the other use case(s)!":

*Used2 can be seen one time (red box), but is actually being used a second time (red arrow), but the user can't see that. Granted, this is a very simple node graph, but imagine a more complex one with a lot of nodes*
image

Source/Editor/Surface/Archetypes/Parameters.cs
```diff
@@ -425,6 +426,20 @@ public void OnParamDeleted(SurfaceParameter param)
                 UpdateCombo();
             }
 
+            /// <inheritdoc />
+            public bool IsParamreferenced(SurfaceParameter param)
+            {
+                for (int i = 0; i < Surface.Nodes.Count; i++)
```
mafiesto4, line 432: This loop can be removed and whole function body can be:
```cs
return (Guid)Values[0] == param.ID;
```

Source/Editor/Surface/Archetypes/Parameters.cs
```diff
@@ -937,13 +952,25 @@ public void OnParamDeleted(SurfaceParameter param)
...
                     _combobox.SelectedIndex = -1;
-                }
 
                 UpdateCombo();
             }
 
+            /// <inheritdoc />
+            public bool IsParamreferenced(SurfaceParameter param)
+            {
+                for (int i = 0; i < Surface.Nodes.Count; i++)
```
mafiesto4, line 963: This loop can be removed and whole function body can be:
```cs
return (Guid)Values[0] == param.ID;
```

mafiesto4 (changes requested):

Source/Editor/Surface/Archetypes/Parameters.cs
xxSeys1: Added in https://github.com/FlaxEngine/FlaxEngine/pull/3010/commits/38d1e25604cc260fb2f60227a0bae90b43cc30ca

Source/Editor/Surface/Archetypes/Parameters.cs
xxSeys1: Added in https://github.com/FlaxEngine/FlaxEngine/pull/3010/commits/38d1e25604cc260fb2f60227a0bae90b43cc30ca

mafiesto4 (approved):

PR #3021, opened 2024-10-25
Add named terrain layers

xxSeys1:
Adds the ability to assign names to terrain layers via a new section *Terrain Layers* in *Layers And Tags Settings*.
Uses a custom editor (`TerrainLayerEditor`) to display the names.
If there is no name assigned in the settings, the layer is displayed as *"Layer" + layer index* (f.e. "Layer 1").

It'd be cool if someone could test this a bit. I did some testing, but it was so much more easy to implement than I expected, I'm worried I missed something.

I'd like to receive feedback on `LayersAndTagsSettings.GetCurrentTerrainLayers()`, especially about the part where I load the settings.

image
image
image

### DIFF
--- a/Source/Editor/Surface/Archetypes/Material.cs
+++ b/Source/Editor/Surface/Archetypes/Material.cs
@@ -1,6 +1,7 @@
 // Copyright (c) 2012-2024 Wojciech Figat. All rights reserved.
 
 using System;
+using FlaxEditor.Content.Settings;
 using FlaxEditor.Scripting;
 using FlaxEditor.Surface.Elements;
 using FlaxEditor.Windows.Assets;
@@ -590,7 +591,7 @@ namespace FlaxEditor.Surface.Archetypes
                 },
                 Elements = new[]
                 {
-                    NodeElementArchetype.Factory.ComboBox(0, 0, 70.0f, 0, FlaxEditor.Tools.Terrain.PaintTerrainGizmoMode.TerrainLayerNames),
+                    NodeElementArchetype.Factory.ComboBox(0, 0, 70.0f, 0, LayersAndTagsSettings.GetCurrentTerrainLayers()),
                     NodeElementArchetype.Factory.Output(0, "", typeof(float), 0),
                 }
             },
--- a/Source/Engine/Core/Config/LayersAndTagsSettings.cs
+++ b/Source/Engine/Core/Config/LayersAndTagsSettings.cs
@@ -1,6 +1,7 @@
 // Copyright (c) 2012-2024 Wojciech Figat. All rights reserved.
 
 using System.Collections.Generic;
+using System.Linq;
 using System.Runtime.InteropServices;
 using System.Runtime.InteropServices.Marshalling;
 using FlaxEngine;
@@ -28,6 +29,29 @@ namespace FlaxEditor.Content.Settings
         public static string[] GetCurrentLayers()
         {
             return GetCurrentLayers(out int _);
+        }
+
+        /// <summary>
+        /// The layers names.
+        /// </summary>
+        [EditorOrder(10), EditorDisplay("Terrain Layers", EditorDisplayAttribute.InlineStyle), Collection(CanResize = false, Display = CollectionAttribute.DisplayType.Inline)]
+        public string[] TerrainLayers = new string[8];
+
+        /// <summary>
+        /// Gets the current terrain layer names. Returns "Layer" + index for layers without a name.
+        /// </summary>
+        /// <returns>The layer names.</returns>
+        public static string[] GetCurrentTerrainLayers()
+        {
+            string[] layerNames = GameSettings.Load<LayersAndTagsSettings>().TerrainLayers.ToArray();
+
+            for (int i = 0; i < layerNames.Length; i++)
+            {
+                if (string.IsNullOrEmpty(layerNames[i]))
+                    layerNames[i] = $"Terrain Layer {i}";
+            }
+
+            return layerNames;
         }
 
         [LibraryImport("FlaxEngine", EntryPoint = "LayersAndTagsSettingsInternal_GetCurrentLayers", StringMarshalling = StringMarshalling.Custom, StringMarshallingCustomType = typeof(FlaxEngine.Interop.StringMarshaller))]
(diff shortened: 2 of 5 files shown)


Source/Engine/Core/Config/LayersAndTagsSettings.cs
```diff
@@ -30,6 +31,29 @@ public static string[] GetCurrentLayers()
...
+        [EditorOrder(10), EditorDisplay("Terrain Layers", EditorDisplayAttribute.InlineStyle), Collection(CanResize = false, Display = CollectionAttribute.DisplayType.Inline)]
+        public string[] TerrainLayers = new string[8];
+
+        /// <summary>
+        /// Gets the current terrain layer names. Returns "Layer" + index for layers without a name.
+        /// </summary>
+        /// <returns>The layer names.</returns>
+        public static string[] GetCurrentTerrainLayers()
+        {
+            string[] layerNames = GameSettings.Load<LayersAndTagsSettings>().TerrainLayers.ToArray();
```
WolfInABowl, line 46: I think the ToArray() call here isn't needed as `TerrainLayers` is already a string array

Source/Engine/Core/Config/LayersAndTagsSettings.cs
```diff
@@ -30,6 +31,29 @@ public static string[] GetCurrentLayers()
             return GetCurrentLayers(out int _);
         }
 
+        /// <summary>
+        /// The layers names.
+        /// </summary>
+        [EditorOrder(10), EditorDisplay("Terrain Layers", EditorDisplayAttribute.InlineStyle), Collection(CanResize = false, Display = CollectionAttribute.DisplayType.Inline)]
+        public string[] TerrainLayers = new string[8];
```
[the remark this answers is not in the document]
xxSeys1: Added in https://github.com/FlaxEngine/FlaxEngine/pull/3021/commits/1d7054265d1e5702cc1d2007eb6647925f2ca838.

Source/Editor/Surface/Archetypes/Material.cs
```diff
@@ -590,7 +591,7 @@ internal enum MaterialTemplateInputsMapping
                 },
                 Elements = new[]
                 {
-                    NodeElementArchetype.Factory.ComboBox(0, 0, 70.0f, 0, FlaxEditor.Tools.Terrain.PaintTerrainGizmoMode.TerrainLayerNames),
+                    NodeElementArchetype.Factory.ComboBox(0, 0, 70.0f, 0, LayersAndTagsSettings.GetCurrentTerrainLayers()),
```
mafiesto4, line 594: Then we can remove `FlaxEditor.Tools.Terrain.PaintTerrainGizmoMode.TerrainLayerNames` if it won't be needed anymore.

mafiesto4 (changes requested):

Source/Engine/Core/Config/LayersAndTagsSettings.cs
```diff
@@ -30,6 +31,29 @@ public static string[] GetCurrentLayers()
...
+        [EditorOrder(10), EditorDisplay("Terrain Layers", EditorDisplayAttribute.InlineStyle), Collection(CanResize = false, Display = CollectionAttribute.DisplayType.Inline)]
+        public string[] TerrainLayers = Enumerable.Repeat(string.Empty, 8).ToArray();
+
+        /// <summary>
+        /// Gets the current terrain layer names. Returns "Layer" + index for layers without a name.
+        /// </summary>
+        /// <returns>The layer names.</returns>
+        public static string[] GetCurrentTerrainLayers()
+        {
+            string[] layerNames = GameSettings.Load<LayersAndTagsSettings>().TerrainLayers.ToArray();
```
mafiesto4, line 46: Use `FLAX_TESTS` macro to return empty array so Tests will pass (some C# tests run without C++ engine backend thus they fail on compilation).

Source/Editor/Surface/Archetypes/Material.cs
xxSeys1: Done in https://github.com/FlaxEngine/FlaxEngine/pull/3021/commits/c8146a8e99d02f5c1e0bda2023d62189f6c05f07

Source/Engine/Core/Config/LayersAndTagsSettings.cs
```diff
@@ -30,6 +31,29 @@ public static string[] GetCurrentLayers()
             return GetCurrentLayers(out int _);
         }
 
+        /// <summary>
+        /// The layers names.
+        /// </summary>
+        [EditorOrder(10), EditorDisplay("Terrain Layers", EditorDisplayAttribute.InlineStyle), Collection(CanResize = false, Display = CollectionAttribute.DisplayType.Inline)]
+        public string[] TerrainLayers = Enumerable.Repeat(string.Empty, 8).ToArray();
```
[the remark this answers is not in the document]
xxSeys1: Done in https://github.com/FlaxEngine/FlaxEngine/pull/3021/commits/75d4a110de45b5494fd7ee3c1f49597797dd5605.

I also had to remove the `ToArray()`, which leads to the Terrain layer names being populated with `Layer {index}` by default:
image
Before it looked like this:
image

The new behavior might actually make more sense because there will always be 8 terrain layers, so they all need names.

However if that's a problem I can try to find a solution.

Source/Engine/Core/Config/LayersAndTagsSettings.cs
xxSeys1: Done in https://github.com/FlaxEngine/FlaxEngine/commit/75d4a110de45b5494fd7ee3c1f49597797dd5605

Source/Engine/Core/Config/LayersAndTagsSettings.cs
```diff
@@ -30,6 +30,33 @@ public static string[] GetCurrentLayers()
...
+        public string[] TerrainLayers = new string[8];
+
+        /// <summary>
+        /// Gets the current terrain layer names. Returns "Layer" + index for layers without a name.
+        /// </summary>
+        /// <returns>The layer names.</returns>
+        public static string[] GetCurrentTerrainLayers()
+        {
+    #if FLAX_TESTS
+            return Array.Empty<string>();
```
mafiesto4, line 46: Missing `System.Array` to fix compilation (or add using on top of the file).

xxSeys1: Added in https://github.com/FlaxEngine/FlaxEngine/pull/3021/commits/573b57ed4a512246bcfe2a47fa45511b7e0442ff

mafiesto4 (approved):

PR #3030, opened 2024-10-30
Collections maintenance

mtszkarbowiak:
Each change is trivial. Fixing them ad-hoc negatively impacts my focus xD

### DIFF
--- a/Source/Engine/Core/Collections/Array.h
+++ b/Source/Engine/Core/Collections/Array.h
@@ -17,8 +17,8 @@ API_CLASS(InBuild) class Array
 {
     friend Array;
 public:
-    typedef T ItemType;
-    typedef typename AllocationType::template Data<T> AllocationData;
+    using ItemType = T;
+    using AllocationData = typename AllocationType::template Data<T>;
 
 private:
     int32 _count;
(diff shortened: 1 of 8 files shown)


mtszkarbowiak:
This PR is supposed to be trivial. I'm going to use this branch to make more meaningful changes.

mtszkarbowiak:
One more thing: Is there any specific predicate defining which methods are `FORCE_INLINE`d? Seems quite random to me.

mtszkarbowiak:
@mafiesto4 Could you take a look on this PR?

mtszkarbowiak:
I was thinking about inlining in debug builds the alloctor get function... Wouldn't it be beneficial to get rid of `IsEmpty`, `IsDeleted` and other bucket methods?

mafiesto4:
> This PR is supposed to be trivial. I'm going to use this branch to make more meaningful changes.

Sure, good approach. I'll merge this one in master as it doesn't change external APIs that much.

> Is there any specific predicate defining which methods are FORCE_INLINEd? Seems quite random to me.

Usually, things that are small or could benefit from inlining (eg. compiler could improve memory access if code is inlined so data could stay in registers)

> I was thinking about inlining in debug builds the alloctor get function... Wouldn't it be beneficial to get rid of IsEmpty, IsDeleted and other bucket methods?

If it won't produce that much overhead code it might be worth it. Especially in perf-intense methods like `FindPosition` or `SetCapacity`

mtszkarbowiak:
Thank you for merging this one. I'm on my way to deliver the rest.

PR #3032, opened 2024-10-31
Dictionary integrity protection

mtszkarbowiak:
Currently `Dictionary` uses buckets to expose data. This means that the key can be changed, potentially ruining the state of the collection.

**This is an API breaking change.**

`HashSet` is to be revised in the future.

### DIFF
--- a/Source/Editor/Cooker/Steps/CookAssetsStep.cpp
+++ b/Source/Editor/Cooker/Steps/CookAssetsStep.cpp
@@ -120,7 +120,7 @@ void CookAssetsStep::CacheData::InvalidateCachePerType(const StringView& typeNam
     LOG(Info, "Invalidating cooker cache for {0} assets.", typeName);
     for (auto e = Entries.Begin(); e.IsNotEnd(); ++e)
     {
-        if (e->Value.TypeName == typeName)
+        if (e->Value().TypeName == typeName)
         {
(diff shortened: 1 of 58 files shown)


mtszkarbowiak:
Can someone help me with changes in the bindings generator?

mafiesto4:
> Can someone help me with changes in the bindings generator?

You can take a look into generated code here: `<repo>\Cache\Intermediate\FlaxEditor\Windows\x64\Development\Streaming\Streaming.Bindings.Gen.cpp`
C++ codegen is generated by `Source\Tools\Flax.Build\Bindings\BindingsGenerator.Cpp.cs`. In this case probably `ManagedDictionary.h` needs to be updated and codegen inside `GenerateCpp` function that sets `Variant` from input dictionary data.

mafiesto4:
This refactor introduced very large change and potential perf overhead (such as assert on bucket state and function call). Thus I have 2 ideas that maybe could solve this problem but with different approache:
Change `Key` to be `const` and use `const_cast` in `Dictionary` internals to setup/cleanup value.

I think more important would be addition of `ConstIterator` for collections that would provide read-only access to both key and value. That would help with `const-correctness` of foreach loops over dictionaries.

What do you think?

mtszkarbowiak:
I like the idea a lot. I will try it.

PR #3043, opened 2024-11-04
New allocations model

mtszkarbowiak:
# Issues

This PR features experimental changes to Flax allocation policy model. Unfortunately, previous one suffered issues:

1. **Calculating capacity growth and allocating is not synchronized.** This means that current model can not work in multithreaded environment. If internal state of context injected to data changes between calls it could result in undefined state. Different solution could be assuming that `CalculateCapacityGrowth` is pure, but this means it could be moved completely outside of the Data class.
2. **Allocation policies should not govern lifecycle of the objects.** This problem is more subtle: The issue is that using allocators to govern lifecycles means that both collection and allocator must share information about validity of stored object. For trivial situations, where objects are valid at [0, n), the information can be passed with *count* of objects. For any more complicated scenario this method fails.

# Decisions

1. **Calculate capacity grow on allocation.** Now collection itself is responsible for tracking the capacity - the memory given by the allocation. Thanks to this solution, allocator itself can determine the state of the context and give a better result.
2. **Add constraints on allocators: `MinCapacity` and `MaxCapacity`.** This will help the collection automatically select the capacity of the collection.
3. **Remove ability to manually select capacity of a collection, and replace it with `ClearToFree` and `ShrinkToFit` (with old `EnsureCapacity`).** So collections themselves are given more control over the capacity of the collections.
4. **Implement set of shared operations on collections to execute typical operations.** For example, type-agnostic transferring objects between allocations. Not only, it allows to erase type from the allocation policy itself, but also allows to use `Platform::Copy` for trivially/bitwise copyable types. This method also allows to automatically select the method of transferring allocations data (e.g. pointer swap).

# Tasks

- [x] Remove problematic methods from allocation: `CalculateCapacityGrow`, `Relocate`.
- [x] Implement set of operations on linear allocations.
- [ ] Fix collections to use shared allocation operations.
- [ ] Complete Allocation type erasure.
- [ ] Implement new allocation policies.

# Final Note

Stabilizing this branch will take me weeks, I would be super glad if I was given feedback frequently.

### DIFF
--- a/Source/Engine/ContentExporters/AssetsExportingManager.cpp
+++ b/Source/Engine/ContentExporters/AssetsExportingManager.cpp
@@ -128,8 +128,7 @@ bool AssetsExportingManagerService::Init()
 void AssetsExportingManagerService::Dispose()
 {
     // Cleanup
-    AssetsExportingManager::Exporters.Clear();
-    AssetsExportingManager::Exporters.SetCapacity(0);
+    AssetsExportingManager::Exporters.ClearToFree();
 }
 
 #endif
--- a/Source/Engine/Core/Collections/Array.h
+++ b/Source/Engine/Core/Collections/Array.h
@@ -398,10 +398,7 @@ public:
     /// <summary>
     /// Tries to reduce the capacity of the collection to fit its current size.
     /// </summary>
-    /// <remarks>
-    /// Equivalent of <c>std::vector::shrink_to_fit</c>.
-    /// </remarks>
-    FORCE_INLINE void Compact()
+    FORCE_INLINE void ShrinkToFit()
     {
         if (_count == 0)
         {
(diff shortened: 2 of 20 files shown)


mafiesto4:
I ported a significant amount of changes in f5280eab740b1a0a58fb276c457bac8d5548c36c
Additionally, now Dictionary and HahsSet share more code via a base class: 38f74c8cf73d3fbfe89714c44064dc775bb157f9

mtszkarbowiak:
@mafiesto4 Significant improvements! I'm very glad you decided to revise that part.

As anyone can follow, I decided to experiment in an isolated environment of my own [lib](https://github.com/mtszkarbowiak/ktl/). I'm having successes, but the ideas I tried to introduce in this particular PR give limited advantages. When collections support type-erased allocators, adding new allocators is very simple, but getting to that point takes significant effort. I'm definitely going to make a longer post at some time, as I have a lot of thoughts.

For now, I'm giving up on this branch. Too much stuff at once.

Maybe I'll give some time to Flax in weeks to come. I have a game to finish xD

Issue #3061, opened 2024-11-16
Add more script template options


MrCapy0:
**Issue description:**

Pls, add more cm options like create custom json asset or editor window
![image](https://github.com/user-attachments/assets/215a1369-ef32-402c-aa89-47ea52b4fcff)

**Flax version:**
1.9 master

xxSeys1:
Also one for for an interface

Tryibion:
I did write docs for how to create your own script templates <https://docs.flaxengine.com/manual/scripting/advanced/templates.html>

xxSeys1:
That's nice, however I think some more commonly used templates should come with the engine by default

Tryibion:
Added in #3073

PR #3073, opened 2024-11-22
Add more scripting templates.

Tryibion:
Templates that are added in this commit:
- C# Actor
- C# Class
- C# Struct
- C# Interface
- C++ Actor

This also splits some of the C# templates into different sub menus. Generic (Class, Struct, Empty File, Interface) and Scripting (Actor, Script).

Related issue: <https://github.com/FlaxEngine/FlaxEngine/issues/3061>


mafiesto4:
Personally, I think we don't need to add nested `Scripting` and `Generic` groups inside `C#` templates. All templates from that language could be in one list.

PR #3082, opened 2024-11-29
Add UIMode switch to prefab window.

Tryibion:
Adds a UIMode to the prefab window to switch between UI and 3d modes. This also caches the state of the mode for each prefab.

image

### DIFF
--- a/Source/Editor/Viewport/PrefabWindowViewport.cs
+++ b/Source/Editor/Viewport/PrefabWindowViewport.cs
@@ -9,6 +9,7 @@ using FlaxEditor.SceneGraph;
 using FlaxEditor.Scripting;
 using FlaxEditor.Viewport.Cameras;
 using FlaxEditor.Viewport.Previews;
+using FlaxEditor.Viewport.Widgets;
 using FlaxEditor.Windows.Assets;
 using FlaxEngine;
 using FlaxEngine.GUI;
@@ -44,7 +45,7 @@ namespace FlaxEditor.Viewport
         private sealed class PrefabUIEditorRoot : UIEditorRoot
         {
             private readonly PrefabWindowViewport _viewport;
-            private bool UI => _viewport._hasUILinkedCached;
+            private bool UI => _viewport.ShowUI;
 
             public PrefabUIEditorRoot(PrefabWindowViewport viewport)
             : base(true)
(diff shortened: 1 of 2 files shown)


Source/Editor/Viewport/PrefabWindowViewport.cs
```diff
@@ -137,6 +208,12 @@ public PrefabWindowViewport(PrefabWindow window)
             _uiRoot = new PrefabUIEditorRoot(this);
             _uiRoot.IndexInParent = 0; // Move viewport down below other widgets in the viewport
             _uiParentLink = _uiRoot.UIRoot;
+            
+            var container = new ViewportWidgetsContainer(ViewportWidgetLocation.UpperLeft);
+            container.Parent = this;
+            _uiModeButton = new ViewportWidgetButton("UI Mode", SpriteHandle.Invalid, null, true);
```
mafiesto4, line 214: I think we should use exiting `View` button and add a checkbox for `UI Mode` (via `ViewWidgetShowMenu`).

Source/Editor/Windows/Assets/PrefabWindow.cs
```diff
@@ -355,13 +355,22 @@ private void OnPrefabApplied(Prefab prefab, Actor instance)
         private void OnPrefabOpened()
         {
             _viewport.Prefab = _asset;
-            _viewport.UpdateGizmoMode();
+            if (Editor.ProjectCache.TryGetCustomData($"UIMode:{_asset.ID}", out bool value))
+                _viewport.SetInitialUIMode(value);
+            else
+                _viewport.SetInitialUIMode(false);
```
mafiesto4, line 361: Could you try using `_hasUILinked` here? So we can still utilize automatic-UI mode when opening UI widget for the first time.

mafiesto4 (changes requested):

Source/Editor/Viewport/PrefabWindowViewport.cs
Tryibion: Done

Source/Editor/Windows/Assets/PrefabWindow.cs
Tryibion: Done

mafiesto4 (approved):

PR #3096, opened 2024-12-11
Add highlight when actor reference is single clicked

### LINKED ISSUE #3091: Show highlight when select an actor by editor property
Unity adds a highlight when i click on a gameObject property on editor. This help a lot to find actors on scene tree.

I think this is a good feature to add in flax

image

**Flax version:**
1.10 master

xxSeys1:
This adds a highlight to the Actor in the scene tree that the actor reference is referencing.

As always, I'd appreciate some feedback on how this looks and the code implementation.

https://github.com/user-attachments/assets/bb7083a4-6d82-48f0-a308-c894ae4e5a44

*Thanks to Tryibion for fixing this :)*
Also I'm encountering an issue where the highlight only works after the actors was deselected and selected again once. This only happens after the actor was dragged into the reference.

I gave fixing this problem a few attempts but couldn't get down to the root cause, so I'd appreciate some help :)

## TODO
- [x] Handle not doing highlight when double click select
- [x] Fix bug where highlight only works after actor deselect and select (most likely to do with drag over)

Closes #3091

### DIFF
--- a/Source/Editor/CustomEditors/Editors/FlaxObjectRefEditor.cs
+++ b/Source/Editor/CustomEditors/Editors/FlaxObjectRefEditor.cs
@@ -7,6 +7,7 @@ using FlaxEditor.CustomEditors.Elements;
 using FlaxEditor.GUI;
 using FlaxEditor.GUI.Drag;
 using FlaxEditor.SceneGraph;
+using FlaxEditor.SceneGraph.GUI;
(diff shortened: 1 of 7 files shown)


Tryibion:
I made a PR into your branch <https://github.com/xxSeys1/FlaxEngine/pull/1>

PR #3104, opened 2024-12-13
Add rotate position shape as particle module.

Tryibion:
This allows to do something like this <https://docs.flaxengine.com/manual/particles/tutorials/rotate-particles.html> with only this code:
image

It is nice to be able to rotate the initial shape that the particles are starting at.

Let me know if anyone thinks of a better name than "Rotate Position Shape".


xxSeys1:
Wait so this rotates the initial spawn shape of the particles, right? 
If so, then I think "Rotate Spawn Shape" would be a good name

Tryibion:
Yes, it is for the position shape that is selected. I thought about that name initially, but the shape modules are called "Position ([shape])" so I was trying to come up with a naming scheme that would prompt the user to think, "oh, it is to rotate the position shape module".

PR #3107, opened 2024-12-13
Dont add skeleton node if it is a mesh.

### LINKED ISSUE #3112: SetMasterPoseModel` will not work with the same skinned model if it has different skeleton nodes.
**Issue description:**
When importing just a skeleton to use as the master, it has a different number of skeleton nodes than the same skeleton imported with a mesh attached because the mesh is counted as a node. This causes problems with `SetMasterPoseModel` because the existing code needs the number of nodes to be the same as seen here: <https://github.com/FlaxEngine/FlaxEngine/blob/master/Source/Engine/Level/Actors/AnimatedModel.cpp#L781>.

**Steps to reproduce:**
1. Import just the skeleton with no mesh attached.
2. Import meshes with the same skeleton.
3. Use code as found here <https://docs.flaxengine.com/manual/animation/animated-model.html#modular-characters> to set the master pose. The `AnimatedModel` "master" actor should be set to the imported skeleton and the others be set to the imported skinned mesh with the same skeleton.
4. See that it wont work. View the skeleton nodes and notice the difference in nodes with the imported skinned model with a mesh having 1 or more extra nodes.

**Flax version:**
1.9

Tryibion:
This PR removes mesh nodes from the Skeleton Nodes container. This prevents issue with the same skeleton not wanting to sync because of extra mesh nodes lingering in the skeleton nodes. This fixes things like using `SetMasterPose`. - Can see conversation starting here <https://discord.com/channels/437989205315158016/438021321134309377/1317147522363560016> for an issue with the `MasterPose`

Closes #3112

### DIFF
--- a/Source/Engine/Tools/ModelTool/ModelTool.Assimp.cpp
+++ b/Source/Engine/Tools/ModelTool/ModelTool.Assimp.cpp
@@ -112,6 +112,11 @@ struct AssimpNode
     String Name;
 
     /// <summary>
+    /// Whether this node contains a mesh.
+    /// </summary>
+    bool IsMesh;
+
+    /// <summary>
     /// The LOD index of the data in this node (used to separate meshes across different level of details).
     /// </summary>
     int32 LodIndex;
@@ -163,12 +168,15 @@ struct AssimpImporterData
     {
     }
 
-    int32 FindNode(const String& name, StringSearchCase caseSensitivity = StringSearchCase::CaseSensitive)
+    int32 FindNode(const String& name, const bool excludeMeshes, StringSearchCase caseSensitivity = StringSearchCase::CaseSensitive)
     {
+        int32 numMeshes = 0;
         for (int32 i = 0; i < Nodes.Count(); i++)
         {
+            if (excludeMeshes && Nodes.At(i).IsMesh)
+                numMeshes += 1;
             if (Nodes[i].Name.Compare(name, caseSensitivity) == 0)
-                return i;
+                return i - numMeshes;
         }
         return -1;
     }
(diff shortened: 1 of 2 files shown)


Tryibion:
related: <https://github.com/FlaxEngine/FlaxEngine/issues/3112>

Tryibion:
Please review this well as I dont want to break importing if I missed something.

ElevenArt:
@Tryibion Could Mesh Nodes be preserved as an Import Option? This could be useful when you need to import a non-skeleton Animated Model into Flax as a Skinned Model, for example for Opening, Closing Doors and for other non-skeleton Animations. In such cases, Mesh Nodes replace the skeleton, and even non-skeleton Animations could be controlled via the Animation Graph. If someone needed it, they would turn it on in Import Options.

Tryibion:
> Could Mesh Nodes be preserved as an Import Option?

Importing a mesh with no skeleton nodes as a skinned model would still give the root node which could be used.

This is just my opinion... but I would prefer that mesh nodes be excluded from skeleton nodes because it is cleaner to not have both labeled as `skeleton nodes` if a node is not a skeleton node but is in fact just a mesh.

ElevenArt:
@Tryibion I'm attaching this model from Sketchfab so you can see. It's a car model where there is animation without a skeleton (Opening and closing doors, etc.). Try it out to see if it will run the animation even with this PR for skeleton nodes.

Skeleton Nodes:

Flax_SkinnedModel_MeshNodes

Animation Nodes:

Flax_Animation_Nodes

The Animation Nodes names match the Skeleton Nodes names and also match the Mesh Names in the Source Model. This is also in the Flax Docs. Deleting the Mesh Nodes would not allow using animations in Flax in this way, since in order to use the AnimationEditor and AnimGraph, the model must be imported as a SkinnedModel. The Animation Preview in AnimGraph uses the SkinnedModel as an Asset Reference and without this Asset Reference it will not run the animation in Game Mode.

Don't get me wrong, it is important that Animation Retargeting works well for this case, which is solved by this PR, but it is also important that the use case I mentioned is preserved, otherwise it would be a Downgrade.

Tryibion:
Yeah, it wont work in that case. It also looks like other engine do convert mesh nodes to skeleton nodes... so I guess I will close this and have to come up with another solution for fixing some issues. Thanks for your input!

PR #3111, opened 2024-12-19
Fix heightmap invert

Tryibion:
Texture:
image

Existing generation (top = +z, bottom = -z):
image

Correct y (top = +z, bottom = -z):
image


mafiesto4:
The reason for the current layout is that it matches texcoords orientation in a material:

image

Was this reported as a bug somewhere? (eg. discord)

Tryibion:
Yeah, it was reported as a bug on discord that the y was inverted.

Issue #3116, opened 2024-12-23
Log warning in `MaterialBase.SetParameterValue` when the value does not match the type defined in the material


ZaPx64:
**Issue description:**
When calling `myMaterial.SetParameterValue(name, myValue)` there is no log whatsover when `myValue` is of a different type than whatever is defined in `myMaterial`.
I think `GPUTexture` and `Texture` can easily be confused for example (which I did 🙄).

**Flax version:**
1.9

PR #3117, opened 2024-12-23
Warn if wrong type while setting material parameter.

Tryibion:
Closes <https://github.com/FlaxEngine/FlaxEngine/issues/3116>


ZaPx64:
Thanks!

mafiesto4 (changes requested):
You could add `Variant::CanCast(const Variant& v, const VariantType& to)` and use Variant casting so it will work better if someone passes eg. `int` for `float` param.

Tryibion:
> You could add `Variant::CanCast(const Variant& v, const VariantType& to)`

I added that now.

mafiesto4 (approved):

PR #3119, opened 2024-12-24
Allow user to add splash image to splash screen.

Tryibion:
This allows users to put a texture named "SplashImage.flax" inside of the "Content/SplashImage" folder and it will use it as the project splash screen image at start up. Recommended texture size is 600x200. I am open to suggestions on the text layout positioning.

Image of a custom splash screen:
image


xxSeys1:
This looks awesome, my suggestion for text positioning would be to add a little padding to the left text

Tryibion:
> This looks awesome, my suggestion for text positioning would be to add a little padding to the left text

Maybe something like this?
image

PR #3123, opened 2024-12-30
Add `ControlReference` for easier UI referencing.

### LINKED ISSUE #3120: Request: UIControl selection atribute for editor
**Issue description:**
Currently there is no a way to select the correct UIControl on the editor based on control type, like button, label, progress bar or any other type. It's only possible select using UIControl actor.

Script example of my game:
image

My suggestion is add a filter to select only correct type that i want.

Maybe something like this ?
```
[ControlField(typeof(Button))] public UIControl MyButton;
[ControlField(typeof(Label))] public UIControl MyLabel;
```

**Flax version:**
1.10 master

Tryibion:
Closes #3120

Filters UIControls by their control type in the drop down and drag drop.

Usage:
```
public ControlReference<Button> TestControlRefButton;
public ControlReference<Label> TestControlRefLabel;
```

image

This wrappers gives access to the Control (handles the casting to the control type) and the UIControl.

### DIFF
--- a/Source/Editor/CustomEditors/Editors/ControlReferenceEditor.cs
+++ b/Source/Editor/CustomEditors/Editors/ControlReferenceEditor.cs
@@ -1,0 +1,472 @@
+﻿using System;
+using System.Collections.Generic;
+using FlaxEditor.CustomEditors;
+using FlaxEditor.CustomEditors.Elements;
+using FlaxEditor.GUI;
+using FlaxEditor.GUI.Drag;
+using FlaxEditor.SceneGraph.GUI;
+using FlaxEngine;
+using FlaxEngine.GUI;
+using FlaxEngine.Utilities;
+using Object = FlaxEngine.Object;
+
+namespace FlaxEditor.CustomEditors.Editors;
+
+/// <summary>
+/// The reference picker control used for UIControls using ControlReference.
+/// </summary>
+public class UIControlObjectRefPickerControl : Control
+{
+    private Type _controlType;
+    public UIControl _value;
+    private ActorTreeNode _linkedTreeNode;
+    private string _valueName;
+    private bool _supportsPickDropDown;
+    
+    private bool _isMouseDown;
+    private Float2 _mouseDownPos;
+    private Float2 _mousePos;
+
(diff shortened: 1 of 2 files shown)


Source/Engine/UI/ControlReference.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// </summary>
+    public T Control
+    {
+        get
+        {
+            if (_uiControl.Control is T t)
+                return t;
+            else
+            {
+                Debug.LogWarning("Trying to get Control from ControlReference but UIControl.Control is null or UIControl.Control is not the correct type.");
```
Tryibion, line 57: Log here.

Source/Engine/UI/ControlReference.cs
```diff
@@ -0,0 +1,102 @@
...
+        if (uiControl == null)
+        {
+            Clear();
+            return;
+        }
+        
+        if (uiControl.Control is T castedControl)
+            _uiControl = uiControl;
+        else
+            Debug.LogWarning("Trying to set ControlReference but UIControl.Control is null or UIControl.Control is not the correct type.");
```
Tryibion, line 78: Log here.

Tryibion:
There is an issue where the reference value does not get set correctly and the prefab will always say that the value is not the reference value. I have dug into it a little bit and I can not figure out what is causing the reference value in the prefab to be set the way that it is being set...

MrCapy0 (approved):

Source/Editor/CustomEditors/Editors/ControlReferenceEditor.cs
```diff
@@ -0,0 +1,472 @@
...
+
+namespace FlaxEditor.CustomEditors.Editors;
+
+/// <summary>
+/// The reference picker control used for UIControls using ControlReference.
+/// </summary>
+public class UIControlObjectRefPickerControl : Control
+{
+    private Type _controlType;
+    public UIControl _value;
```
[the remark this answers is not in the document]
Tryibion: yes, it was fixed in the last commit. Thanks.

PR #3125, opened 2024-12-31
Fix local origin option with multi-mesh import.

### LINKED ISSUE #2777: Import Option Local import doesn't preserve sub object origins.
**Issue description:**
I'm importing kit pieces for the level, and I didn't realize that local import uses the main parents origin not the child's origin.

In blender you can see the orange dot that is the origin. 
image

However the origin in flax is here. 
image

I've attached the test model. 

Thanks!

**Steps to reproduce:**
Import model, check splitting, check use local transform. 

[sata_dish_roller_kit.zip](https://github.com/user-attachments/files/16276492/sata_dish_roller_kit.zip)

Tryibion:
Fix #2777

### DIFF
--- a/Source/Engine/Graphics/Models/ModelData.h
+++ b/Source/Engine/Graphics/Models/ModelData.h
@@ -91,7 +91,7 @@ public:
     Array<BlendShape> BlendShapes;
 
     /// <summary>
-    /// Global translation for this mesh to be at it's local origin.
+    /// Local translation for this mesh to be at it's local origin.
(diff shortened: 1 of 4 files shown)


ElevenArt:
@Tryibion Does the Local Origin option only work for SkinnedModels? It would be nice if this option worked for splitted Static Model and also for Model Prefab.

Tryibion:
As stated above, this PR does not target skinned models, only static models.

PR #3127, opened 2024-12-31
For Android SDK API >= 30 use non deprecated way of setting to Fullscreen.

Tryibion:
I have not tested this yet, but there should be no reason it doesnt work because this is the suggested way from Google for SDK API >= 30.


Tryibion:
I tested this now and it errors out during the gradle build. need to figure out why.

PR #3128, opened 2024-12-31
Add import path UI and button to open folder in asset import settings.

Tryibion:
Adds import path (non-editable) and button to open the import folder for each asset type

image

### DIFF
--- a/Source/Editor/Utilities/Utils.cs
+++ b/Source/Editor/Utilities/Utils.cs
@@ -384,6 +384,28 @@ namespace FlaxEditor.Utilities
         }
 
         /// <summary>
+        /// Creates an Import path ui that show the asset import path and adds a button to show the folder in the file system.
+        /// </summary>
+        /// <param name="parentLayout">The parent layout container.</param>
+        /// <param name="assetItem">The asset item to get the import path of.</param>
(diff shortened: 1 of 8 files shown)


xxSeys1:
This is super cool, but one issue I can see is people getting confused because they can't edit the import path. I don't have much experience with Flax text rendering, but is there a way to maybe grey out the path slightly to make it more obvious that it can't be edited?

Right now the white text suggests that the text is editable in my opinion.

Tryibion:
maybe, the disabled text color is the text color * 0.6f to be more grey. I guess we could do something inbetween for read only. maybe textcolor * 0.85f? it would look like this: 
image

Tryibion:
I added the color change to a different PR here <https://github.com/FlaxEngine/FlaxEngine/pull/3129>

mafiesto4 (changes requested):
Please, add a utility method in `FlaxEditor.Utilities.Utils` to reuse in different windows and reduce code duplication.

Also, you could try changing button text to `FlaxEditor.Utilities.Constants.ShowInExplorer` so it uses `Finder` name on Mac.

Tryibion:
> Please, add a utility method in `FlaxEditor.Utilities.Utils` to reuse in different windows and reduce code duplication.

Done.

mafiesto4 (approved):

PR #3131, opened 2025-01-02
Add android emulation options to game cooker android tab.

Tryibion:
Adds an easier way to emulate a built android application. The UI is rough right now, but it does the job.
This will find AVDs on the system, let the user select one and start it. Can find started AVDs and connected physical devices and can auto install it on the selected device. It will even try to start the activity automatically if it can. 

Editor options (in the android section of the Game Cooker):
image

Running on emulated device:
image

This needs a little clean up before it is merged. I had to use C# process instead of Flax's Platform process because it seems that using the `SaveOutput` option crashes the editor (and this option is needed to get the different lists). Using the built in Flax Platform process it also will freeze the editor in certain cases like starting the adb log collection (even when using `WaitForEnd = false`).

A few notes from me creating this. 

- the AVDs worked better with these options (which I had to edit in the ini files):
hw.gpu.enabled=no
hw.gpu.mode=off

- I had to add ANDROID_AVD_HOME environment variable where the AVDs are stored (usually %user%\.android\avd)


Source/Editor/Windows/GameCookerWindow.cs
```diff
@@ -343,6 +347,269 @@ public override void Initialize(LayoutElementsContainer layout)
...
+                                
+                                var processStartInfo = new System.Diagnostics.ProcessStartInfo
+                                {
+                                    FileName = Path.Combine(sdkPath, "emulator", "emulator.exe"),
+                                    Arguments = "-list-avds",
+                                    RedirectStandardOutput = true,
+                                    CreateNoWindow = true,
+                                };
+
+                                var process = new System.Diagnostics.Process
```
mafiesto4, line 386: Have you tried using `FlaxEngine.Platform.CreateProcess` and `FlaxEngine.CreateProcessSettings` to start process via Flax API instead of .NET one? It supports capturing output as array of chars and auto-logging process to the Output Log.

Tryibion: Yes I tried this originally, but there is a crash when using the built in Flax platform to start a process when capture output is enabled and I could not figure out why it was crashing.

Tryibion:
This could also just be migrated into an `AndroidTools` project plugin if it would better fit there.

Source/Editor/Windows/GameCookerWindow.cs
Tryibion: There was also an issue using the built in platform process where the editor just freezes when starting an emulator.

mafiesto4: Okay fine. Let's use it for now.

mafiesto4:
@Tryibion Could you move that Android-specific code into some `virtual` function inside `GameCookerWindow.Platform` like `virtual void OnCustomTools(layout)` and override it specific for Android? In future, we might add similar utilities for other platforms. At some point maybe shared Device Manager or similar but for now that's good.

PR #3150, opened 2025-01-12
Add .aab bundling for Android.

Tryibion:
Adds the option to bundle the android app as an .aab instead of .apk. This can be done via the `Build .aab` checkbox or add a custom define of `BuildAAB` which the checkbox does automatically.

image

### DIFF
--- a/Source/Editor/Cooker/Platform/Android/AndroidPlatformTools.cpp
+++ b/Source/Editor/Cooker/Platform/Android/AndroidPlatformTools.cpp
@@ -379,7 +379,7 @@ bool AndroidPlatformTools::OnPostProcess(CookingData& data)
             }
         }
         // Copy result package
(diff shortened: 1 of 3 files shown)


mafiesto4:
Great! I think we can add additional option to `AndroidPlatformSettings` in case game cooker is used from command line (eg. on build machine) and output should be `.aab` for Google Play Store

Tryibion:
So you want the option moved to the platform settings instead of in the game cooker?

mafiesto4:
Yeah `Build .aab` makes sense to be platform option. Just like Textures Quality. Otherwise we might end up with lots of CustomDefines being used in the cooker so lets keep platform options only.

Tryibion:
Okay, Moved the setting into the `AndroidPlatformSettings` and made it side-by-side with the .apk build.

PR #3151, opened 2025-01-13
Add rubber band select to scene editor viewport for selecting multiple actors.

Tryibion:
Adds left click and drag to select multiple actors in the scene editor view window.

https://github.com/user-attachments/assets/55e239e0-f67d-4929-a7a2-a364934bd34b

### DIFF
--- a/Source/Editor/Viewport/MainEditorGizmoViewport.cs
+++ b/Source/Editor/Viewport/MainEditorGizmoViewport.cs
@@ -7,10 +7,12 @@ using FlaxEditor.Gizmo;
 using FlaxEditor.GUI.ContextMenu;
 using FlaxEditor.SceneGraph;
 using FlaxEditor.Scripting;
+using FlaxEditor.Tools;
 using FlaxEditor.Viewport.Modes;
 using FlaxEditor.Windows;
 using FlaxEngine;
 using FlaxEngine.GUI;
+using FlaxEngine.Tools;
 using Object = FlaxEngine.Object;
 
 namespace FlaxEditor.Viewport
@@ -107,6 +109,10 @@ namespace FlaxEditor.Viewport
         private double _lockedFocusOffset;
         private readonly ViewportDebugDrawData _debugDrawData = new ViewportDebugDrawData(32);
         private EditorSpritesRenderer _editorSpritesRenderer;
+
+        private bool _isRubberBandSpanning;
+        private Float2 _cachedStartingMousePosition;
+        private Rectangle _rubberBandRect;
 
         /// <summary>
         /// Drag and drop handlers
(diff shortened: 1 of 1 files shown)


Source/Editor/Viewport/MainEditorGizmoViewport.cs
```diff
@@ -576,15 +610,202 @@ protected override void OrientViewport(ref Quaternion orientation)
                 base.OrientViewport(ref orientation);
         }
 
+        /// <inheritdoc />
+        public override void OnMouseMove(Float2 location)
+        {
+            base.OnMouseMove(location);
+
+            // Dont allow rubber band selection when gizmo is controlling mouse, vertex painting mode, or cloth painting is enabled
+            if (_isRubberBandSpanning && ((Gizmos.Active.IsControllingMouse || Gizmos.Active is VertexPaintingGizmo || Gizmos.Active is ClothPaintingGizmo) || IsControllingMouse || IsRightMouseButtonDown))
```
mafiesto4, line 619: I would move whole logic from this function (and some local state vars fro the top) into a separate class like `RubberBandSelector` to be reusable in other viewports (like prefab in future) and have more organized code.

Source/Editor/Viewport/MainEditorGizmoViewport.cs
```diff
@@ -576,15 +610,202 @@ protected override void OrientViewport(ref Quaternion orientation)
...
+                        }
+                        else if (a is UICanvas uiCanvas)
+                        {
+                            if (uiCanvas.Is2D)
+                                continue;
+                        }
+
+                        var containsAllPoints = true;
+                        var fallBackToBox = false;
+                        if (a is StaticModel sm)
```
mafiesto4, line 691: A little documentation comment would help to understand why we do this mesh-specific code. Also, it would be nice to make it virtual call to `SceneGraphNode` so other models could implement or customize it in future.

Source/Engine/Graphics/Mesh.cs
```diff
@@ -633,5 +633,14 @@ public ushort[] DownloadIndexBufferUShort(bool forceGpu = false)
                 throw new Exception("Failed to download mesh data.");
             return result;
         }
+
+        /// <summary>
+        /// Gets the collision proxy points for the mesh.
+        /// </summary>
+        /// <returns>The triangle points in the collision proxy.</returns>
+        public Float3[] GetCollisionProxyPoints()
```
mafiesto4, line 641: Can you make it `internal` for now? On 1.10 branch there was refactor for mesh data and maybe I'll followup on this code. But for editor-only purpose it's fine.

mafiesto4 (changes requested):

Source/Editor/Viewport/MainEditorGizmoViewport.cs
Tryibion: Done

Source/Editor/Viewport/MainEditorGizmoViewport.cs
Tryibion: Done

Source/Engine/Graphics/Mesh.cs
Tryibion: Done

Issue #3195, opened 2025-01-31
[Issue/Bug] The variable declared with the Steam plugin class gives me an error


ThePhantomMask:
**Issue description:**
When due to technical problems on the project I had to reconfigure the project I went to check the script that is dedicated to online services and realised that the variable that is initialised using the Steam plugin class gives an error as if the class did not exist, as can be seen in the screenshot below:
![Image](https://github.com/user-attachments/assets/6b8c77e1-f4b8-41e6-ad04-a39791945491)
But it works when I put a debug print on the steam nickname or the friend list and when I see Steam it shows me that I am playing Space War.

**Steps to reproduce:**
1. Open the sln file
2.  Go to check the Game Online Services Manager (GOSManager) script
3.  Try instantiating an OnlinePlatformSteam variable

**Minimal reproduction project:**
[https://drive.google.com/file/d/12AwcG6HUJe6ku4_n628Lyv6c87_vfy1a/view?usp=sharing](url)

**Flax version:**
Master

GoaLitiuM:
Visual Studio expects the plugin project to be fully compiled before the Intellisense starts working properly, but in this project the build process didn't seem to work for some reason, so it was never compiled, thus the errors. The packaged project seems to have built files for `Editor.Development` configuration, so I assume the screenshot was taken with `Editor.Debug` configuration where the files were missing?

ThePhantomMask:
> Visual Studio expects the plugin project to be fully compiled before the Intellisense starts working properly, but in this project the build process didn't seem to work for some reason, so it was never compiled, thus the errors. The packaged project seems to have built files for `Editor.Development` configuration, so I assume the screenshot was taken with `Editor.Debug` configuration where the files were missing?

Checked, had Editor.Debug

ThePhantomMask:
The problem was effectively solved

PR #3196, opened 2025-01-31
Fix C# projects not compiling in VS with native code plugins present

GoaLitiuM:
`Globals.Project.IsCSharpOnlyProject` was expected to be true for C#-only game projects, but this also includes plugins which may have native modules present, so this can't be used with the main project detection. Use instead the same value that is used to select the start up project for VS solution projects.


GoaLitiuM:
Maybe not fully fixes the #3195 but at least allows the project to be compiled in IDE.

PR #3205, opened 2025-02-05
GPUFence,back end code, Graphic API changes

cNori:
will be converted to pr when ready

ToDo:
- [x] implment DX11
- [x] implment DX12
- [x] implment Vulkan maybe
- [x] remove from map the 
https://github.com/FlaxEngine/FlaxEngine/blob/8542f78b0bebe3f624a93649fbb2f00c362b97a7/Source/Engine/GraphicsDevice/DirectX/DX11/GPUBufferDX11.cpp#L50
it is incorrect usage it is a error `The previous blit operation that is transferring information to or from this surface is incomplete.`
and is ignored

- [x] Add proper error handling for map and unmap,(DX11,DX12)
 log for **Debug** and **Development** builds **Release** build hard crash (Assert) if returned status is not ok 

 maybe more

### DIFF
--- a/Source/Engine/Graphics/GPUContext.h
+++ b/Source/Engine/Graphics/GPUContext.h
@@ -623,4 +623,7 @@ public:
     /// Forces graphics backend to rebind descriptors after command list was used by external graphics library.
     /// </summary>
     virtual void ForceRebindDescriptors();
+
+public:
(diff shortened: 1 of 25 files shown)


cNori:
unable to implment it correctry API design is not up to the task (DX12 and vulkan) ishue

PR #3214, opened 2025-02-11
Triplanar features, Shader String Builder, Color blend Nodes, Default gradient change. 

Muzz:
This pull request is fairly extensive. 

The triplanar normal mapping nodes are quite expensive as they have to do a fair bit of math to work correctly. I used this article as reference. 
https://bgolus.medium.com/normal-mapping-for-a-triplanar-shader-10bf39dca05a

It adds these nodes
- local space triplanar
- local space triplanar normal
- world space triplanar normal
- tangent to world normal
- world to tangent normal
- local generated UVS

it also adds sampler dropdowns for each. 

Ontop of that I'm contributing a ShaderStringBuilder in visject utilities to make writing in c++ shader code less of a headache. 

I have a test project file that's a little big for github that i can send you to verify the features work. I'll optimize the textures when i get a chance to throw it online.


Muzz:
I've added some more utility nodes, and changed the default gradient to black and white, as that's the most common starting point for a gradient node.

mafiesto4:
Good job!

I'm doing merging now. I'll cleanup some codestyle and probably merge those World and Local triplanar nodes into a single one with a checkbox whether to evaluate it in object local-space for simplicity and less code. Also, the offset could be refactored into `Float3` as it's world-space and even made as just `Position` input where user could plug `World position` node with custom offset or even custom wsPos from other place to sample at.

Finally, I'll remove those Tangent/World vector transform nodes as we have better utility that does that and some more:
image

mafiesto4:
Merged in 7885590593e1e46441b469f5dada9720fe40207e. I've got to squish your commits as there was `flax` file modification that fails Git LFS.

New triplanar works great! I hope it's all good on your end after merge.

image

Also, I removed that Local 3D UVs node as I was unable to use it properly. You can open new PR with it if you think it's usable for others.

Muzz:
@mafiesto4 

Great, it was a lot of code, so I expected some changes! thanks for that. 

I'll revisit what i was doing with the local generated UV's and see if it was actually needed, I didn't realize there was those vector transforms, I think vector transform as a name tripped me up a bit. 

If you have any feedback on how I can make my pull requests less work in the future, that would be helpful too!

PR #3251, opened 2025-02-25
Adding Diffuse light to env probes

Muzz:
This adds the ability to use environment probes for ambient light. 

This is an alternative to DDGI. 

Atm I have not coded in settings for it, as I'm not sure what's the best practice for doing shader permutations. So the final bit of work is having a switch between this, and the old behavior. Easiest way to do that is to just blacken the diffuse buffer, but that's a bit wasteful on resources. 

Here is a direct comparison of the techniques. 

Looks like our editors have a bit of a difference in how it handles white space, so sorry about the fake changed lines. 

image

### DIFF
--- a/Source/Engine/Renderer/ReflectionsPass.cpp
+++ b/Source/Engine/Renderer/ReflectionsPass.cpp
@@ -33,212 +33,212 @@ GPU_CB_STRUCT(Data {
 
 namespace PreIntegratedGF
 {
-	static const int Resolution = 128;
-	static const int NumSamples = 512;
+    static const int Resolution = 128;
+    static const int NumSamples = 512;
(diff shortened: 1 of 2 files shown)


TheIronLefty:
Hello,
I am curious isn't this very similar to what The Order 1886 was doing by employing Spherical Gaussian techniques to contain both Diffuse and Specular information into their probes?

https://therealmjp.github.io/posts/new-blog-series-lightmap-baking-and-spherical-gaussians/

Muzz:
@TheIronLefty Looking into it briefly, it looks very similar to me. Thanks for the link, I'll do some reading and see if there is any improvements i can add from it.

PR #2691, opened 2024-06-07
Color picker improvements

xxSeys1:
This pr aims to make multiple improvements to the color picker. Mostly UX stuff and polish but also a few new features.
I'm planing to get this ready for so that it can be merged for 1.9 but no promises on that :)

**If you have the time please try this out and give me some feedback, I don't wanna make things worse by trying to make them better :)**

There are several mockups for a new layout [here](https://www.figma.com/design/g7w9SI7RFP18CJHVonIPV3/FlaxColorPicker?node-id=0-1&t=13KdiP5tRbNXUsz2-1).

This is the current state of the color picker with all of the progress made so far:

https://github.com/FlaxEngine/FlaxEngine/assets/111653551/3c12eb00-f412-4bee-b69b-b1a76ecb32a6

*Tasks marked with " * " are tasks where I need help or don't know how to approach them atm*
### Todo

- [ ] BUG: Fix bug where you can't drag the wheel when *HSV = 0, 0, 0* and *Alpha > 1*
- [ ] BUG: After exiting eye dropper with right click you need to click on the editor once for it to register mouse events again
- [ ] Include A and S slider handles in calculation if clicked
- [ ] Get rid of *Ok* and *Cancel* buttons (see #2446)
- [ ] Checkerboard background for A slider similar to color preview
- [ ] A way to reset color to Old color from color preview
- [ ] Multiple improvements to A and S sliders
- [ ] Try drawing previews on sliders similar to Unreal (see if it looks good)
- [ ] Tint HV wheel to reflect current Saturation *
- [ ] Try to counteract fuzzy edges on HV wheel by drawing an outline (I can't get bezier curves to work like I wanna atm) *

### Todo if there is enough time

- [ ] Add SV square and H slider
- [ ] Rework saved colors system
- [ ] Procedurally draw HV wheel *

### Done

- [x] Old and new color preview
- [x] Resizing HV wheel selection box that shows the currently selected color
- [x] Hide mouse cursor when dragging on HV wheel and AS sliders

PR #3281, opened 2025-03-13
Improve how ColorValueBoxes draw transparent colors

xxSeys1:
This pr aims to make it easier to recognize if a color has transparency or not.

For that, if the alpha < 1, a rectangle is displayed that shows the color with transparency.

I originally wanted to just draw a grid behind the color like Godot does it, but I think it's too distracting.
*Godot*
image

What could improve this even further is adding a grid behind the transparent part of the color, but idk how that would be performance wise because it'd be a bunch of rectangle draws. Maybe a shader would be a better fit here, but I don't even know where to start with that.

(#2691 could also benefit from such a shader since it uses transparency grids in a few places).

Before:
image

With this pr:
image

### DIFF
--- a/Source/Editor/GUI/Input/ColorValueBox.cs
+++ b/Source/Editor/GUI/Input/ColorValueBox.cs
@@ -129,11 +129,20 @@ namespace FlaxEditor.GUI.Input
         {
             base.Draw();
 
-            var style = Style.Current;
-            var r = new Rectangle(0, 0, Width, Height);
+            bool isTransparent = _value.A < 1;
 
-            Render2D.FillRectangle(r, _value);
-            Render2D.DrawRectangle(r, IsMouseOver || IsNavFocused ? style.BackgroundSelected : Color.Black);
+            var style = Style.Current;
+            var fullRect = new Rectangle(0, 0, Width, Height);
+            var colorRect = new Rectangle(0, 0, isTransparent ? Width * 0.7f : Width, Height);
+
+            if (isTransparent)
+            {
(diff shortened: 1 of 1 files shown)


mafiesto4:
I think we can do it without a new shader but just like the Color Picker dialog does: draws the checkerboard below the color so it nicely blends with alpha:

image

mafiesto4:
The code from `ColorPickerDialog.cs`:
```cs
// Color difference
var newRect = new Rectangle(_cOK.X - 3, _cHex.Bottom + PickerMargin, 130, 0);
newRect.Size.Y = 50;
Render2D.FillRectangle(newRect, Color.White);
var smallRectSize = 10;
var numHor = Mathf.FloorToInt(newRect.Width / smallRectSize);
var numVer = Mathf.FloorToInt(newRect.Height / smallRectSize);
// Draw checkerboard for background of color to help with transparency
for (int i = 0; i < numHor; i++)
{
    for (int j = 0; j < numVer; j++)
    {
        if ((i + j) % 2 == 0)
        {
            var rect = new Rectangle(newRect.X + smallRectSize * i, newRect.Y + smallRectSize * j, new Float2(smallRectSize));
            Render2D.FillRectangle(rect, Color.Gray);
        }
    }
}
Render2D.FillRectangle(newRect, _value);
```

xxSeys1:
> I think we can do it without a new shader but just like the Color Picker dialog does: draws the checkerboard below the color so it nicely blends with alpha:
> image

@mafiesto4 Done in https://github.com/FlaxEngine/FlaxEngine/pull/3281/commits/be7871c29206c0df5cdec0c9d382aa54aacc1d70

PR #3332, opened 2025-04-02
Small `WheeledVehicle` api and Properties Panel improvements

xxSeys1:
Reviews are highly welcome as this is the first time I'm doing C++ stuff :)

This does the following things:

### Move "Anti Roll Bars" into the "Vehicle" group
image

### Add getters for more commonly used vehicle fields
- `GetHandbrake()`
- `GetBrake()`
- `GetSteering()`

### Add an option to debug draw some text to make dealing with wheels easier
*Every* time I use Wheeled Vehicle I implement this in C#, so now it's time to put this into the engine by default.

Currently this only display the index of the Wheel in the wheels array and the collider name, cause imo that's the most important/ hard to tell thing.
This debug draw can be disabled with the "Display Wheel Info" checkbox, and the color can be customized with "Wheel Info Color".

image

image


mafiesto4:
Thanks! But I did some changes, I hope it's fine.

PR #3344, opened 2025-04-07
Initial support for API_FUNCTION in structs.

Tryibion:
Adds initial support for using API_FUNCTION in structs and creating the bindings.


mafiesto4:
I get some compile errors with a test structure:
```cpp
// Test structure.
API_STRUCT(NoDefault) struct TestStruct1
{
    DECLARE_SCRIPTING_TYPE_MINIMAL(TestStruct1);

    // Var
    API_FIELD() Float3 Vector = Float3::One;

    // Member function in a structure
    API_FUNCTION() int32 TryMe(Actor* actor) { return 1; }
};
```

Probably need to add some tweaks. Also, it would be nice to add a simple member function like this one to `TestScripting.h` so this coodegen is validated on Tests.

PR #3360, opened 2025-04-09
Add controls and filter to actor toolbox search (follow up to #3354)

xxSeys1:
This pr adds a search filter dropdown to the search actor toolbox tab and it allows to search for ui controls.

image

Current filter options are:
- UI
- Actors
- Models
image

### TODO
*(I'm trying to rush this into 1.10, because with this #3354 makes a lot more sense, hence why this is a draft for now. Will pick this back up later when I have time again)*

- [ ] Fix search box not having proper width directly after creation:
image

### DIFF
--- a/Source/Editor/Windows/ToolboxWindow.cs
+++ b/Source/Editor/Windows/ToolboxWindow.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Collections.Generic;
+using FlaxEditor.GUI.ContextMenu;
 using FlaxEditor.GUI.Input;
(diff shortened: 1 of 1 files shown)


Tryibion (commented):
Just make this this:
```
 _searchBox = new SearchBox
  {
      AnchorPreset = AnchorPresets.HorizontalStretchTop,
      Parent = _groupSearch.Parent.Parent,
      Bounds = new Rectangle(_viewDropdown.Right + 2, 2, _actorGroups.Width - 4, TextBoxBase.DefaultHeight),
  };
```

PR #3368, opened 2025-04-10
Fix VisualScriptWindow and tabs not getting restored properly

GoaLitiuM:


mafiesto4:
This one conflicts with #4227

PR #3372, opened 2025-04-10
Add a info message on Debug Log pause on error

xxSeys1:
I was confused for a good few minutes about why my game would randomly pause, until I noticed that the "Pause on Error" button was turned on in the Debug Log. Hopefully this log message will save another user "a good few minutes" in the future.

image

This also add a trailing "." to some tool tips that had it missing.

### DIFF
--- a/Source/Editor/Windows/DebugLogWindow.cs
+++ b/Source/Editor/Windows/DebugLogWindow.cs
@@ -488,7 +488,7 @@ namespace FlaxEditor.Windows
             // Pause on Error (we should do it as fast as possible)
             if (newEntry.Group == LogGroup.Error && _pauseOnErrorButton.Checked && Editor.StateMachine.CurrentState == Editor.StateMachine.PlayingState)
             {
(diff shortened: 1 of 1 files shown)


Source/Editor/Windows/DebugLogWindow.cs
```diff
@@ -488,6 +488,7 @@ public void Add(ref LogEntryDescription desc)
             // Pause on Error (we should do it as fast as possible)
             if (newEntry.Group == LogGroup.Error && _pauseOnErrorButton.Checked && Editor.StateMachine.CurrentState == Editor.StateMachine.PlayingState)
             {
+                Debug.Write(LogType.Info, "Pause Play mode on error (toggle this behaviour in the Debug Log panel)");
```
Tryibion, line 491: You can use `Editor.Log` here.

xxSeys1: Done in https://github.com/FlaxEngine/FlaxEngine/pull/3372/commits/169d3e964d9e51e074319242a57509b84176bb3d

PR #3375, opened 2025-04-11
Add option to draw ticks to slider

xxSeys1:
First time doing ui stuff, please tell me if I'm doing something wrong.

This is inspired by Godots option to draw ticks on sliders, but with some extra options like offset or tick size.

image

image

image


Tryibion:
Should there be an option to snap to the ticks?

xxSeys1:
> Should there be an option to snap to the ticks?

Maybe an option to snap the slider knob to fixed increments? I'll look into that later.

That'd allow the snapping to be independent from the ticks.

PR #3398, opened 2025-04-24
Add input event and input axis to Cpp scripting.

Tryibion:
Also removes Triggered from input event.

I had to implement using the custom editor alias before the flax obj reference to make the editor work once moving it to C++. I think there should be a better way to pick between a reference or not.

### DIFF
--- a/Source/Editor/CustomEditors/CustomEditorsUtil.cs
+++ b/Source/Editor/CustomEditors/CustomEditorsUtil.cs
@@ -84,6 +84,14 @@ namespace FlaxEditor.CustomEditors
 
                 return new ArrayEditor();
             }
+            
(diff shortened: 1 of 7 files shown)


mafiesto4:
Changing this from basic # object into `ScriptingObject` means it needs to be disposed with Flax `Destroy` and adds some overhead (more memory and code logic involved). I'm not sure if it's viable just to have the same code in C++ and C#. Maybe we could just add similar class in C++ that would be a wrapper over `Input` system instead. I'm open for suggestions from others.

Tryibion:
What ever you think is best. I was trying to combine them so that way they were combined, but if you want them separated, then that is pretty easy to do.

PR #3406, opened 2025-04-26
Add holding down the shift key to skip auto resizing a box collider on actor add

xxSeys1:
Does exactly what the title says.

The reason for this pr is that sometimes I don't want the collider to auto resize.


xxSeys1:
lmao doesn't work half of the time in my branch, gonna close. Maybe I'll revisit this in the future.

PR #3429, opened 2025-05-04
Refactor .NET version detection error reporting

Weyzu:
This PR improves the clarity of error messages when detecting .NET SDK and runtime versions.

### Background
When a specific .NET version is configured, the tool should use exactly that version, not a higher one. Previously, error messages were confusing when higher versions were installed but not the specifically the configured one. For example, if .NET 9 was installed but .NET 8 was configured, the log would say "Unsupported .NET SDK versions found: 9.0.101. Minimum version required is 8" which incorrectly suggested a bug in version detection.

The PR originally aimed to fix version detection while it was pointed out that it's the error message that needs to be updated. Leaving previous PR description below.

<details>
<summary>Previous PR Description</summary>
The current logic in `DotNetSdk.GetVersion` incorrectly filters out potentially valid higher versions of the .NET SDK when `Configuration.Dotnet` is set to only a major version number (e.g., "8" or "7").

When `Configuration.Dotnet` is "8":
1.  `ParseVersion(Configuration.Dotnet)` returns `null` because "8" isn't a full version string.
2.  The `int.TryParse` fallback goes in motion, and `dotnetVerNum` is correctly set to `8`.
3.  The code then enters iterates through available SDK versions.
4.  The check `if (dotnetVerNum != v.Major) continue;` is executed on every version. The check incorrectly causes the loop to `continue` for any installed SDK version where the major version `v.Major` is *not equal* to `dotnetVerNum` (e.g., it skips .NET 9 when `dotnetVerNum` is 8, while .NET 9 is an allowed maximum .NET version).

This prevents the selection of a valid, installed higher SDK version (like .NET 9) even if it falls within the acceptable range defined by `MinimumVersion` and `MaximumVersion`. 

**Solution:**

A simple change in version comparison does the trick:

```csharp
if (dotnetVerNum > v.Major)
    continue;
```

This change ensures that we only skip SDK versions whose major version is *lower* than the specified `dotnetVerNum`. Versions with a major version *equal to or greater than* `dotnetVerNum` will now correctly proceed to the subsequent check:

```csharp
if (v.Major >= MinimumVersion.Major && v.Major <= MaximumVersion.Major)
    return version;
```

Other option would be to remove the `if (dotnetVerNum != v.Major)` check completely, though I'm not sure if it served some other purpose.
</details>

### DIFF
--- a/Source/Tools/Flax.Build/Build/DotNet/DotNetSdk.cs
+++ b/Source/Tools/Flax.Build/Build/DotNet/DotNetSdk.cs
@@ -283,8 +283,10 @@ namespace Flax.Build
             Log.Verbose($"Found the following .NET SDK versions: {string.Join(", ", dotnetSdkVersions)}");
             Log.Verbose($"Found the following .NET runtime versions: {string.Join(", ", dotnetRuntimeVersions)}");
 
+            string configuredDotnetVersion = Configuration.Dotnet;
             var dotnetSdkVersion = GetVersion(dotnetSdkVersions);
             var dotnetRuntimeVersion = GetVersion(dotnetRuntimeVersions);
+
             if (!string.IsNullOrEmpty(dotnetSdkVersion) && !string.IsNullOrEmpty(dotnetRuntimeVersion) && ParseVersion(dotnetRuntimeVersion).Major > ParseVersion(dotnetSdkVersion).Major)
             {
                 // Make sure the reference assemblies are not newer than the SDK itself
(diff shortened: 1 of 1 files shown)


GoaLitiuM:
The change would cause CI builds to use .NET 9 SDK despite 8 being requested, and this in turn will cause compilation issues due to mixed runtime versions for end-users if they only have .NET 8 SDK installed on their system. If the `Configuration.Dotnet` value is not given, the build system should use the highest available .NET SDK version found in the system.

Weyzu:
I see, good to know!

Then the only thing to update would then be the logs. When a project is being built on a host with .NET 9 SDK, while .NET 8 SDK is set in the configuration (AFAIK it's currently somewhere set as the default) the following logs are served (an example from our CI):

```
[ 00:00:38.476 ]: [Info] Unsupported .NET SDK versions found: 9.0.101, 7.0.405, 6.0.418. Minimum version required is .NET 8.
[ 00:00:38.478 ]: [Info] Exception: Missing .NET SDK 8.
```

First messages suggest a bug in version detection/comparision (which lead my train of thought), the other is a bit better suiting to what actually was expected/is required (tho log level should be error, I think).

I can perhaps use the occasion (and this PR 🙃) and propose a different commit, with small improvement of that phrasing.

mafiesto4:
Yeah we could change message to inform user that specific .NET version is required.

Weyzu:
I updated the PR description, title and pushed updated commit with a small refactor of the messages. I left them at "Warning" level. I Hope it looks good. :)

PR #3430, opened 2025-05-04
Reuse native array allocations in wrapped managed to native calls

GoaLitiuM:


mafiesto4:
That's cool but what about adding some shared stack-allocator to the engine for single-frame allocs so it could be used in such alloc+free scenarios like those marshalling? I'd like to use sth like that in other engine areas where we need some memory but just for a single func within engine update.

mafiesto4:
I think we could solve both this and improve #4230 with a shared allocator for bindings.

PR #3432, opened 2025-05-04
Add create file Content Panel toolstrip button

### LINKED ISSUE #2406: [Proposal] Button to quickly add a new folder in Content panel
**Issue description:**

Add a button to quickly add a new folder in the currently open directory to the Content panel (maybe also a button to create a new prefab? Or a general New Button that gives u a dropdown similar to the right click context menu alongside the new folder button?)

FlaxNewFolderIcon Proposal
_I've used [this](https://pixabay.com/vectors/folder-file-document-icon-2935356/) icon and adjusted line width, color and added a plus_

**Steps to reproduce:**

None

None

**Flax version:**

1.8

xxSeys1:
Closes #2406.

Options that aren't available in the current folder will be disabled (grayed out).
Allows creating a new folder on shift + left click.

image

### DIFF
--- a/Source/Editor/Windows/ContentWindow.ContextMenu.cs
+++ b/Source/Editor/Windows/ContentWindow.ContextMenu.cs
@@ -39,8 +39,7 @@ namespace FlaxEditor.Windows
                 folder = CurrentViewFolder;
             }
             Assert.IsNotNull(folder);
-            bool isRootFolder = CurrentViewFolder == _root.Folder;
-
+            
             // Create context menu
             ContextMenuButton b;
             ContextMenu cm = new ContextMenu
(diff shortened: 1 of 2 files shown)


mafiesto4:
In my opinion, unavailable options should be gone as they make the menu pretty large and less readable.

xxSeys1:
> In my opinion, unavailable options should be gone as they make the menu pretty large and less readable.

My idea was to not break muscle memory by keeping unavailable options there, just in a disabled state.

Maybe this could be an editor setting? All of the functions to create the menu entries are already set up in a way where you can toggle disabling of unavailable options by passing in a bool.

xxSeys1:
> In my opinion, unavailable options should be gone as they make the menu pretty large and less readable.

@mafiesto4 I added an editor option for this in https://github.com/FlaxEngine/FlaxEngine/pull/3432/commits/b43ed886b9a6e71585bb774b4f360649f1f0f992 (unavailable options will be hidden by default).

In my opinion this is better for muscle memory. Also Unity does it like this, so this will help ex- Unity users to feel more at home.

Issue #3258, opened 2025-03-03
Graphic API full suport for compute shader


cNori:
currently engine partially supports the compute shaders
any read from RWBuffer are not supported do to not being synchronized correctly (GPU->CPU)
there is no work around for it right now i tried everything
back end is missing the memory GPU fences and other synchronization objects
GetData funcion for dx11 is ignoring actual error

PR #3437, opened 2025-05-07
FinishGPUCommands (Force exec and sync)

cNori:
force sync GPU-> CPU

dx11 tested works
dx12 crash not on (FinishGPUCommands)
vulkan crash  not on (FinishGPUCommands)

usage
```c#
GPUDevice.Instance.MainContext.CopyBuffer(NormalsReadback, NormalsGPUBuffer, sizeInBytes);
GPUDevice.Instance.MainContext.FinishGPUCommands();
var pBuff = NormalsReadback.Map(GPUResourceMapMode.Read);
```
with out this func Readback cant work explened in https://github.com/FlaxEngine/FlaxEngine/issues/3258

funcions not changed
|  broken funcions |   status                                                           |
|-----------------|-------------------------------------------------|
|GetData            | broken, sync point is missing result UB|
|DownloadData |  broken, sync point is missing result UB|

example project
[ComputeShadersExample.zip](https://github.com/user-attachments/files/20074434/ComputeShadersExample.zip)

### DIFF
--- a/Source/Engine/Graphics/GPUContext.h
+++ b/Source/Engine/Graphics/GPUContext.h
@@ -618,6 +618,8 @@ public:
     /// </summary>
     API_FUNCTION() virtual void Flush() = 0;
 
+    API_FUNCTION() virtual void  FinishGPUCommands() = 0;
+
     /// <summary>
     /// Sets the state of the resource (or subresource).
     /// </summary>
--- a/Source/Engine/GraphicsDevice/DirectX/DX11/GPUContextDX11.cpp
+++ b/Source/Engine/GraphicsDevice/DirectX/DX11/GPUContextDX11.cpp
@@ -754,6 +754,25 @@ void GPUContextDX11::Flush()
         _context->Flush();
 }
 
+void GPUContextDX11::FinishGPUCommands()
+{
+    // Create a query object to wait for GPU completion
+    D3D11_QUERY_DESC queryDesc = {};
+    queryDesc.Query = D3D11_QUERY_EVENT;
+
+    ID3D11Query* query;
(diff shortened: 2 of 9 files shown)


mafiesto4:
There is `GPUDevice::WaitForGPU()` that has been implemented on all backends (incl. PS4/PS5..) which could be used instead (eg. expose to C#?). But it needs the command buffer to be closed and submitted.

In my opinion, flushing GPU to go idle in sync with CPU kills performance and even during loading screen it might cause stuttering. Your idea with programmable fence objects that game code could inject and wait on completion event sounds better to be implemented across the board.

cNori:
>There is GPUDevice::WaitForGPU() that has been implemented on all backends (incl. PS4/PS5..) which could be used instead (eg. expose to C#?). But it needs the command buffer to be closed and submitted.

>In my opinion, flushing GPU to go idle in sync with CPU kills performance and even during loading screen it might cause stuttering. Your idea with programmable fence objects that game code could inject and wait on completion event sounds better to be implemented across the board.

im awere of "flushing GPU to go idle in sync with CPU kills performance" it but i have done a deep dive in to the source it just feels like it needs to be re thinked
unity has concept of 
https://docs.unity3d.com/6000.1/Documentation/ScriptReference/Rendering.CommandBuffer.html
and
https://docs.unity3d.com/6000.1/Documentation/ScriptReference/Rendering.GraphicsFence.html

so unity CommandBuffer is a list of instructions to execute for some back end api

and can be runed multiple times but constructing or chnaging the CommandBuffer has a small a cost

cNori:
the Finish Commands is more like a hack u can just merge it and depreciate it letter if there will be a better way of actually reading the buffer implemented

PR #3456, opened 2025-05-14
Fix need to scroll when searching in output log window.

### LINKED ISSUE #3442: [Bug] Searching in the output log needs to scroll up to see results
I think this is new to 1.10. If you search in the output log, when you type something in, it stays at the same scroll height even if the search results are few, and the scroll bar disappears, meaning you can't see anything until you roll the scroll wheel.

Image

Tryibion:
Fix #3442


mafiesto4:
Fixed in 057489e5b9cfe213f5b68a8d4e47ca5a9c9a3599.

PR #3457, opened 2025-05-14
Genericized InputOptions, Allows MouseButton, KeyboardKeys, and MouseScroll

transcendtient:
The 171 files is me gitignoring .flax. 

I did hook into the C++ bindings generator to make the KeyboardKeysString and MouseButtonString classes, which might be the most controversial thing.

I really, really, dislike using strings that are not consts, and especially if they are used for logic. This is my way of getting that in there for now. They are used in the defaults for InputOptions,cs.

The GetMouseScrollDelta abstract method in RootControl.cs, I'm not sure. There is probably some property I'm missing that is in RootControl but I didn't find it.

I refactored all the Process to make them much simpler, they do not take a key/mouse anymore. If a key is required it would be much better, in my opinion, to define it as a member and just check it without passing a key. You can get those with the Window or Control context, as well as by passing any override of GetKey, GetMouseButton, and GetMouseScrollDelta.

I also am new to XML comment standards...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ Source/*.csproj
 PackageEditor_Cert.command
 PackageEditor_Cert.bat
 PackagePlatforms_Cert.bat
+*.flax
 
 # User-specific files
 *.suo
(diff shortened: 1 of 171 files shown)


xxSeys1:
Does this by any chance fix #3471?

transcendtient:
It does fix [#3471](https://github.com/FlaxEngine/FlaxEngine/issues/3471) yes.
I'm working on expanding this to a separate namespace and simplifying/genericizing system wide user input.
image

transcendtient:
Working on a fix to respect the maximum input length, it is currently 4, but allows more. When parsed it will silently trim this to 4 while displaying the full number of inputs.

PR #3458, opened 2025-05-14
Animated Model Additions

Swiggies:
Adds struct `ModelBoneNode` to AnimatedModel with some functions that use it, which is to help with getting and setting a series of model bone nodes. This helps avoid calling `OnAnimationUpdated();` repeatedly when setting a series of bones, which can be rather slow since it causes a full loop through the entire mesh to recalculate the bounds.

I also noticed that `UpdateBounds()` is called far more often than it should be, so I've made it conditional on UpdateWhenOffscreen to mirror the same functionality in other engines. This helps speed up rendering AnimatedModels quite a lot. If users are experiencing flickering they can change the bounds scale of the Actor.

Most of this is to help facilitate a future PR for physics bones I've been working on but figured these specific additions/fixes can be a seperate PR.

### DIFF
--- a/Source/Engine/Level/Actors/AnimatedModel.h
+++ b/Source/Engine/Level/Actors/AnimatedModel.h
@@ -9,14 +9,6 @@
 #include "Engine/Renderer/DrawCall.h"
 #include "Engine/Core/Delegate.h"
 
-API_STRUCT() struct ModelBoneNode
-{
-    DECLARE_SCRIPTING_TYPE_MINIMAL(ModelBoneNode);
-
-    API_FIELD() uint32 NodeIndex;
-    API_FIELD() Matrix NodeMatrix;
-};
-
 /// <summary>
 /// Performs an animation and renders a skinned model.
 /// </summary>
@@ -25,6 +17,24 @@ class FLAXENGINE_API AnimatedModel : public ModelInstanceActor
 {
     DECLARE_SCENE_OBJECT(AnimatedModel);
     friend class AnimationsSystem;
+
+    /// <summary>
+    /// Keeps the data of a Node and its relevant Transform Matrix together when passing it between functions.
+    /// </summary>
+    API_STRUCT() struct NodeTransformation
+    {
+        DECLARE_SCRIPTING_TYPE_MINIMAL(NodeTransformation);
+
+        /// <summary>
+        /// The index of the node in the node hierarchy.
+        /// </summary>
+        API_FIELD() uint32 NodeIndex;
+
+        /// <summary>
+        /// The transformation matrix of the node
+        /// </summary>
+        API_FIELD() Matrix NodeMatrix;
+    };
 
     /// <summary>
     /// Describes the animation graph updates frequency for the animated model.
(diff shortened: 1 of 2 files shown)


Source/Engine/Level/Actors/AnimatedModel.h
```diff
@@ -9,6 +9,14 @@
 #include "Engine/Renderer/DrawCall.h"
 #include "Engine/Core/Delegate.h"
 
+API_STRUCT() struct ModelBoneNode
```
mafiesto4, line 12: I would name it `NodeTransformation`  and move it inside `AnimatedModel` as nested struct. Then add simple docs comments.

Source/Engine/Level/Actors/AnimatedModel.h
```diff
@@ -9,6 +9,14 @@
...
+{
+    DECLARE_SCRIPTING_TYPE_MINIMAL(ModelBoneNode);
+
+    API_FIELD() uint32 NodeIndex;
```
mafiesto4, line 16: Add doc comments.

Source/Engine/Level/Actors/AnimatedModel.h
```diff
@@ -252,6 +268,14 @@ class FLAXENGINE_API AnimatedModel : public ModelInstanceActor
     /// <param name="worldSpace">True if convert matrices from world-space, otherwise values will be in local-space of the actor.</param>
     API_FUNCTION() void SetNodeTransformation(const StringView& nodeName, const Matrix& nodeTransformation, bool worldSpace = false);
 
+    /// <summary>
+    /// Sets a group of nodes final transformation.
+    /// </summary>
+    /// <param name="modelBoneNodes">Array of the final node transformation matrix.</param>
+    /// <param name="worldSpace">True if convert matrices from world-space, otherwise values will be in local-space of the actor.</param>
+    /// <returns></returns>
+    API_FUNCTION() void SetNodeTransformation(Array<ModelBoneNode>& modelBoneNodes, bool worldSpace = false);
```
mafiesto4, line 277: `const Array<ModelBoneNode>&`

Source/Engine/Level/Actors/AnimatedModel.h
```diff
@@ -236,6 +244,14 @@ class FLAXENGINE_API AnimatedModel : public ModelInstanceActor
     /// <param name="worldSpace">True if convert matrices into world-space, otherwise returned values will be in local-space of the actor.</param>
     API_FUNCTION() void GetNodeTransformation(const StringView& nodeName, API_PARAM(Out) Matrix& nodeTransformation, bool worldSpace = false) const;
 
+    /// <summary>
+    /// Gets the node final transformation for a series of nodes.
+    /// </summary>
+    /// <param name="modelBoneNodes">The series of nodes that will be returned</param>
+    /// <param name="worldSpace">True if convert matrices into world-space, otherwise returned values will be in local-space of the actor.</param>
+    /// <returns></returns>
+    API_FUNCTION() void GetNodeTransformation(API_PARAM(Out) Array<ModelBoneNode>& modelBoneNodes, bool worldSpace = false) const;
```
mafiesto4, line 253: `API_PARAM(Ref)` as it works both ways so bindings will properly convert input array and output result.

Swiggies:
Thank you @mafiesto4 for all the requested changes. I think I've fixed all of them up but won't resolve them myself just in case you want some more.

PR #3482, opened 2025-05-21
EditorViewport refactor

transcendtient:
Depends on #3457 , please review and merge that PR first.
This branch starts at 0af816d647f2fc30faae0ff74d8583f3dfc6ee10 committed on May 19, 2025.
This PR aims to break out discrete pieces of EditorViewport.cs e.g.
FPS counter
Utility methods (ProjectPoint, ConvertMouseToRay, etc...)
CameraWidget
CameraViewpoints
The camera movement persistence (speed, orthographic scale, FOV)
The viewport input (and more complete implementation of gamepad bindings)
Among others...

The initial commit of this PR moves Input to its own namespace, InputConfig (so named to avoid conficts with Flaxengine.Input that is not namespaced. It also removes InputBindingList in favor of a simpler List.
The second commit in this PR has fixes from #3457 and moves some input options in the menus that I believe better reflect their categorization.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ Source/*.csproj
 PackageEditor_Cert.command
 PackageEditor_Cert.bat
 PackagePlatforms_Cert.bat
+*.flax
 
 # User-specific files
 *.suo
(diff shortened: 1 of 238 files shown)


Tryibion:
.Flax files are needed binary files for the editor/engine so they should not be ignored in the gitignore. They also should not be pushed with a PR because they use a separate lfs server and will always fail in a PR because of lack of user access to that server. You will need to remove the file change to the .Flax files from this PR. There are several ways to do it, you can Google and pick your favorite.

transcendtient:
I'll look into it.

PR #3484, opened 2025-05-23
Add parameter to check if the output log command box is focused

### LINKED ISSUE #3483: [REQUEST/ BUG] Make text typed into the Output Log debug console not register in game
Currently every key that is pressed when typing something into the Output Log debug console gets registered in game.

Is is really irritating because it can trigger functionality in game which you don't want and which might even destroy the setup you had for typing a command.

It would be the best if the console could consume every input typed into it.

### Version
1.10

xxSeys1:
A possible solution for #3483. (for github to recognize it: closes #3483)

Allows the user to do something like:
```cs
public Float2 MovementVec => Editor.Instance.Windows.OutputLogWin.IsCommandLineBoxFocused ? Float2.Zero : CalculateMovementVec();
```

This does not really fix the underlying issue, but it provides a way for the user to work around it. Maybe also good to have api in general.

### DIFF
--- a/Source/Editor/Windows/OutputLogWindow.cs
+++ b/Source/Editor/Windows/OutputLogWindow.cs
@@ -446,6 +446,11 @@ namespace FlaxEditor.Windows
             }
         }
 
+        /// <summary>
+        /// Gets a value indicating whether the command line box is focused. Useful for blocking input when typing into the command line box. 
+        /// </summary>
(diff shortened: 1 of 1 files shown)


xxSeys1:
See https://github.com/FlaxEngine/FlaxEngine/issues/3483#issuecomment-2922773506

PR #3486, opened 2025-05-23
Perform layout two times to fix scroll bar not updaing in context menus on search.

### LINKED ISSUE #3470: [BUG] Scroll bar in asset dropdown isn't shown (sometimes) after search is cleared and search cause scrollbar to hide
Does not happen every time, but for example at 0:12 in the video:

https://github.com/user-attachments/assets/ce69354b-810c-497b-9df3-78d578fadd1d

### Version
1.10

Tryibion:
Fix #3470


mafiesto4:
I did more general fix to handle Panel's scroll bars update after layout to avoid need for double layouts in future in other elements :) 1d0a0a409de70502304b6702265c12b17a9d4c07

PR #3487, opened 2025-05-23
Perform layout on tab when selecting the tab.

### LINKED ISSUE #3467: [BUG] Scrollbars don't show when switching back to inactive tab after other tab was resized until the tab is resized
Watch the scrollbars on the "Properties" tab. It's also not possible to scroll with the mouse wheel until resized.

https://github.com/user-attachments/assets/95e73391-0de4-4e71-871e-bf0c016841f0

### Version
1.10

Tryibion:
Fix #3467


mafiesto4:
That was close :) I moved one lone down to fix it in eb90ab84f20bd4bbc960a3c16dbb6bcebc82d10a

PR #3495, opened 2025-05-23
Add option to show vertices and triangles of collision model in a collider data window.

### LINKED ISSUE #3473: [REQUEST] Display some stats for mesh colliders
It would be very useful to know how much vertecies are making up your collision mesh.

The best way to implement this is prolly to do a similar thing to the info toggle button in the mesh editor:

Image

### Version
1.10

Tryibion:
Closes #3473

### DIFF
--- a/Source/Editor/Windows/Assets/CollisionDataWindow.cs
+++ b/Source/Editor/Windows/Assets/CollisionDataWindow.cs
@@ -182,7 +182,14 @@ namespace FlaxEditor.Windows.Assets
         private class CollisionDataPreview : ModelBasePreview
         {
             public bool ShowCollisionData = false;
-            public Model CollisionWireModel;
+            private int _verticesCount = 0;
+            private int _trianglesCount = 0;
+
+            public void SetVerticesAndTriangleCount(int verticesCount, int triangleCount)
+            {
+                _verticesCount = verticesCount;
+                _trianglesCount = triangleCount;
+            }
             
             /// <inheritdoc />
             public CollisionDataPreview(bool useWidgets) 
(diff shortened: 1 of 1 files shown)


mafiesto4:
That's nice but I think it would be more informative to show the stats of the cooked collision data:
* actual triangle count (cooking might use lower LOD as source geo) - see `PhysicsBackend::GetTriangleMeshTriangles` that reads vertex and triangle count from `PxTriangleMesh`
* support conex meshes by showing amount of convex polygons and their vertices count (see `PhysicsBackend::GetConvexMeshTriangles` that reads from `PxConvexMesh`)
* it would be nice to show memory size of collisions but that's more complex (maybe use Chunk[0] size of that asset or estimate via PhysX backend impl)

Tryibion:
> * actual triangle count (cooking might use lower LOD as source geo) - see `PhysicsBackend::GetTriangleMeshTriangles` that reads vertex and triangle count from `PxTriangleMesh`
> * support conex meshes by showing amount of convex polygons and their vertices count (see `PhysicsBackend::GetConvexMeshTriangles` that reads from `PxConvexMesh`)

I am pretty sure that is what this line is returning... <https://github.com/FlaxEngine/FlaxEngine/blob/954cf3eb5b42027d8ad51ab58d1a6cf8235708aa/Source/Editor/Windows/Assets/CollisionDataWindow.cs#L287>

I can directly use those results instead of deconstructing the mesh that is created from those results.

Tryibion:
> * it would be nice to show memory size of collisions but that's more complex (maybe use Chunk[0] size of that asset or estimate via PhysX backend impl)

For this... can I just use the Asset.MemoryUsage on the C# side?

mafiesto4:
> For this... can I just use the Asset.MemoryUsage on the C# side?

Yeah that should be good. So devs can learn how much memory overheads specific collision uses.

PR #3506, opened 2025-05-30
Eye Adaptation Fix

### LINKED ISSUE #3480: [Bug] Eye Adaptation post FX is Inversed
**Issue Description**

Generally, in bright conditions, exposure becomes low and vice versa.
But in Flax, this is the opposite. The exposure becomes high in bright conditions (like looking at a _light source)_ and lower in dimmer conditions, which is not Realistic.

**Steps to reproduce**

1. Create a new project with the ""_Basic_"" template
2. Play the scene and look **at** the sun and then towards the floor.

**Minimum reproducible version**
Any

**Engine version**
Master build 1.10.6075

Would request immediate attention.

TONYflare123:
Closes #3480 
According to https://github.com/FlaxEngine/FlaxEngine/issues/3480 Eye Adaptation was inversed. Here is a fix that should work.

#### Changes
Time tracking is now correct.
All adaptation-related values (brightness, exposure, histogram, speed) are inverted to fix the adaptation direction.
_Some comments were removed for code clarity_


mafiesto4:
It breaks existing Flax Samples projects to be super bright:

image

mafiesto4:
Additionally:
* the fix should happen in shader file (`EyeAdaptation.shader`), as it makes it super confusing if min is swapped with max (don't commit `EyeAdaptation.flax` file)
* don't remove comments about rendering stages, those are useful, code clarity has nothing to do with it
* remove comments related to your internal thinking process: `// Rest of the implementation remains structurally the same...`, `// FIXED: Was incorrectly set to 0.0f`, `// ... (rest of the file remains unchanged)`

TONYflare123:
After a long research, it looks that *default eye adaptation* is actually correct, similiar to what *should* happen in real-life cameras

PR #3522, opened 2025-06-07
Add node alignment formatting options to visject

### LINKED ISSUE #3519: [REQUEST] A button to align Visject Nodes horizontally and vertically (2 seperate buttons) like in Unreal
Title.

This would save me a ton of time I "waste" keeping my nodes clean.
I know the "Format Nodes" option exists, but often times it does too much when I just want to align the selected nodes.

### Version 
1.10

Zode:
This adds vertical and horizontal alignment options to Visject surfaces. The existing option `format node(s)` has been renamed to `auto format` inside the menu so that the original option could be turned into a submenu instead. The alignment options ignore any potential node overlap.

https://github.com/user-attachments/assets/2fb17a34-a6b8-40df-b247-b47921133477

~~It might be worth adding hotkeys for these actions like in Unreal, however i opted to not add them in case #3457 gets PR'd in.~~ went ahead and added hotkeys similar that of Unreal's.

This closes #3519

### DIFF
--- a/Source/Editor/Options/InputOptions.cs
+++ b/Source/Editor/Options/InputOptions.cs
@@ -647,5 +647,45 @@ namespace FlaxEditor.Options
         public InputBinding VisualScriptDebuggerWindow = new InputBinding(KeyboardKeys.None);
 
         #endregion
+
+        #region Node editors
+
+        [DefaultValue(typeof(InputBinding), "Shift+W")]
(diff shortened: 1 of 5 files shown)


Zode:
Also decided to add distribution options:

https://github.com/user-attachments/assets/33d04251-07ee-43b6-bbed-41864c8d240f

PR #3540, opened 2025-06-09
Visject larger input boxes

Zode:
This makes Visject inputs uniform size of 50 pixels (instead of float3s being 20 pixels each etc). It also makes comparison nodes automatically expand to contain the input boxes instead of allowing them to overflow and potentially block the output node.

image
While not perfect, this is a huge improvement and makes using the node graphs less painful.
The only notable issue I could spot is that comparisons don't collapse back to their original size if both inputs are populated, indicating that the `SurfaceNode`'s `ReizeAuto()` for some reason does not ignore invisible children despite having a child visibility check.


xxSeys1:
Superseded by #3866

PR #3545, opened 2025-06-11
Tag editor quick expand

Zode:
Very simple addition to make the tag editor expand all nodes if shift is held while the editor is opened. Makes it bit less painful in larger projects :eyes:


xxSeys1:
👀 This would have saved me a lot of pain when I had to batch assign tags in a similar category to different actors (every tag was in `MyTagA.MyTagB.[...]`)

PR #3552, opened 2025-06-13
Add support to move visject socket connections

xxSeys1:
This pr adds the ability to move all connections to a box to a different box (see video).
For that I refactored visject to allow for multiple `IConnectionInstigator`s.

This pr also fixes doing a connection with a box that already has a connection. Before, the old connection would be cleared and no new one would be created. Now this works as expected.

I am unsure about how I get if the user is pressing the Ctrl key, idk if I chose the right way (see Box.cs L 553 in changed files).

**That edgecase with the bool and node has been fixed :)**

https://github.com/user-attachments/assets/69f8c492-9381-4402-8735-103f067569c1

### TODO

- [x] Fix primary context menu
- [x] Fix edgecase shown in video with bool AND node

### DIFF
--- a/Source/Editor/Surface/VisjectSurface.ContextMenu.cs
+++ b/Source/Editor/Surface/VisjectSurface.ContextMenu.cs
@@ -289,7 +289,7 @@ namespace FlaxEditor.Surface
 
             _cmStartPos = location;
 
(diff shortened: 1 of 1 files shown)


xxSeys1:
Marked as ready for review because it's fundamentally working now. Just need to eliminate that one edge case

Issue #3524, opened 2025-06-07
[ Request ] Increase max points count for material gradient node


MrCapy0:
I am working on a leve that must have a lot of variations. To handle it i'm using *Per Instance Random* and *Color Gradient* in my material.

The gradient node is nice, but it's limited because i can't create as many variations that i want.

Currently i can add 7 points. Maybe this can be increased to 16 or higher ?

![Image](https://github.com/user-attachments/assets/3fe3600c-e7f8-4b9d-beca-65697cf1bb31)
![Image](https://github.com/user-attachments/assets/d0261c8a-4006-4609-93ee-9d709fd3de92)

Version:
Master 1.10

TONYflare123:
You can try using multiple `Color Gradient` nodes and then randomly blend between them.

Set the `per instance random` to a `float2` and use the `x coordinate` for gradients and `y coordinate` to `lerp` between two `color gradient nodes`

PR #3586, opened 2025-07-05
Change color gradiant node stops count 8 to 12

MrCapy0:
Closest to #3524


MrCapy0:
I tried increase to 16 but the node usability is not good for many points.

PR #3587, opened 2025-07-05
Add editor option for build configuration when using cook and run.

Tryibion:
Allows the user to change the build configuration for cook and run in the editor options.


xxSeys1:
Maybe this should be moved to *Interface -> Cook & Run*?
image

PR #3603, opened 2025-07-18
Add `IsSceneActor` method to physics backend. Check if physx actor exists before removing from scene.

Tryibion:
Prevents a disabled terrain from throwing errors in the phyx backend. I dont think  the errors are harmful, so this may just add overhead... but it is an option to do this.


mafiesto4:
I think we could have a better solution for this terrain-issue as that would reduce performance of scenes unloading.

Probably handle it for terrain specifically as it can add/remove terrain patches collision to the scene in some cases (eg. when heightmap gets loaded).

mafiesto4:
Fixed in 8f3b80492ed2d9c44f3366ab0af784b8f1a206a1..

Issue #3615, opened 2025-07-25
Shading is incorrect due to shadow artifact fix in lighting.hlsl


Muzz:
I've noticed the shading is in flax is incorrect, thought i was going crazy, but i finally worked out the issue. 

```
#if !LIGHTING_NO_DIRECTIONAL
    // Reduce shadow mapping artifacts
    shadow.SurfaceShadow *= saturate(NoL * 6.0f - 0.2f) * NoL;
#endif
```

this line here darkens the shading curve, but it's not lambert, this line atm is actually providing ALL the shading math, and it's not the correct physical curve. 

the Diffuse_Lambert function actually doesn't do the lambert math at all, all that function does is energy-conserving normalization.

Here is what i propose as an alternative. 

```

ShadowSample GetShadow(LightData lightData, GBufferSample gBuffer, float4 shadowMask, float NoL)
{
    ShadowSample shadow;
    float rawShadow = shadowMask.r;
    
    // Reduce shadow strength at grazing angles to hide artifacts
    float shadowFade = saturate(NoL * 6.0f - 1.0f);
    shadow.SurfaceShadow = gBuffer.AO * lerp(1.0, rawShadow, shadowFade);
    shadow.TransmissionShadow = shadowMask.g;
    return shadow;
}
```

bias the normal in GetShadow to hide the shading biases, rather than using that hack. 

hook up diffuse properly:
`    lighting.Diffuse = Diffuse_Lambert(diffuseColor) * NoL;
`
This removes the shadow artifacts, and gets us correct shading. 

<img width="1537" height="2857" alt="Image" src="https://github.com/user-attachments/assets/75eacb1b-1a35-4080-a47c-e0eed3306f8b" />

Muzz:
I'll do a pull request once i'm finished testing these changes.

mafiesto4:
I got rid of that `saturate(NoL * 6.0f - 0.2f)` in 6a8553a2777f8dd7fa2ffa9e7ec4db070252c462. I don't recall the exact reason or point of reference used there other than just reducing shadow mapping artifacts by moving shadow towards the light.

Now, it looks far better, thanks for spotting this.

PR #3616, opened 2025-07-25
Lighting fix

Muzz:
https://github.com/FlaxEngine/FlaxEngine/issues/3615

this is a fix that improves the base shading for objects, I added a normal bias hack to remove the shadow artifacts, but conditionally am applying it according to if it's a directional light or not, as the hack doesn't work on point or spotlights.

### DIFF
--- a/Source/Shaders/Lighting.hlsl
+++ b/Source/Shaders/Lighting.hlsl
@@ -6,14 +6,23 @@
 #include "./Flax/LightingCommon.hlsl"
 
 
-ShadowSample GetShadow(LightData lightData, GBufferSample gBuffer, float4 shadowMask, float NoL)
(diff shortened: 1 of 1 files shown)


mafiesto4:
Thanks for looking into this and fixing it!

I'lll provide similar fix but with some changes to:
* handle `LIGHTING_NO_DIRECTIONAL` which is used by volumetric particles to lit smoke/dust without `NoL` shading:
image

* keep applying `NoL` to `SurfaceShadow`, otherwise `Diffuse` color will break subsurface scattering and specular will glow:
image

Muzz:
> Thanks for looking into this and fixing it!
> 
> I'lll provide similar fix but with some changes to:


Yeah I'm sure there was things I missed with the full fix as i only have a partial understanding of the entire renderer, so thanks for looking into this!

mafiesto4:
Fixed in 6a8553a2777f8dd7fa2ffa9e7ec4db070252c462. Thanks!

PR #3622, opened 2025-08-01
Audio events and time slider

Tryibion:
Add `ClipStarted` and `ClipFinished` events to audio source.

Add a slider to the debug editor for the audio source to select time.
image

### DIFF
--- a/Source/Engine/Audio/AudioSource.cpp
+++ b/Source/Engine/Audio/AudioSource.cpp
@@ -168,7 +168,7 @@ void AudioSource::Play()
     else
     {
         // Source was nt properly added to the Audio Backend
-        LOG(Warning, "Cannot play unitialized audio source.");
+        LOG(Warning, "Cannot play uninitialized audio source.");
     }
 }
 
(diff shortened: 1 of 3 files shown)


Source/Engine/Audio/AudioSource.h
```diff
@@ -76,6 +76,16 @@ class FLAXENGINE_API AudioSource : public Actor
     API_FIELD(Attributes="EditorOrder(10), DefaultValue(null), EditorDisplay(\"Audio Source\")")
     AssetReference<AudioClip> Clip;
 
+    /// <summary>
+    /// Event fired when the audio clip starts.
+    /// </summary>
+    API_EVENT() Delegate<> ClipStarted;
```
mafiesto4, line 82: `Action` can be used instead of `Delegate<>`

Source/Engine/Audio/AudioSource.cpp
```diff
@@ -583,3 +588,11 @@ void AudioSource::BeginPlay(SceneBeginData* data)
             SetTime(GetStartTime());
     }
 }
+
+void AudioSource::EndPlay()
+{
+    Actor::EndPlay();
+
+    ClipStarted.UnbindAll();
```
mafiesto4, line 596: While this might be useful to unbind events automatically, I'm not sure if that's a good call .
For object pooling it could help but if someone is using that event then shuld unregister on end manually, just like for other events in the engine.

What do you think?

PR #3296, opened 2025-03-17
Add option to create a Canvas Scaler to the Widget create menu

xxSeys1:
Closes #3272.

image

This also changes the default mode to be `Canvas` instead of `Control`, I made [a poll on discord](https://discord.com/channels/437989205315158016/437989205315158018/1351329542584668160) to figure out what ppl prefer.

If Controls "wins" I will revert that change.
</br>
</br>
</br>
</br>
    
    
    

~~*(Maybe this could be made an editor setting so ppl can set their own default lmao)*~~
Haha guess what happened: Now they can

PR #3628, opened 2025-08-08
Fbx importer improvements

alsed:
_Edit: ~~Moved this PR into a draft~~. Most of the changes done here could also be incorporated into the Assimp importer (with the exception of pre and post-rotation). 
The axis correction should also be incorporated into the import root motion, which is done somewhere else. 
I'm unsure wether keep adding more things until this current proposal is reviewed._ 

- Correct import of FBX files with different coordinate systems (e.g., Z up).
- Correct conversion of FBX meshes with geometric transformation.
- Correct the orientation of the viewport camera. Now preview and scene cameras are matched.
- ~~Prefab is now the default for import models.~~
- An Import Animation option `Bake Rotations` is added. If enabled, pre and post rotation data will be added into the keyframes animations. This can correct the import of animations assets that don't come with prebaked rotations (e.g. Unreal assets).

Before:
Captura desde 2025-08-29 12-30-36

After:
Captura desde 2025-08-30 20-45-02

### DIFF
--- a/Source/Editor/Viewport/Cameras/ViewportCamera.cs
+++ b/Source/Editor/Viewport/Cameras/ViewportCamera.cs
@@ -24,10 +24,10 @@ namespace FlaxEditor.Viewport.Cameras
         private EditorViewport _viewport;
 
         private static Quaternion viewportOrientation = new(0.424f, -0.094f, 0.044f, 0.899f);
-        // private static Quaternion viewportOrientation = Quaternion.Euler(-4, 59, 0);
+        //private static Quaternion viewportOrientation = Quaternion.Euler(-4f, 59f, 0f);
 
         // private static Quaternion arcBallOrientation = new(-0.08f, -0.92f, 0.31f, -0.23f);
-        private static Quaternion arcBallOrientation = Quaternion.Euler(15, -45, 0);
+        private static Quaternion arcBallOrientation = Quaternion.Euler(15f, -45f, 0f);
 
         /// <summary>
         /// Gets the parent viewport.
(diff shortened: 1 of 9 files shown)


Tryibion:
My opinion: I don't use the prefab option that often... so I would prefer it to not be the default.

alsed:
I assumed that prefab is the option that most people need when importing a file.
Of course, this is just a proposal. The change is in a specific commit and is only a few lines of code, so it can be easily canceled/reverted.

xxSeys1:
> My opinion: I don't use the prefab option that often... so I would prefer it to not be the default.

Maybe another thing to put in a "Defaults" category in the editor settings? Like in #3296

alsed:
Additional description to the changes.

Changes to the FBX importer:
- Most of the functions inside the struct "OpenFbxImporterData" where moved outside.
- The code to parse the fbx system coordinate was also moved to its own function and a additional struct was created to ease the process ("Axis System").
- Also, a struct "CoordSystem" was created to interpret the system.
- The quaternion "RootConvertRotation" was replaced by a transformation matrix "To Flax" which is defined inside "CoordSystem".

Changes to the preview cameras:
- The magic values that controlled the camera orientation were centralized in private values "ViewPort Orientation" and "ArcBallOrientation".
- "ArcBallOrientation" was reoriented and expressed in Euler angles.

Notes:
- prefab as default has been reverted as suggested by @Tryibion.
- In case the previewlight needs to be reoriented, this can be done in `AssetsPreview` file.

alsed:
**Update:**

An Import Animation option `Bake Rotations` is added:
If enabled, pre and post rotation data will be calculated into the keyframes animations.
This can correct the import of animations assets that don't come with prebaked rotations (e.g. Unreal assets).

Bake Rotations option

Before:
bef

After:
after

PR #3636, opened 2025-08-13
Add material base ctor for `MaterialBrush`

Tryibion:


mafiesto4:
What about changing the existing one into `MaterialBase` to accept both?

Tryibion:
> What about changing the existing one into `MaterialBase` to accept both?

Done

PR #3648, opened 2025-08-21
Expose hideOnClose and scrollBars in constructor for CustomEditorWindows

Zode:
This exposes the HideOnClose and ScrollBars arguments for CustomEditorWindows, skipping the need to re-implement the exact same logic for editor plugins that require changes to those two.
The constructor uses default values to match previous behavior, so no changes are needed for older scripts.

### DIFF
--- a/Source/Editor/CustomEditorWindow.cs
+++ b/Source/Editor/CustomEditorWindow.cs
@@ -18,8 +18,8 @@ namespace FlaxEditor
             private readonly CustomEditorPresenter _presenter;
             private CustomEditorWindow _customEditor;
 
-            public Win(CustomEditorWindow customEditor)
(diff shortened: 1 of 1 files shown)


Zode:
Noticed has issues when scripts/plugins are reloaded, the custom window no longer opens when using `FocusOrShow()` however it starts working again once layout is reset to default. 

Marking as draft for now.

edit: looks like calling `Show()` directly always works... weird

Zode:
Okay so further investigation: the FocusOrShow() bug affects even windows that do not use CustomEditorWindow, so the issue is elsewhere (most likely in the layout restoring bit). If I end up figuring it out I'll make a separate PR as that bug is replicable even without this PR.

Issue #2266, opened 2024-02-20
Add automatic monitoring of which mesh is static or animated when importing model as prefab


mafiesto4:
When importing model file as prefab we assume all meshes are static, but also import any animations from the source file. The importer should check if mesh uses blend indices and assume it's animated then import it properly.

alsed:
I have been able to make progress in implementing this.
But there is a problem: in `ModelTool`, I have to set prefab to include `ImportDataType:Skeleton`

```
case ModelType::Prefab:
        options.ImportTypes = ImportDataTypes::Geometry | ImportDataTypes::Nodes | ImportDataTypes::Skeleton | ImportDataTypes::Animations;
```

Further on, it is defined that if an Skeleton is imported, one must be created, even if there is no data:

```
    if (EnumHasAnyFlags(options.ImportTypes, ImportDataTypes::Skeleton))
    {
        LOG(Info, "Imported skeleton has {0} bones and {1} nodes", data.Skeleton.Bones.Count(), data.Nodes.Count());

        // Add single node if imported skeleton is empty
        if (data.Skeleton.Nodes.IsEmpty())
        {
            data.Skeleton.Nodes.Resize(1);
            data.Skeleton.Nodes[0].Name = TEXT("Root");
            data.Skeleton.Nodes[0].LocalTransform = Transform::Identity;
            data.Skeleton.Nodes[0].ParentIndex = -1;
        }
```

To implement this correctly, the option to not import Skeleton if there is no data must be created.
But I wanted to ask first, the reason behind creating a node even if there is no skeleton.
And perhaps this implies that if a Skinned Model is imported without a Skeleton, this should be reverted to import a Static Model.

mafiesto4:
@alsed Sure, let import backend (OpenFBX/Assimp) import skeleton and nodes for Prefab import, then decide if any model meshes uses it to import as Skinned Model.

mafiesto4:
Added in #3669

PR #3669, opened 2025-09-08
Add automatic recognition of Skinned Models for Prefabs

alsed:
Implements #2266

Changes:
- Prefab automatically detects if the model contains Blendweigths (or BlendShapes) and imports the mesh as a Skinned Model.
- If a model without Bones is imported as a Skinned Model, it will automatically "fallback" as (Static) Model.

Captura desde 2025-09-08 12-52-02

- Backend:
    - Added a `bool IsSkinned` to the Prefab struct to handle the import of `AnimatedModels` in prefabs.
    - In `ImportModel`, the system checks whether the model has Bones; if not, it changes the `ModelType` from Skinned Model to (Static) Model.
    - Additional validation conditions to `ImportOptions Skeleton` to avoid unnecesary generation of skeleton elements.
    - Update for BlendShape remapping in `ModelTool`: only valid blendshapes are added (instead of copying all and later removing the empty ones).

Known Limitations:
- BlendShape detection only checks the first mesh in the file. Ideally, this should be evaluated per mesh, with import options set individually. Currently, only Prefab supports per-mesh evaluation; the rest of the system applies options globally.

### DIFF
--- a/Source/Engine/Tools/ModelTool/ModelTool.cpp
+++ b/Source/Engine/Tools/ModelTool/ModelTool.cpp
@@ -1119,6 +1119,12 @@ bool ModelTool::ImportModel(const String& path, ModelData& data, Options& option
             options.ImportTypes |= ImportDataTypes::Textures;
         break;
     case ModelType::SkinnedModel:
+        if (!data.Skeleton.Bones.HasItems())
+        {
+            LOG(Warning, "Model is not Skinned, it will be imported as Static");
+            options.ImportTypes = ImportDataTypes::Geometry | ImportDataTypes::Nodes;
+            options.Type = ModelType::Model;
+        }
         options.ImportTypes = ImportDataTypes::Geometry | ImportDataTypes::Nodes | ImportDataTypes::Skeleton;
         if (options.ImportMaterials)
             options.ImportTypes |= ImportDataTypes::Materials;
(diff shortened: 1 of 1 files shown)


Tryibion:
I believe, although I could be wrong, the old logic allowed importing a static mesh as a skinned mesh so that way animations could be imported that just affect a static mesh and could be played in the engine. Hence the single root bone being created when it is imported.

alsed:
> I believe, although I could be wrong, the old logic allowed importing a static mesh as a skinned mesh so that way animations could be imported that just affect a static mesh and could be played in the engine. Hence the single root bone being created when it is imported.

You're right. The main problem that I had implementing this was that any model imported with the `Skeleton` import flag generates Bones, Bone Nodes, and BlendIndices/BlendWeights for the meshes, even when none existed. Thus, making it impossible in Prefab to distinguish between Static and Skinned models. 
(Prefab currently solves this by not using the import flag Skeleton, so every prefab is static by default).

Researching this, also found that Source engine keeps a root Bone for **every model**:

https://developer.valvesoftware.com/wiki/Skeleton

> In Source [engine], every model has a hidden, implicit root bone, which positions and orients the model relative to its model entity. This special bone is created by studiomdl during MDL compilation and serves as the boundary between the model's internal structure and its location in the world. The model is not aware of its implicit root bone; ergo, this bone cannot be animated by a model's $sequences.

So, this is a design decision.

I went for a minimal change approach. but the assumption is: **_if it has bones or blendshapes is skinned, otherwise is not._**

PR #3681, opened 2025-09-16
Add TIFF Support for Linux (and Mac)

alsed:
Basic import for TIFF images, for Linux and Mac, has been implemented.
`libtiff `library is used for reading.

Other:
- In `TextureTool.Build`, check for tinyexr and astc dependency has been merged since both used the same first condition.

Notes:
- The `libtiff` library dependency can be built by running: 
    `Flax.Build -log -BuildDeps -depsToBuild=libtiff -verbose`
- Build dependency script for Mac has not been tested.

### DIFF
--- a/Source/Engine/Tools/TextureTool/TextureTool.Build.cs
+++ b/Source/Engine/Tools/TextureTool/TextureTool.Build.cs
@@ -21,7 +21,6 @@ public class TextureTool : EngineModule
 
         bool useDirectXTex = false;
         bool useStb = false;
-        bool useExr = options.Target.IsEditor;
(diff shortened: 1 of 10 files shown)


mafiesto4:
I tried to build that lib on macOS and it fails on some errors within `/configure`:
```
...
checking whether ln -s works... yes
checking for cmake... /usr/local/bin/cmake
checking if LD -Wl,--version-script works... no
./configure: line 5891: syntax error near unexpected token `win32-dll'
./configure: line 5891: `LT_INIT(win32-dll)'
```

PR #3684, opened 2025-09-16
Check if control is control meant for gui editor use in "Create Widget" dialog 

xxSeys1:
This pr fixes a bug where the user is able to create a widget with a base control of *any* class that is a control, even ones that are hidden in the gui editor.

*Before:*
image

*After:*
image

### DIFF
--- a/Source/Editor/Content/Create/PrefabCreateEntry.cs
+++ b/Source/Editor/Content/Create/PrefabCreateEntry.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Wojciech Figat. All rights reserved.
 
 using System;
+using System.Linq;
 using FlaxEditor.Scripting;
 using FlaxEngine;
 using FlaxEngine.GUI;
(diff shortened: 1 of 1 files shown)


Source/Editor/Content/Create/PrefabCreateEntry.cs
```diff
@@ -117,7 +118,8 @@ public enum WidgetMode
 
             private static bool IsValid(Type type)
             {
-                return (type.IsPublic || type.IsNestedPublic) && !type.IsAbstract && !type.IsGenericType;
+                var controlTypes = Editor.Instance.CodeEditing.Controls.Get();
+                return (type.IsPublic || type.IsNestedPublic) && !type.IsAbstract && !type.IsGenericType && controlTypes.Any(c => c.Type == type);
```
mafiesto4, line 122: UI is C#-only so we can ignore Script Type and check for C# controls quicker via:
```cs
controlTypes.Contains(new ScriptType(type))
```

mafiesto4 (approved):

PR #3687, opened 2025-09-18
Don't clear content panel folder- and scene tree search box on script reload and play mode enter/ exit

xxSeys1:
Idea is just to re- perform search whenever script reload ends or play mode is entered/ exited.

Additional `PerformLayout()` calls where added because to minimize (minimize a lot) this happening:

Screenshot 2025-09-18 221930

### DIFF
--- a/Source/Editor/Windows/SceneTreeWindow.cs
+++ b/Source/Editor/Windows/SceneTreeWindow.cs
@@ -603,7 +603,7 @@ namespace FlaxEditor.Windows
             _dragHandlers = null;
             _tree = null;
             _searchBox = null;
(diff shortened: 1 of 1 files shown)


Source/Editor/Windows/SceneTreeWindow.cs
```diff
@@ -599,6 +603,7 @@ public override void OnDestroy()
             _dragHandlers = null;
             _tree = null;
             _searchBox = null;
+            ScriptsBuilder.ScriptsReloadEnd += OnSearchBoxTextChanged;
```
GoaLitiuM, line 606: Typo

PR #3688, opened 2025-09-19
Make particle age/ lifetime nodes more clear

### LINKED ISSUE #3308: Add some keywords to particle lifetime and age in emmiter visject search field to make it more intuitive
**Issue description:**

image

I've seen people searching for a normalized lifetime node a few times (myself included), and they never find it on the first go because it is named "Particle Normalized Age" (so it does not show up in the search menu when you type in "Lifetime").

I *think* there is some custom filtering for that menu, so I suggest adding the keyword "Lifetime" to the normalized age node, and maybe add the keyword "Age" to the Particle Lifetime node.

**Steps to reproduce:**

see *Issue Description*

none

**Flax version:**

latest launcher master as of 3/23/2025

xxSeys1:
**You can find the changes made listed out below, feel free to give me your opinions on them :)**

This pr aims to make particle age/ lifetime related nodes more clear in what they do and what differentiates them (for reasons look at #3308).

It also adjusts the size of the nodes to make them more visually pleasing.

There are some new alternative titles/ names for the nodes, so that every node can be found just by searching "Age" or "Lifetime", which should further minimize confusion, since all options are immediately shown, and the new names and descriptions are more clear.

*New:*
image

### Total Lifetime
Previous name: `Particle Lifetime`
Previous description: `Particle lifetime (in seconds).`
New name: `Total Lifetime`
New description: `Total particle lifetime (in seconds), assigned when the particle is created. Always the same, no matter the particles age.`

### Particle Age
Previous description: `Particle age (in seconds).`
New description: `Particle age (in seconds). How long the particle has been alive since it was created.`

### Particle Normalized Age
Previous description: `Particle normalized age to range 0-1 (age divided by lifetime).`
New description: `Particle normalized age represented as from 0 (max lifetime) - 1 (max age) (age divided by lifetime).`

In case you want to see what the nodes to for yourself, you can create a new "Constant Burst" particle emitter and modify it like this:
image

Now connect the different particle lifetime/ age nodes to the color ramp and set the lifetime to something between 0 - 1.
Or you download that emitter here:
[LifetimeStuff.zip](https://github.com/user-attachments/files/22429499/LifetimeStuff.zip)

Closes #3308.

### DIFF
--- a/Source/Editor/Surface/Archetypes/Particles.cs
+++ b/Source/Editor/Surface/Archetypes/Particles.cs
@@ -436,10 +436,11 @@ namespace FlaxEditor.Surface.Archetypes
             new NodeArchetype
             {
                 TypeID = 102,
-                Title = "Particle Lifetime",
-                Description = "Particle lifetime (in seconds).",
+                Title = "Total Lifetime",
+                Description = "Total particle lifetime (in seconds), assigned when the particle is created. Always the same, no matter the particles age.",
+                AlternativeTitles = new[] { "Age" },
                 Flags = NodeFlags.MaterialGraph | NodeFlags.ParticleEmitterGraph,
-                Size = new Float2(200, 30),
+                Size = new Float2(180, 30),
                 Elements = new[]
                 {
                     NodeElementArchetype.Factory.Output(0, string.Empty, typeof(float), 0),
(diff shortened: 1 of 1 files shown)


mafiesto4:
It does make sense to improve those descriptions. Regarding `Particle Lifetime` change I'm not so such tbh as that can vary per-particle so every one can have different lifetime - that's why it has `Particle` prefix as it's attribute of it.

xxSeys1:
> It does make sense to improve those descriptions. Regarding `Particle Lifetime` change I'm not so such tbh as that can vary per-particle so every one can have different lifetime - that's why it has `Particle` prefix as it's attribute of it.

Good point, makes a lot of sense.
I re- added the "Particle" prefix in https://github.com/FlaxEngine/FlaxEngine/pull/3688/commits/0917a743cd8d91ca56eadf710f538dc81b146f34

PR #3715, opened 2025-10-05
Add Missing functions on Vector4 (C++)

ThePhantomMask:
I noticed that Vector4 on the C++ side lacks Normalise functions that other vectors have, so I added them.


ThePhantomMask:
Add Normalised also on the C# side, as it wasn't there.

ThePhantomMask:
I found what was causing the error that was causing other non-Windows systems to fail. With b5b8f32 I fixed it. If the test is redone, all 11 should pass. @mafiesto4

mafiesto4:
Merged in 834c4553b22d2d8ec771293f5edf8cea9ba0fbed as a squashed commit. Sorry abut that way but the PR history was long for such change :)

PR #3735, opened 2025-10-15
Fix not setting value on box collider auto resize.

### LINKED ISSUE #3670: Box Collider "Resize to Fit" does not enable the save button for a prefab.
As title says, Clicking the button to Resize a box collider to fit the bounding box of it's parent, the save button does not become enabled, requiring an extra manual change to be able to save the prefab.

Tryibion:
Fix #3670


mafiesto4:
Simpler way done in 72bb2dd932310ef4e3b5e15620a472cb5b156047

PR #3752, opened 2025-10-19
Enforce pointer alignment for InlinedAllocation

GoaLitiuM:
AssetReferences stored in inlined allocation needs to be aligned to pointer sized boundary due to atomic operations having strict requirements for such. Unaligned access seems to only crash on Windows on ARM systems when trying to allocate TextRender draw chunks.

I'm not sure if this is the right way to fix this but it seems to fix it in this case.


GoaLitiuM:
This can be verified by adding additional checks for alignment right before `_InterlockedCompareExchange64` calls where the crash occurs.

Issue #2646, opened 2024-05-24
Add Video backend for Linux


mafiesto4:
Add missing video backend for Linux platform. Find a good lib or 3rd party plugin for `.mp4` files loading (eg. [minimp4](https://github.com/lieff/minimp4)), then decode `h.254` video frames into `YUV` format (engine can automatically convert into `RGB` image). Use one of the existing backends as base for proper implementation: https://github.com/FlaxEngine/FlaxEngine/tree/1.9/Source/Engine/Video

All other platforms provide system-lib for video media playback so only Linux is missing.

Video docs: https://github.com/FlaxEngine/FlaxDocs/blob/1.9/manual/graphics/video/index.md

Here is a prototype with minimp4 that loads data from a file but requires frames decoding: [PrototypeMinimp4.zip](https://github.com/FlaxEngine/FlaxEngine/files/15433208/PrototypeMinimp4.zip)

mafiesto4:
> Could be dynamically linked so that it uses system binaries for the job, so it can use "nonfree" decoders (such as nvidia's gpu decoding).

I think it would be a good starting point. Maybe we could use `ffmpeg`?

Zode:
> > Could be dynamically linked so that it uses system binaries for the job, so it can use "nonfree" decoders (such as nvidia's gpu decoding).
> 
> I think it would be a good starting point. Maybe we could use `ffmpeg`?

FFmpeg is bit harder as it provides a low level API from what i gather, GStreamer provides a high level API similar to the MFPlay API used on Windows side of things so it would be way simpler to integrate (and more importantly: maintain)

alsed:
I have made a simple mp4 player, trying to solve this problem. I just uploaded to github:

https://github.com/alsed/simple-mp4-player

It uses minimp4 as demuxer, openh264 as decoder, (and SDL2 to play the frames since it takes directly the YUV format).
The first part of the code is a copy of minimp4, then comes the decoding, where the hardest part was to use correctly the stride data.
The decoder is slow, but optimizations can be made. I left the code simple on purpose.
I hope you find it useful @mafiesto4

TyGrze:
Whats the state of this?
Do we need someone new to take over implementing this as the other PR has failed automation and hast been touched in 8-9 months?
I can but dont want to step on any feet.

PR #3756, opened 2025-10-24
Add video backend for Linux

alsed:
- implements #2646
- Uses the `minimp4` and `openh264` libraries for demuxing and decoding video.
- `fdk-aac` library is implemented to decode audio. See license issue below.

Other technical stuff:
- added weight: ~5MB
- player AudioSource is now created during player creation, instead on UpdateAudioBuffer().
- player AudioSource TaskGraph destruction is also added.

To Build:
- The video libraries dependencies can be built by running:
    `Flax.Build -log -BuildDeps -depsToBuild=openh264`

License issue:
- This implementation uses `fdk-aac` for audio decoding. This library isn’t free so is disabled by default.
- Please consider the possibility that Flax could get a special license for use. 
- Otherwise, another audio library should be implemented (e.g. FAAC).

### DIFF
--- a/Source/Engine/Video/Mini/VideoBackendMini.cpp
+++ b/Source/Engine/Video/Mini/VideoBackendMini.cpp
@@ -1,0 +1,683 @@
+// Copyright (c) 2012-2024 Wojciech Figat. All rights reserved.
+
+#if VIDEO_API_MINI
+
+#include <vector>
+#include <queue>
+#include "VideoBackendMini.h"
+#include "Engine/Profiler/ProfilerCPU.h"
+#include "Engine/Threading/TaskGraph.h"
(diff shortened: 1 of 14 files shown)


TyGrze:
Is this pull request abandoned?
Because I was looking into adding video backed to Linux as well and was wondering if this was ever going to be completed or if I could start looking into supporting ffmpeg libraries like `libavutil`, `libavcodec`, `libswscale`, and `libswresample` to get demuxing and decoding working on Linux

alsed:
No, this pull request has not been abandoned, and it is not true that it's incomplete. It works fine.
It's currently awaiting review, but that decision, and its eventual implementation, don't depend on me.
You can ask the developers if they would prefer a different implementation.

PR #3758, opened 2025-10-24
Better rotation on UI handles which handles total parent rotation.

Tryibion:
Small UI adjustment to GUI handles. Allows the handle to handle all of the parent rotations instead of just the immediate parent rotations to correctly rotate the handle rect.

Image of handling rotations:
image

This was a little hacky, but seems to work pretty well.


xxSeys1:
The original reason for moving the ui drag handles outside of the bounds of the control where that they covered parts of the control, especially when the control is smaller.

Maybe Flax should do something like Godot, where the handles are *just* outside of the control bounds:
image

I think this strikes a really good balance between looking good while at the same time not covering up any area of the selected control.

PR #3787, opened 2025-11-01
Add alternative node titles to pack/unpack nodes

xxSeys1:
Now every pack/ unpack node can be found by searching:
"Pack":
- Make
- Construct
- Compose

"Unpack":
-  Break
- Deconstruct
- Decompose
- Split

If there is any commonly used terminology for "pack"/ "unpack" that I have not added yet, please let me know and I will add it.

These are the names used in other engines that I could find.
I think this makes Flax more approachable and easy to switch to for new users, since they prolly have used another tool like Unreal Engine or Godot before, which all use different terminology for this operation.

Another reason as to why I think this is useful is because I always mess up what tool uses which terminology. With this pr I can just type in whatever and it will show me the correct nodes.

image
image
image

image
image
image

### DIFF
--- a/Source/Editor/Surface/Archetypes/Layers.cs
+++ b/Source/Editor/Surface/Archetypes/Layers.cs
@@ -149,7 +149,7 @@ namespace FlaxEditor.Surface.Archetypes
                 TypeID = 7,
                 Title = "Unpack Material Layer",
                 Description = "Unpack material properties",
-                AlternativeTitles = new[] { "Break Material Layers", "Deconstruct Material Layers", "Decompose Material Layers" },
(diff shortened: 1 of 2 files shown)


MrCapy0:
I think add "split" can be very nice for unpack nodes because Unity use this on material graph

TONYflare123:
Maybe differentiate the `System` and `FlaxEngine` based Pack and Unpack nodes?

Possible syntax:
Pack Float2 (FlaxEngine)
Pack Float2 (System)

xxSeys1:
> Maybe differentiate the `System` and `FlaxEngine` based Pack and Unpack nodes?
> 
> Possible syntax: Pack Float2 (FlaxEngine) Pack Float2 (System)

Are there even two sets of Pack nodes? I have only ever seen one.

Issue #3813, opened 2025-11-22
[Bug]: Several animation slot bugs


Bits360:
### Description

1: Attempting to replay a finished animation slot without calling ResetAnimation() is impossible.
2: Attempting to change the animation in an animation slot without it finishing on its own is impossible.
3: Not a bug and out of the scope of this report but really weird, StopSlotAnimation requires an animation input but only one animation can play in the slot at a time making it a seemingly useless parameter.

Minimal reproduction project:
[AnimBug.zip](https://github.com/user-attachments/files/23687072/AnimBug.zip)

Q and W play 2 different animation without calling ResetAnimation() (Will not work after the first time)
A and D do the same with ResetAnimation() (If you try to interrupt one animation with another, it triggers the second bug listed)

### Steps to reproduce

1: Play animation in an animation slot
2: Attempt to do so again(It wont work)

Expected behavior: When playing an animation more than once, it will work. When playing a different/the same animation, it will cut off whatever was playing, and play that animation instead.

### Version

1.11

### Relevant logs

```shell

```

Tryibion:
As of right now the work flow is to stop the currently playing animation before playing a new one in the slot.

Something like this will work once #3816 is merged so you dont need the animation parameter in the isplaying and stop methods
```
if (CurrentModel.IsPlayingSlotAnimation(AnimSlot))
{
     CurrentModel.StopSlotAnimation(AnimSlot);
}

CurrentModel.PlaySlotAnimation(AnimSlot, Anim1, 1f, 0, 0);
```

Doing this will blend between the currently playing animation and next one.

PR #3816, opened 2025-11-23
Fix anim slot replay

Tryibion:
Fix part of #3813

And add Stop slot and is playing slot method without anim param.


mafiesto4:
What about just changing default value of that `anim` to `nullptr` and in that case, pausing any animation for the given slot name? For simplicity of fewer APIs :)

Tryibion:
> What about just changing default value of that `anim` to `nullptr` and in that case, pausing any animation for the given slot name? For simplicity of fewer APIs :)

Done.

Issue #3824, opened 2025-11-25
[Request]: Improve hash function of GUID


VitaminCpp:
### Description

Currently `Guid::GetHash()` is implemented like this:
```cpp
inline uint32 GetHash(const Guid& key)
{
    return static_cast<int32>(key.A ^ key.B ^ key.C ^ key.D);
}
```

The simple XOR of A^B^C^D allows many correlations (e.g., identical upper halves) and can concentrate keys in a few buckets in a hash map/set (like HashSet).
By using for example an Avalanche mixer ensures that even small bit differences in the 128-bit GUID affect all output bits. This reduces clustering in weakly distributed or correlated GUIDs, reduces collisions in hash sets/dictionaries and also improves performance a bit.

An example implementation:
```cpp
inline uint32 GetHash(const Guid& key)
{
    // Mix 128 bits down to 32 using a simple avalanche (variant of xxHash32 finalizer)
    uint32 h = key.A ^ key.B ^ key.C ^ key.D;
    h ^= h >> 15;
    h *= 0x2c1b3c6d;
    h ^= h >> 12;
    h *= 0x297a2d39;
    h ^= h >> 15;
    return h;
}
```

**NOTICE:** Keep in mind that this change also requires a reimport/rebuild of the existing .flax assets!

### Benefits

- Performance
- Stability

PR #3825, opened 2025-11-25
Fixed HashSet compact rehash under heavy collisions

VitaminCpp:
- Compact now iterates over the old bucket array using the saved oldSize, and frees with that size, avoiding out-of-bounds when _size changes.
- If reinsertion finds no free slot during compaction (pathological collisions), the table grows once and retries, preventing AVs.
- This fix addresses problems with weak hash keys (like #3824).


VitaminCpp:
Test scenario:
1. Checkout ArizonaFrameworkSample.
2. Start editor as host and connect with a client.
3. Disconnect on client.
4. Reconnect on client.
5. Goto step 3. Repeat ~70 times.

PR #3833, opened 2025-11-30
Add ui edit mode to edit game window and remove from game window

Tryibion:
This will remove the ability to edit UI from the game window and adds the option to edit the UI in the editor window using the GUI editor. The size of the UI editor background will match the resolution selected in the game window.

image

Bugs:
- Seems like Canvas scalar or panel with no colors under UICanvas can not be selected. Maybe something with the raycasting logic?

- Free Aspect and Aspect Ratio game viewport options give not great results for the ui editor view. They will scale the background of the UI editor to be the same size as the game viewport. I am open to suggestions on how to handle this. We could tie this into its own scaling options, but I thought it would be nice to set both in 1 spot.

### DIFF
--- a/Source/Editor/Gizmo/UIEditorGizmo.cs
+++ b/Source/Editor/Gizmo/UIEditorGizmo.cs
@@ -116,7 +116,7 @@ namespace FlaxEditor
                 if (RayCastChildren(ref location, out hit))
                     return true;
 
-                // Check external roots
+                // Check external root
                 var uiRoot = (UIEditorRoot)Parent;
                 if (uiRoot.ExternalRoot != null)
                 {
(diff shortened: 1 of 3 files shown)


antonieoqq:
In my humble opinion, when editing a game's runtime UI, developers typically start by designing for a single aspect ratio—such as 16:9—and only after the layout is mostly finalized do they check how it behaves under other aspect ratios (e.g., 16:10, 21:9, etc.). Since the UI is usually divided into several distinct modules, fine-tuning the anchoring rules for each module can be deferred until the final stage.

It might be helpful to allow the UI editor view to toggle between two modes: by default, its background size matches the game viewport; however, if needed, users could enable an "override" mode, which would let them directly set the background size of the UI editor view independently of the game viewport.

PR #3838, opened 2025-12-04
Add RoundedRectangle and Circle to Render2D.

Tryibion:
Adds Draw and Fill methods for RoundedRectanges and Circles in the Render2D class.

RoundedRectangle methods allow for individual corners as well.


mafiesto4:
Cool! I'll try to think about ideas to remain the current vertex format without adding new data elements there.

Tryibion:
Yeah I tried to combine them in a couple different ways into the existing data structure, but I was getting a lot of artifacts, so I ended up adding an element there.

PR #3846, opened 2025-12-10
Add "Script obsolete" and "Script only in prefab instance" buttons to the script header bar

### LINKED ISSUE #3842: [Request]: Different style for scripts added to prefabs in the scene
### Description

Currently, all scripts have the same style (color, icon, font, etc).

This script added to this actor does not belong to the prefab asset, was added only for this actor on this scene.
Image

There is no way to see it, i need to click on "view changes" to know which script was added.

### Benefits

Will be easy identify what script was added / not belong to the prefab asset.

### LINKED ISSUE #3845: [Request]: Mark scripts that are `[Obsolete]` in Properties Panel
### Description

Currently there is no visual indicator for when you are using a script marked as deprecated (`[Obsolete]`) in Flax.

For example this icon (from the Icon Atlas) next to the script title in the Properties Panel script panel:

Image

### Benefits

This would be very helpful when finding scripts after marking them as deprecated and it would warn the user of using out of support stuff.

xxSeys1:
```cs
using System;
using FlaxEngine;

namespace Game;

[Obsolete("The message", DiagnosticId = "This is the ID")]
public class ObsoleteScript : Script
{
}
```

results in:

image

image

Closes #3845 (sorry for opening an issue first, I thought I wouldn't have the time to do PRs for the time being)

**EDIT:**
Now also closes #3842

image

### DIFF
--- a/Source/Editor/CustomEditors/Dedicated/ScriptsEditor.cs
+++ b/Source/Editor/CustomEditors/Dedicated/ScriptsEditor.cs
@@ -852,7 +852,7 @@ namespace FlaxEditor.CustomEditors.Dedicated
                 var title = Utilities.Utils.GetPropertyNameUI(scriptType.Name);
                 var group = layout.Group(title, editor);
                 if (!hasAllRequirements)
-                    group.Panel.HeaderTextColor = FlaxEngine.GUI.Style.Current.Statusbar.Failed;
+                    group.Panel.HeaderTextColor = _style.Statusbar.Failed;
                 if ((Presenter.Features & FeatureFlags.CacheExpandedGroups) != 0)
                 {
                     if (Editor.Instance.ProjectCache.IsGroupToggled(title))
@@ -865,9 +865,10 @@ namespace FlaxEditor.CustomEditors.Dedicated
                     group.Panel.Open();
 
                 // Customize
+                float totalHeaderButtonsOffset = 0f;
                 group.Panel.TooltipText = Editor.Instance.CodeDocs.GetTooltip(scriptType);
                 if (script.HasPrefabLink)
-                    group.Panel.HeaderTextColor = FlaxEngine.GUI.Style.Current.ProgressNormal;
+                    group.Panel.HeaderTextColor = _style.ProgressNormal;
 
                 // Add toggle button to the group
                 var headerHeight = group.Panel.HeaderHeight;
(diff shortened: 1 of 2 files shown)


Source/Editor/CustomEditors/Dedicated/ActorEditor.cs
```diff
@@ -115,7 +115,7 @@ public override void Initialize(LayoutElementsContainer layout)
                 {
                     if (actor != null)
                         group.Panel.TooltipText = Surface.SurfaceUtils.GetVisualScriptTypeDescription(TypeUtils.GetObjectType(actor));
-                    var settingsButton = group.AddSettingsButton();
+                    var settingsButton = group.AddHeaderButton("Settings.", 0.0f, FlaxEngine.GUI.Style.Current.Settings);
```
mafiesto4: It's nit-picking but Editor tooltips that as very very short (eg. `Undo` or `Pick something`) don't go with a dot in the end as they not a full sentence so just `Settings` is enough :) At least, that was implemented based on Visual Studio tooltips.

Source/Editor/CustomEditors/Dedicated/ScriptsEditor.cs
```diff
@@ -905,11 +907,30 @@ public override void Initialize(LayoutElementsContainer layout)
...
+                var settingsButton = group.AddHeaderButton("Settings.", 0.0f, _style.Settings);
                 settingsButton.Tag = script;
                 settingsButton.Clicked += OnSettingsButtonClicked;
 
-                group.Panel.HeaderTextMargin = new Margin(scriptDrag.Right - 12, 15, 2, 2);
+                // Add script obsolete icon to the group
+                if (scriptType.HasAttribute(typeof(ObsoleteAttribute), false))
+                {
+                    ObsoleteAttribute attribute = null;
+                    foreach (var e in scriptType.GetAttributes(false))
```
mafiesto4: I would simplify the loop with a LINQ expression:
```cs
var attribute = (ObsoleteAttribute)scriptType.GetAttributes(false).First(x => x is ObsoleteAttribute);
```

Source/Editor/CustomEditors/Elements/Container/GroupElement.cs
```diff
@@ -40,21 +40,21 @@ public class GroupElement : LayoutElementsContainer
         /// Adds utility settings button to the group header.
         /// </summary>
         /// <returns>The created control.</returns>
-        public Image AddSettingsButton()
```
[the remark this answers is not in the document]
xxSeys1: Done in 0d8ac5b7acef060d6c880c311972a2a69d9518c5

Source/Editor/CustomEditors/Dedicated/ActorEditor.cs
xxSeys1: Done in 0d8ac5b7acef060d6c880c311972a2a69d9518c5

Source/Editor/CustomEditors/Dedicated/ScriptsEditor.cs
xxSeys1: Done in 0d8ac5b7acef060d6c880c311972a2a69d9518c5

Thanks for the review :)

PR #3856, opened 2025-12-15
Use standard output for logging by default

GoaLitiuM:
For Unix platforms we should assume the engine is launched from terminal or through a debugger so we wont miss any important output without the `-std` parameter.


mafiesto4:
Maybe do it like this for Debug/Development builds only? Release should be rather slim even if it uses logging

GoaLitiuM:
Changed the behavior to show standard output by default in non-release builds.

PR #3863, opened 2025-12-17
Fix scaling contains point location in CanvasScalar.

Tryibion:
Fixes an issue when just using the aspect ratio options in the game viewport.


Tryibion:
Nevermind, while this fixes the selection in the gui editor, it breaks the selection when playing the game. I will think of a better fix.

PR #3885, opened 2026-01-02
Add C++ preprocessor definitions `FLAX_EDITOR` and `FLAX_GAME`

GoaLitiuM:
Fixes inconsistent preprocessor definitions for both editor and game builds for C++ scripts. For C# scripts these definitions are present and supported in both editor and game builds respectively, but for C++ scripting user is expected to use the `USE_EDITOR` definition over these ones.


Tryibion:
We should probably settle on one the use and remove the other come 2.0

mafiesto4:
Yeah good idea. We should unify those (and update docs too).

(project auto-upgrader for 2.0 could do text search and replace in game code/plugins)

PR #3920, opened 2026-02-04
Add some Visject connection tooling (lazy connect, rip node from connections, slot node in existing connection)

### LINKED ISSUE #2998: [Proposal] Visject: Lazy Connect
Blender's Node Wrangler addon has a nifty feature that Visject could benefit from adapting.
https://docs.blender.org/manual/en/latest/addons/node/node_wrangler.html#lazy-connect

Instead of dragging one socket to another, you can drag one node to another.
This is especially useful when working with larger graphs as selecting sockets when zoomed out can be quite finicky.

If the source node has multiple output sockets or if the destination node has multiple input sockets, then a dropdown would be shown, allowing the user to choose the desired socket(s). Otherwise the sole socket would be used.

Not sure which shortcut would be best for this.

xxSeys1:
closes #2998

This was requested on discord. Also Blender has this with the Node Wrangler Addon, and I see it being used very often.

This pr adds the following tools for connecting and disconnecting nodes in Visject:

### Lazy connect

https://github.com/user-attachments/assets/981b1450-54ad-47f1-bc64-bbde8509bce5

To start lazy connecting, hold down the ALT key and drag the mouse while holding down the right mouse button.
If the lazy connect start and/ or end are directly over a node, it will pick those nodes as the start and end nodes.

If there are no nodes directly under start and/ or end, it will find the nearest node to the start and end position (`GetClosestNodeAtLocation()`). This right now will "only" check for node centers, but in my testing this has proven to be accurate enough in 99% of the cases.
If users run into precission errors, this function could be extended to also check for `Top`, `Bottom`, `Left`, `Right`, but I'm worried about performance in huge graphs for that. 

The code tries to find the most ideal box to connect to, for conditions see the Linq expressions.

### Rip node from connection and slot into existing connection

https://github.com/user-attachments/assets/d619af75-dbe1-4bba-9c48-622a74ed8786

When starting to move node(s) while holding down the ALT key, it will rip it from any existing connections. It will also patch the connection as if the node had never existed.

You can also slot a node into an existing connection by dragging it over a connection. If there is enough space for the node, it will just slot in. If not, the nodes to the right of the moved node will move to the right to make space.
This feature is currently only supported when moving one node.

### DIFF
--- a/Source/Editor/Surface/Elements/OutputBox.cs
+++ b/Source/Editor/Surface/Elements/OutputBox.cs
@@ -34,11 +34,6 @@ namespace FlaxEditor.Surface.Elements
         /// </summary>
         public const float DefaultConnectionOffset = 24f;
 
-        /// <summary>
-        /// Distance for the mouse to be considered above the connection
-        /// </summary>
-        public float MouseOverConnectionDistance => 100f / Surface.ViewScale;
-
         /// <inheritdoc />
         public OutputBox(SurfaceNode parentNode, NodeElementArchetype archetype)
         : base(parentNode, archetype, archetype.Position + new Float2(parentNode.Archetype.Size.X, 0))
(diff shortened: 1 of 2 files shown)


snoweuph:
this looks cool :D

xxSeys1:
@mafiesto4 

Hi, 

will this PR get a milestone? If not, is there anything I can change so it gets one?

I get that it may look like bloat a bit, but it looks like people want this, see [here](https://discord.com/channels/437989205315158016/509056735844106251/1468637521746595892), [here](https://discord.com/channels/437989205315158016/509056735844106251/1468381465707282443), [here](https://discord.com/channels/437989205315158016/509056735844106251/1453367966543642820), #50:

image

, #2998 or the reactions to the PR.

PR #3921, opened 2026-02-05
Add NLerp methods on Vectors

ThePhantomMask:
And I added the Normalised Lerps on Vectors.
Resource:
https://keithmaggio.wordpress.com/2011/02/15/math-magician-lerp-slerp-and-nlerp/

### DIFF
--- a/Source/Engine/Core/Math/Vector2.cs
+++ b/Source/Engine/Core/Math/Vector2.cs
@@ -982,6 +982,33 @@ namespace FlaxEngine
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


mafiesto4:
I don't think it's partially useful to be inside the core engine as it's just lerp+normalize operations combined.

ThePhantomMask:
> I don't think it's partially useful to be inside the core engine as it's just lerp+normalize operations combined.

Hm...I can also close this pull because you're actually right, because I thought it was very useful.

PR #3926, opened 2026-02-07
Fix exception on visject cm when removing search text.

Tryibion:


xxSeys1:
Hi,
I think I already have a fix for this in #3786 (https://github.com/FlaxEngine/FlaxEngine/pull/3786/changes/e78191cc827bb47a878f7d2bdd258bdbde4c7e79)

Tryibion:
If you change it in your PR to be `==` instead of `>` then it should be correct and I will close this one. That code is supposed to run when there is no text and nothing is selected.

xxSeys1:
Done in b19611e3d2626cda2de1e05836d8062b8a28cd1e

PR #3944, opened 2026-02-17
Initial support for MSDF fonts

fibref:
Hi! This PR adds basic MSDF rendering support for the engine.

#### Changes:
1. Integrates ```msdfgen```.
2. Adds a field for FontAsset::_options, and now it looks like this:
    ```cpp
    struct FontOptions
    {
        FontHinting Hinting;
        FontFlags Flags;
        FontRasterMode RasterMode; // enum { Bitmap, MSDF }
    };
    ```
    That also means bumping the ```FontAsset```'s ```serializedVersion``` to 4.
2. Adds MSDF font atlas generating logic in ```FontManager```.
3. Creates a font material to render MSDF fonts (attached to this PR).

#### Build
To build this branch, you would have to manually build msdfgen with this command:
```
.\Binaries\Tools\Flax.Build.exe -log -ReBuildDeps -verbose -depsToBuild=msdfgen
```
You can do this after running GenerateProjectFiles.

#### Usage
In ```TextRender```:
1. Create a MSDF version of the font you want to use:
    
2. Select it as font and MSDFFontMaterial as material.

In UI:
MSDF support is built-in, so you just have to select a font with MSDF enabled.

It's recommended to use a fixed font size for all the texts using a specific font and use ```scale``` to adjust their size (to save space and computation power).

#### Advantages
1. Sharp edges
    
    font size is 32 for both
2. Arbitrary scaling
    
    size: 32 for Bitmap, 16 for MSDF

#### Disadvantages
1. More expensive than Bitmap.
2. It have to be rendered as transparent objects, so SSR won't work.
3. May not be as good as Bitmap at small sizes.

Material:
[MSDFFontMaterial.zip](https://github.com/user-attachments/files/25353362/MSDFFontMaterial.zip)

Some areas I'm unsure about:
1. Does adding ```RasterMode``` to ```FontAsset``` make sense, or would it be better handled elsewhere?
2. Is there a better way to arrange the MSDF atlas generation logic?
3. I currently have no idea how I can use this with editor UI fonts. Many of them call ```DrawText``` to render, but it don't have a ```scale``` to control. The font sizes are quite small as well.

If you see any design issues or better approaches, please don't hold back. I'm happy to refactor at this stage. Thanks!

### DIFF
--- a/Source/Engine/Render2D/MSDFGenerator.h
+++ b/Source/Engine/Render2D/MSDFGenerator.h
@@ -114,10 +114,10 @@ public:
         int32 width = static_cast<int32>(Math::CeilToInt(bounds.r - bounds.l + pxRange));
         int32 height = static_cast<int32>(Math::CeilToInt(bounds.t - bounds.b + pxRange));
         msdfgen::Bitmap<float, 4> msdf(width, height);
+
+        auto transform = msdfgen::Vector2(Math::Ceil(-bounds.l + pxRange / 2.0), Math::Ceil(-bounds.b + pxRange / 2.0));
         
-        msdfgen::SDFTransformation t(
-            msdfgen::Projection(1.0, msdfgen::Vector2(-bounds.l + pxRange / 2.0, -bounds.b + pxRange / 2.0)), msdfgen::Range(pxRange)
-        );
+        msdfgen::SDFTransformation t(msdfgen::Projection(1.0, transform), msdfgen::Range(pxRange));
         correctWinding(shape, bounds);
         generateMTSDF(msdf, shape, t);
 
@@ -136,7 +136,7 @@ public:
         }
         outputWidth = width;
         outputHeight = height;
-        top = static_cast<int16>(Math::CeilToInt(bounds.t + pxRange / 2.0));
-        left = static_cast<int16>(Math::FloorToInt(bounds.l - pxRange / 2.0));
+        top = height - static_cast<int16>(transform.y);
+        left = -static_cast<int16>(transform.x);
     }
 };


ZaPx64:
I would love to test this, but I can't build your branch. Does this require any specialized setup?

fibref:
> I would love to test this, but I can't build your branch. Does this require any specialized setup?

You would have to build msdfgen manually, since it's not in the LFS registry. Execute this before building the engine:
```
.\Binaries\Tools\Flax.Build.exe -log -ReBuildDeps -verbose -depsToBuild=msdfgen
```

If it's something else, please provide more details such as console outputs.
I will add this to the PR description.

ZaPx64:
Thanks for providing the build info - I got it to build.
I tried this on a label control, ~~but unfortunately it did not render anything~~. TextRender on the other hand worked fine with one font but produced a somewhat inverted result with another font:

image

Edit: Ah right, the provided material is in the surface domain - changing it to UI actually works nicely, the inverted font problems remains though. Some characters render fine for some reason and I think it looks great on small font sizes:

image

fibref:
@ZaPx64 
OK, I think the flipped color is caused by different winding directions of contours. I'll try to add a gussing logic and see if it can fix it. Would you mind sharing your font so I could check my fix?

By the way, there seems to be other artifacts (the 'e' in your first image) caused by shape self-intersection (I guess). I don't think this can be solved. You may have to fix the font manually using something like FontForge before using it in the game. There may be some tools that can do this automatically? I don't know.

ZaPx64:
Thanks for looking into it. The font is [available here](https://www.dafont.com/lt-cushion.charmap?f=3).

fibref:
I believe I fixed it.

fibref:
Update: Now MSDF fonts in UI no longer need specifying a material.

ZaPx64:
> Update: Now MSDF fonts in UI no longer need specifying a material.

This is big! No need for a migration utility then.

I have tested it again, and I think the results are really good, well done!
The 'e' seems to work in almost all places, but now the 't' has some vertical offset. Generally, the difference between bitmap and msdf is quite noticable by direct comparison - msdf is not blurry at all:

msdf

fibref:
@ZaPx64 
I think it's not a bug. When font atlases are generated, translation is calculated in pixels, so it's normal to have some deviation.
Try to increase your font size and scale it back down to let calculations done more precisely.
small
size:16 scale:5

big
size:80 scale:1

ZaPx64:
I see, that makes sense. And it's not possible to do the offset calculation with sub-pixel precision because the font atlas uses int16 for offset, right?

From the user experience perspective that introduces a problem: when working with UI you would usually work with font sizes around 10-20 px, and scaling every label is a bit cumbersome and introduces a whole set of other problems (layouting, animations, ...)

If I understand correctly, MSDF does not generate any raster graphics, so maybe the concept of font size could be scrapped altogether when working with MSDF. Instead, a virtual font size makes more sense I think, which could serve as a simple scaling factor, i.e. the font atlas would be generated at a nice resolution of 80 px, and when setting the font size on a label the scaling automagically happens. Just thinking out loudely here, I have no real grasp of the tech.

Alternatively, maybe font atlas offsets could be refactored to float. Not sure about the implications though.

fibref:
I made a fix to reduce vertical jitter of MSDF fonts.

> the font atlas would be generated at a nice resolution of 80 px...

~~This feature is on the way too. It will enhance the rendering quality and may save some memory.~~
Ah sorry, but I underestimated the complexity of it. It involves some larger changes. I think it's better to handle them in a separate PR so we can get this one merged first.

mafiesto4:
Thanks! I wrote a short tutorial for docs how to create outline material with MSDF font: https://github.com/FlaxEngine/FlaxDocs/blob/6f067d836b9a3b1049acda103d20eb8adab05cf6/manual/ui/tutorials/create-outline-font-material.md

PR #3966, opened 2026-02-28
Add tree view mode for content window.

Tryibion:
Adds a tree view only mode for the content window. This makes it nice for folks who want a smaller view of the items. This also saves the state of content tree folders between sessions.

I have ran into a bug with the arrow rect when resizing where it gets pushed all of the way to the left on folders several layers down, but it has only happened once to me and I can not get good repo steps to fix it, so I am not sure what is causing that.


mafiesto4:
Can you fix some merge conflicts so I can bring this into `1.12` update?

PR #3974, opened 2026-03-02
Removed incorrect binding generation

lifeformed:
I noticed that `GPUTexture.DownloadData()` stopped working. I tracked it down to a change in the Cpp wrapper generator. It has the condition `if (parameterInfo.IsOut || parameterInfo.IsRef)` and creates `&Temp` or `Temp` depending on the result. But `Temp` doesn't make sense, since the parameter is already always a reference type in the main conditional it's under (`parameterInfo.Type.IsRef`). It would just be an unused copy.

I'm not 100% I'm interpreting this all correctly, since I'm not that familiar with the binding system. But I removed the else condition and DownloadData seems to be working correctly now (no longer always null). Is this the right fix for this bug?


lifeformed:
Oh right I See strings don't do well here. The other fix seems right (https://github.com/FlaxEngine/FlaxEngine/pull/3977)

mafiesto4:
Thanks! I merged #3977 instead but will get you into credits of 1.12

PR #3987, opened 2026-03-04
Improve color picker

### LINKED ISSUE #2446: [Proposal] Find fix to color picker UX issue that GFS called a bug in his Flax 1.8 video
**Issue description:**

**Steps to reproduce:**

In [this video](https://www.youtube.com/watch?v=-7SlaNUIDcQ&t=2s), [Mike from GameFromScratch](https://www.youtube.com/@gamefromscratch) has an UX issue with the color picker where he does not click on _Ok_ to apply the color. Instead he closes the color picker, notices that the color is not applied and the proceeds to call that a bug (which it is not).

While this is clearly user error I think it's a bad UX decision that should be fixed.
I don't know what the best way to fix this is, but here are some ideas in no particular order:

**1. When closing the color picker apply the color**
When you close out of the color picker, apply the selected color. By closing I mean clicking the _X_ in the title bar or clicking anywhere else than the color picker.
The only way to cancel the color selection with this fix implemented would be to use the cancel button. It might also be ok to remove the _Ok_ button, as it's no longer needed. This _could_ confuse some users though.
Also: To really fix the issue at the root we would also need to remove the functionality where you can click anywhere but the color picker to close because some users may expect this to apply the color.

**2. Remove native window title bar and only have _Ok_ and _Cancel_ buttons**
This would eliminate the issue at the root. The user is now forced to click either _Ok_ or _Cancel_. This imo is only fixing the symptoms though, because we introduce a new UX flaw where the user has to click a button every time after selecting a color.
Also the removal of the native window title bar sounds like it could lead to some problems and I don't know if flax can do non native windows right now (would be stupid to implement this for a bad fix).

**3. Make the _Ok_ and _Cancel_ buttons bigger and more eye catching**
Make both buttons bigger and maybe center them at the bottom of the color picker. They are kind of going under in the visual clutter that is the color picker at the moment.

Just for reference this is how Godot does it which I think is a very good and UX friendly implementation:
image
The color picker has no title bar and no buttons. Closing the picker applies the color.
This means you can't cancel out of picking a color once you made a change. Godot combats this by having recent colors in a drop down.
image

none

**Flax version:**

1.8

xxSeys1:
image

Better version of #2691 with (almost) the same changes (but not the same code).

If #2691 was a messy open heart surgery, this is one using one of those fancy Davinci surgery robots (translated: Code is a lot more cleaner).

Closes #2446.

## Main Changes
- Hide mouse cursor when clicking on wheel or sliders (this time in a sane way lol)
- Move mouse position to knob position when letting go of slider or wheel
- Show mouse again when letting go of slider or wheel (obviously)
- Provide some visual feedback when the clicks on the wheel or sliders
- Make sliders wider
- Add alpha grid background to alpha slider
- Color will accept on close dialog or focus loss
- Color will cancel on ESC press
- Increased saved colors count from 8 to 10
- Moved RGB and HSV into tabbed interface
- Moved eyedropper icon
- Removed "Auto Accept Color Picker Change" option since there are no "Cancel" or "OK" button anymore
- Made sure to reset cursor type when the color picker is closed
- Hue value box now loops
- HV wheel now uses a material instead of a sprite
- Improved layout of color picker

### Assets
You can find all assets needed for this (hv wheel material and updated sprite sheet) in this folder:

[UpdatedColorPickerAssets.zip](https://github.com/user-attachments/files/25746439/UpdatedColorPickerAssets.zip)

Make sure to place `HSWheel.flax` at `[FlaxEngineRepoFolder]/Content/Editor` and to re- import the editor sprites atlas.

### DIFF
--- a/Source/Editor/GUI/Dialogs/ColorSelector.cs
+++ b/Source/Editor/GUI/Dialogs/ColorSelector.cs
@@ -194,10 +194,11 @@ namespace FlaxEditor.GUI.Dialogs
             // Wheel
             float boxExpand = (2.0f * 4.0f / 128.0f) * _wheelRect.Width;
             Render2D.DrawMaterial(_hsWheelMaterial, _wheelRect, enabled ? Color.White : Color.Gray);
-            Color hsColor = Color.FromHSV(new Float3(Color.ToHSV().X, 1, 1));
+            Float3 hsv = _color.ToHSV();
+            Color hsColor = Color.FromHSV(new Float3(hsv.X, hsv.Y, 1));
             Rectangle wheelRect = wheelDragRect;
             Render2D.FillRectangle(wheelRect, hsColor);
-            Render2D.DrawRectangle(wheelRect, _isMouseDownWheel ? Color.Gray : Color.Black, _isMouseDownWheel ? 3.0f : 1.0f);
+            Render2D.DrawRectangle(wheelRect, Color.Black, _isMouseDownWheel ? 2.0f : 1.0f);
         }
 
         /// <inheritdoc />


mafiesto4:
Thanks! I haven't updated Icons Atlas as it was modified on `1.12` branch so it would conflict and that removed icon can remain there for now.

PR #3989, opened 2026-03-05
Support for nested physics actors

lifeformed:
This allows for nested rigid bodies and character controllers. It keeps track of IPhysicsActors' nested depth during parenting changes. This requires some calculations from all Actors' parenting changes. When flushing transforms, it performs them grouped by depth, doing the lowest depth ones first, asynchronously if reasonable.

I'm not sure if 25 is a reasonable async limit? It was 50 before but since it's being divided I thought I should lower it, but I'm not sure. I also made this an option since not everyone might need it and there is some overhead. I'm not sure if that's the right place to define the flag.

### DIFF
--- a/Source/Engine/Core/Config.h
+++ b/Source/Engine/Core/Config.h
@@ -45,6 +45,9 @@
 // Enable/disable assertion for Engine low layers
 #define ENABLE_ASSERTION_LOW_LAYERS ENABLE_ASSERTION && (BUILD_DEBUG || FLAX_TESTS)
 
+// Enable nested physics bodies
+#define NESTED_PHYSICS_BODIES 1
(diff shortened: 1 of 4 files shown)


lifeformed:
Actually this seems to have problems when the asynchronous condition happens. Some reason updating the buckets asynchronously causes physics weirdness (some things not colliding at all). Even though the actors in the buckets are all independent from each other. I think it's something to do with the OnTransformChanged propagation. They trigger PhysX SetPose. I thought maybe doing that asynchronously causes some issue with broad phase structures or something. But I tried not doing any SetPoses during that step, and it didn't seem to help. Just updating them sequentially seems to work fine.

PR #3999, opened 2026-03-11
Add unified FlaxEngine.Random random utility for improved performance and thread safety

will11600:
# Summary

This pull request introduces `FlaxEngine.Random`, a new high-performance, thread-safe random number generation utility, and seamlessly refactors the existing `RandomUtil` and `RandomStream` classes to resolve underlying thread-safety and performance issues without breaking backward compatibility.

The new `Random` class  implements a Fast Linear Congruential Generator (LCG) based on [_Numerical Recipes in C_](https://www.cec.uchile.cl/cinetica/pcordero/MC_libros/NumericalRecipesinC.pdf#page=308) to produce pseudo-random integers. It entirely avoids floating-point division by using bitwise operations to discard the lowest-entropy bits and directly construct the IEEE 754 float mantissa.

The prior implementations relied on `System.Random` instances, which are not strictly thread-safe and can corrupt under concurrent access. `FlaxEngine.Random` solves this lock-free by utilizing `[ThreadStatic]` local states (`_threadLocal`).

# Key Changes & Refactors

## `RandomUtil`

* Replaced the static `System.Random.Next()` allocation inside `RandomUtil.Rand()` with a direct call to the new thread-safe `FlaxEngine.Random.Float()`.

## `RandomStream`

* Replaced the parameterless initialization (which previously allocated a new `System.Random` on the heap and used the system clock, risking duplicate seeds in tight loops) with `FlaxEngine.Random.Integer()`.
* Preserved identical stream-stepping behavior by wrapping the `Seed` struct instead of `Rng.State`.

# Newly Added Features

## Basic Value Generation

The class provides fundamental methods for generating random numbers, vectors, and colors, automatically handling the local state for the current thread.

```csharp
// Generate a pseudo-random 32-bit signed integer.
int randomInt = Random.Integer();

// Generate a float between 0.0 (inclusive) and 1.0 (exclusive).
float randomFloat = Random.Float();

// Generate a random float following a Gaussian (normal) distribution.
// By default, mean is 0.0f and standard deviation is 1.0f.
float gaussianFloat = Random.Gaussian(mean: 5.0f, standardDeviation: 2.0f);

// Generate a random color with random hue, saturation, and value.
Color randomColor = Random.ColorHSV();

// Generate a color with a random hue and value, but a specific saturation.
Color vibrantColor = Random.ColorHSV(saturation: 1.0f);

// Generate random, normalized vectors
Float3 randomDir3D = Random.Vector3();
Float2 randomDir2D = Random.Vector2();

```

## Probabilistic Conditions (`Chance` Enum)

Instead of manually checking random floats against thresholds, the `Random.Condition()` method evaluates a probability based on the readable `Chance` enum.

```csharp
// Evaluates to true 75% of the time.
if (Random.Condition(Chance.Likely))
{
    Debug.Log("This was likely to happen!");
}

// Defaults to Chance.Even (50% probability) if no argument is passed.
bool flipCoin = Random.Condition(); 

```

The `Chance` enum defines the following probabilities:

* **Impossible**: 0% probability.
* **Remote**: 1% probability.
* **Unlikely**: 25% probability.
* **Even**: 50% probability.
* **Likely**: 75% probability.
* **Expected**: 99% probability.
* **Certain**: 100% probability.

## Collection Utilities

`Random` includes utilities to safely and easily select or shuffle items in `Span<T>`, `IList<T>`, and `IReadOnlyList<T>` collections.

```csharp
List<string> lootTable = new List<string> { "Sword", "Shield", "Potion", "Gold" };

// 1. Choose: Selects a random element. Throws an ArgumentException if the collection is empty.
string drop = Random.Choose(lootTable);

// 2. TryChoose: Safely attempts to select a random element.
if (Random.TryChoose(lootTable, out string safeDrop))
{
    Debug.Log($"You found: {safeDrop}");
}

// 3. Shuffle: Randomly rearranges the elements in place.
lootTable.Shuffle(); 

```

## Fluctuation

The `Fluctuate` method allows you to apply a random variance to a base value, which is particularly useful for adding natural jitter to game mechanics (like damage numbers or spawn timings). It requires a type that implements `INumberBase<T>`.

```csharp
float baseDamage = 100f;
float maxVariance = 15f;

// Will return a value anywhere between 85f and 115f.
float actualDamage = Random.Fluctuate(baseDamage, maxVariance);

```

## Range Generation with Distribution Biases

While standard random generators provide uniform distributions, `Random` offers methods that bias the outcome towards specific parts of a range. These methods work with any numeric type implementing `IFloatingPoint<T>`.

```csharp
float min = 1f;
float max = 100f; // Exclusive upper bound

// Standard uniform range (equal chance for all values).
float uniform = Random.UniformRange(min, max);

// Biased towards the median (values near 50 are much more likely).
float medianBiased = Random.MedianBiasedRange(min, max);

// Biased towards the extremes (values near 1 or 99 are more likely).
float extremesBiased = Random.ExtremesBiasedRange(min, max);

// Biased towards the maximum (values near 99 are more likely).
float maxBiased = Random.MaxBiasedRange(min, max);

// Biased towards the minimum (values near 1 are more likely).
float minBiased = Random.MinBiasedRange(min, max);

```

## Seed State & Deterministic Sequences (`Seed`)

The `Seed` struct represents a snapshot of a random state at a given point in time. It is strictly sequential and guarantees absolute determinism, making it highly useful for reproducible outcomes, procedural generation seeds, or lock-free concurrency.

`Seed` implements `ISpanFormattable`, `IUtf8SpanFormattable`, `ISpanParsable<Seed>`, and `IUtf8SpanParsable<Seed>`, and supports explicit casts to/from `int` and `float`. A `Seed` can also be advanced directly via the `++` operator, and uninitialized seeds are lazily seeded from `Seed.SharedState` on first use.

Because `Seed` wraps its own underlying integer, it exposes all the same core instance methods as the static `Random` class (such as `Condition`, `Fluctuate`, `UniformRange`, `Choose`, etc.), but evaluates them against its specific frozen state.

### Reproducibility and the `++` Operator

Rather than advancing a hidden global state, you can traverse a deterministic chain of states using the `++` operator. This applies the same Fast LCG algorithm to guarantee that every state points to a mathematically unique subsequent state in the sequence.

```csharp
// Capture the random state at this exact moment.
Seed state = Random.CurrentSeed;

// You can use the snapshot to perform RNG operations tied to this frozen state.
// Calling these multiple times on the SAME state yields the exact same result.
int uniformVal = Random.UniformRange(0, 10, state);
bool conditionA = Random.Condition(state, Chance.Unlikely);

// To get a new random result predictably, advance to the next state in the chain.
Seed nextState = ++state;

// These will be pseudo-random relative to the initial state, but 100% reproducible.
int nextUniformVal = Random.UniformRange(0, 10, nextState);
```

Explicit casts make it easy to work with the underlying value directly:

```csharp
Seed seed = new Seed(42);

int  raw   = (int)seed;   // Extract the raw integer state.
float norm = (float)seed; // Normalize to [0.0, 1.0) using IEEE 754 bit manipulation.
Seed copy  = (Seed)42;    // Construct a seed from a literal.
```

### Thread-Safe Atomic Operations

Specialized static methods are provided for lock-free, thread-safe updates using `Interlocked` operations under the hood. If you are managing concurrent systems (like a multithreaded particle spawner or a shared procedural chunk generator), you can safely swap seeds across threads without `lock` statements.

```csharp
private Seed _sharedSpawnerState = new Seed(12345);

public void UpdateSpawnerState(Seed newState)
{
    // Atomically replaces the shared state and returns the original one.
    Seed previousState = Seed.Exchange(ref _sharedSpawnerState, newState);
    Debug.Log($"Replaced state {(int)previousState} with {(int)newState}");
}

public bool TryCompareAndSwap(Seed expected, Seed desired)
{
    // Atomically compares the current state to 'expected', and if they match, swaps to 'desired'.
    Seed original = Seed.CompareExchange(ref _sharedSpawnerState, desired, expected);
    return original == expected;
}
```

### Equality and Comparisons

`Seed` fully implements `IEquatable<Seed>` and the `==` / `!=` operators. Because a seed is entirely defined by its internal integer, comparing two seeds is effectively free.

```csharp
Seed stateA = new Seed(42);
Seed stateB = new Seed(42);

if (stateA == stateB)
{
    // Resolves to true. Both states will yield the exact same random outputs.
    Debug.Log("States are identical!");
}
```


will11600:
Hi @mafiesto4,

Thanks for adding this to the 1.13 milestone! I'm really glad to see this moving forward.

I had a quick question regarding the older random utilities. Now that the new allocation-free `Seed` struct and thread-safe `Random` class are in place, the old wrappers in `RandomStream`.cs and `RandomUtil`.cs are effectively redundant.

Specifically, `RandomUtil` still exposes a static `System.Random` instance, which bypasses the thread-safety improvements if users access it directly. Additionally, `RandomStream` is a class, which means it still incurs heap allocations.

Would you like me to mark these older classes and methods with the `[Obsolete]` attribute in this PR?

I can add something like: `[Obsolete("RandomUtil is deprecated. Please use FlaxEngine.Random instead.")]` to guide users. However, I know this might cause warning spam for existing projects updating to 1.13, so I wanted to get your approval before pushing that change.

Let me know how you'd prefer to handle this!

mafiesto4:
Sorry but I have to push it to next update 1.14.

PR #4020, opened 2026-03-24
Add missing Length method

duarteroso:
There seems to have been a dodgy commit which resulted in a missing `Length()` method being used but not defined in `Vector4Base`.

This PR adds the definition of said method.
Problem appeared in [Vector4Base::NormalizeFast:536](https://github.com/FlaxEngine/FlaxEngine/blob/master/Source/Engine/Core/Math/Vector4.h#L536) when building on macOS

### DIFF
--- a/Source/Engine/Core/Math/Vector4.h
+++ b/Source/Engine/Core/Math/Vector4.h
@@ -152,6 +152,12 @@ public:
     {
         return Math::IsOne(X) && Math::IsOne(Y) && Math::IsOne(Z) && Math::IsOne(W);
     }
+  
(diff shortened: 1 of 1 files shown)


mafiesto4:
Ah I just fixed it at the same time in 7f1add4bddcc386d1b34ffe3c230f57e276c5814 (cherrypick fix from `1.12` branch).
Also, that inlined code in `Normalize` is on purpose due to performance (eg. in `Debug` builds where inlining doesn't happen that much).

duarteroso:
My bad! Should have been in that branch to begin with. Good job on the reaction time ;)

PR #4025, opened 2026-03-27
Fix gizmo to not use viewport fov and to only change cam orientation on MouseUp

xxSeys1:
See *Title*.

Fixes this from happening at lower/ higher editor camera fov than 60°:
image

I don't know why when I modify the `editorViewport` to `60f` it will also modify the main editor viewport fov????
Like it's not passed as a reference or anything??
Prolly a skill issue on my end.

Anyways I ended up fixing this by just caching the fov and then re- applying it.

Blender btw does the same thing. The gizmo fov is always a fixed number no matter the editor camera one.

Also makes the camera only change direction when the mouse button is let go off (default button behavior).

### DIFF
--- a/Source/Editor/Gizmo/DirectionGizmo.cs
+++ b/Source/Editor/Gizmo/DirectionGizmo.cs
@@ -55,9 +55,13 @@ internal class DirectionGizmo : ContainerControl
         {
             // Inline EditorViewport.ProjectPoint to save on calculation for large set of points
             _viewport = new FlaxEngine.Viewport(0, 0, editorViewport.Width, editorViewport.Height);
+            // Force the FOV to 60° (viewport default) to keep the gizmo size consistent regardless of the editor viewport FOV setting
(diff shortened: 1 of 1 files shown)


mafiesto4:
I don't think changing FOV to get correct matrix is a good idea. Can you try adjusting `heightNormalization` for FOV scale just like I did for `OrthographicScale`? It fixed the ortho scale spread that could be done here too.

xxSeys1:
> I don't think changing FOV to get correct matrix is a good idea. Can you try adjusting `heightNormalization` for FOV scale just like I did for `OrthographicScale`? It fixed the ortho scale spread that could be done here too.

I just tried a few things, but I don't really understand what that thing does lol.

If you already know what to put there, feel free to commit that and close this pr, otherwise I will try to find a proper fix tomorrow.

mafiesto4:
I did a hack in 2b538e4864290335da922a385e8c09b51766f013. Very nasty but we won't have any issues with FOV changes in future. Just i ncase.

PR #4059, opened 2026-04-18
MSDF Font enhancement and bugfix

fibref:
Add support for rasterizing all MSDF fonts under a fixed size (adjustable) to save memory and enhance quality. To achieve this, I added a field to ```Font``` and moved the character cache from ```Font``` to ```FontAsset``` (as multiple fonts may share characters).

PS: Sorry for this PR coming in so late😥. If you don’t have time to review everything, feel free to just take the last commit. It's a bugfix.


mafiesto4:
Thanks! I took that fix for 1.12 update in 28ea7f134d6780ce4cd8cade69898de39b61fef2. PR can be merged later on once it's stable.

mafiesto4:
Hey, I have one question: what about using that MSDF size to be set on font asset instead? So the font reference would be simpler. I assume that would be the optimal and simple case where imported font has pre-set size to be reused in the whole project for font sharing, right?

fibref:
I agree that setting MSDF size in font asset is a better design. My original thoughts were that a font asset is a size-independent objects (by definition), so MSDFsize should be stored in a font object, and that a user may want to vary the MSDF size across widgets. But it turned out that my design made things complicated, and text-related editor options (with two font size sliders) are quite confusing. Also, varying MSDF size is not that important, only 1 or 2 different asset variant are needed in most cases.

So I'll have a rework on this feature, moving MSDF size to font asset. When rendering MSDF fonts, the original size parameter could act as a scaler to match the rendered size with bitmap fonts.

I'll close this PR and open another one later. Thanks for the feedback!

PR #4063, opened 2026-04-19
Add context menu entry for creating material subinstances from instances

envision3d:
The editor already allows manually setting the base material of an instance to another instance. This PR just adds the context menu entry convenience and functionality to the `MaterialInstanceProxy`.

The existing code in `MaterialProxy` could probably be updated and overloaded to handle both, but I'm not sure if it's appropriate to do so.  The only significant difference is the type provided to the `LoadAsync()` call for setting the base material.  I'll leave that up to you.

### DIFF
--- a/Source/Editor/Content/Proxy/MaterialInstanceProxy.cs
+++ b/Source/Editor/Content/Proxy/MaterialInstanceProxy.cs
@@ -2,6 +2,7 @@
 
 using System;
 using FlaxEditor.Content.Thumbnails;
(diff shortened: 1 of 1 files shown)


mafiesto4:
Yeah it would make sense to introduce `MaterialBaseProxy` to share code between those two.

mafiesto4:
Implemented in 6fa38b75d593c8e8805427a7b55e5673c728a4cf.

PR #4064, opened 2026-04-20
Fix Vector Slerp methods

ThePhantomMask:
I realised that when I applied the Slerp methods, the value didn't exceed 1, so I fixed it.


ThePhantomMask:
Nothing; after testing and re-reading the website I used as a basis for the Slerp implementation, it's recommended to use it mainly for rotations, not positions.
So I'm closing this pull request.

PR #4085, opened 2026-05-10
fix: support for non-standard library/executable paths on linux

yuifm:
I ran into some issues on a distro which doesn't use the Filesystem Hierarchy Standard. However I believe these fixes will help for situations other than just NixOS.

Changes have been tested on Ubuntu 26.04 LTS and NixOS.

This PR does a few things:

## Replaces `#!/bin/bash` with `#!/usr/bin/env bash` 

This is considered best practice due to some systems storing the shell binary in different locations.
I made this change for the MacOS code too, as the same situation applies.

## Checks the PATH for zenity/kdialog.

On some systems, these programs aren't necessarily in `/usr/bin`, so it now looks for them in the PATH.

## Uses `pkg-config` for library discovery

I've created a PkgConfig helper class that shells out to pkg-config when looking for libraries such as `glib`, `gio`, and `gobject`. it still falls back to the old paths if pkg-config is unavailable.

### DIFF
--- a/Development/Scripts/Linux/CallBuildTool.sh
+++ b/Development/Scripts/Linux/CallBuildTool.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # Copyright (c) Wojciech Figat. All rights reserved.
 
 set -e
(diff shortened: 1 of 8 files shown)


yuifm:
To be honest, I left in the old hardcoded paths as fallbacks to avoid any regressions, but I do think it'd be good to get rid of it and rely on pkg-config. Let me know what you think.

mafiesto4:
I'm not very experienced with Linux in this case so would be good to get some review/feedback from anyone using Flax on Linux.

PR #4097, opened 2026-05-14
Fix the cursor not hiding when right click dragging on the main editor viewport if the game panel has focus and both panels are docked

### LINKED ISSUE #4040: [Bug]: When docked *Game* Panel was focused previously, moving camera in main editor panel does not hide the mouse cursor
### Description

https://github.com/user-attachments/assets/e312e749-7d46-4902-9dd7-1c2b26231e94

### Steps to reproduce

1. Dock the Game panel somewhere in the editor
2. Click in it or its tab title bar to focus it
3. Hover over the main editor window and right click and drag your mouse around

### Version

1.12

### Relevant logs

```shell

```

xxSeys1:
Fixes #4040 

Basically performs the same logic that was before only executed during play mode during editor mode

### DIFF
--- a/Source/Editor/Windows/GameWindow.cs
+++ b/Source/Editor/Windows/GameWindow.cs
@@ -662,8 +662,8 @@ namespace FlaxEditor.Windows
         {
             base.OnMouseLeave();
 
-            // Remove focus from game window when mouse moves out and the cursor is hidden during game
-            if (ContainsFocus && Parent != null && Editor.IsPlayMode && !Screen.CursorVisible && Screen.CursorLock == CursorLockMode.None)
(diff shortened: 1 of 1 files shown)


mafiesto4:
Two things:

1) I would rather use more explicit parentheses to improve the readability of `&&` and `||` sections:

```cs
// Remove focus from game window
if ((ContainsFocus && !Editor.IsPlayMode) || 
    (Editor.IsPlayMode && !Screen.CursorVisible && Screen.CursorLock == CursorLockMode.None))
{
    Parent?.Focus();
}
```

2) It fixes the issue but breaks undocking Game window from that side-by-side state.

PR #4102, opened 2026-05-18
Improve Dropdown Control

ThePhantomMask:
I also realised that in Color (C#), the boolean value for transparency was missing, as it is in C++, so I’ve added it.
In addition to the colour brightness, I’ve also added the transparency setting to TextColorHighlighted.


mafiesto4:
You don't need to merge with master so frequently, as I have to squish those commits to improve history readability :)

ThePhantomMask:
Ah ok, I won’t do it again.

mafiesto4:
Merged in d1148fdcfd5f299d98581e783e352cf1e433290e.

PR #4117, opened 2026-05-23
Fix MacOS Delete key handling in editor views.

ifromstone:
## Summary

Fix macOS Delete key handling in editor views where the physical Delete key is reported as `Backspace`.

On macOS, pressing Delete while an item is selected in the Content window could trigger folder history navigation instead of deleting the selected content item. This change adds a small shared input helper and uses it in the affected editor views so the key performs deletion when there is an active selection.

## Changes

- Added `InputUtils.IsKeyPerformDeletion()` to centralize platform-specific delete-key behavior.
- Updated the Content view to delete selected content items before falling back to Backspace navigation.
- Updated the Content tree to delete selected content items when the platform delete key is pressed.
- Updated scene editor key handling paths to use the same helper for selected scene objects.

## Notes

The macOS-specific behavior is kept inside `InputUtils`. On non-MacOS platforms, the helper does not change existing Delete shortcut handling.

### DIFF
--- a/Source/Editor/Content/GUI/ContentView.cs
+++ b/Source/Editor/Content/GUI/ContentView.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using FlaxEditor.GUI.Drag;
 using FlaxEditor.Options;
 using FlaxEditor.SceneGraph;
+using FlaxEditor.Utilities;
 using FlaxEditor.Windows;
 using FlaxEngine;
 using FlaxEngine.GUI;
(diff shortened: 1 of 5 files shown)


mafiesto4:
Would it make sense to remap `Delete` key to `Backspace` under the hood so all code handles it that way? (keybinds, game scripts, vidual input, etc.))

ifromstone:
> Would it make sense to remap `Delete` key to `Backspace` under the hood so all code handles it that way? (keybinds, game scripts, vidual input, etc.))

@mafiesto4 This does make sense.
I've prepared a separate PR - https://github.com/FlaxEngine/FlaxEngine/pull/4123; if everything goes well with it, then this one can be closed.

Issue #2746, opened 2024-07-03
[Feature Request] In the property window, being able to set an Interface reference, instead of just Scripts


lifeformed:
If I have a public reference to a Script-derived class in my Script, the Property window lets me drag an actor with that Script into that property value element.  However, If that reference is an interface instead of a Script, it only shows a read-only dropdown:
![image](https://github.com/FlaxEngine/FlaxEngine/assets/1567199/07a14343-96bb-4020-b8c0-faa68d9b4c88)

It would be nice if there was a way to use it like the way the Script one works. I tried having the interface implement ISerializable, but it doesn't change things.

lifeformed:
Or if there's an easy way to implement a ValueContainer (or whatever the correct class is, I'm not super familiar with the editor internals) for that interface, that could work too.  Or maybe there already is, but there isn't much guidance on how to do that.

xxSeys1:
@mafiesto4 Could this be done for 1.13?

It would help me to keep my code more clean, since the script will be referenced as an Interface, so it is a bit more abstract and versatile.

I'm also sure a lot of other people will find this very useful.

mafiesto4:
Editor is one part, other side of this, harder to impl, is serialization to properly put the object ID for inetrfaces. Will see how `Newtonsoft.Json` coudl do it, on C++ it's rather doable.

mafiesto4:
Implemented in #4120 (with soem changes). Use new `ScriptingObjectInterfaceReference<T>` to reference interfaces (both C++ and C#).

PR #4120, opened 2026-05-24
Interface-based object references for scripting objects

ifromstone:
## Summary

This PR adds interface-based object references for scripting objects:

- `ScriptingObjectInterfaceReference<T>`
- `SoftObjectInterfaceReference<T>`

They work similarly to `ScriptingObjectReference<T>` and `SoftObjectReference<T>`, but the generic type is an interface. The stored object is still represented as a scene object reference/object ID, while access to the value is validated through the requested interface type.

## Details

- Added native hard and soft interface reference wrappers with shared interface validation helpers.
- Added C# editor attributes:
  - `ScriptingObjectInterfaceReferenceAttribute`
  - `SoftObjectInterfaceReferenceAttribute`
- Added bindings generator support for:
  - single interface references
  - array-like containers (`Array`, `Span`, `DataContainer`)
  - dictionaries with interface reference values
- Added editor picker support for interface-typed object references, including collection and dictionary value editors.
- Added JSON serialization/deserialization support for C# interface object references and interface reference collections, so editor undo/snapshots and reloads can preserve selected objects by ID.
- Added `JsonSerializer.TryParseID` and refactored GUID hex parsing to share one checked implementation between `ParseID`, `ParseHex`, and safe interface-reference JSON reads.
- Updated `Object.Find` and `Object.TryFind` to support interface type matching consistently.
- Added managed/native conversion support for interface references and dictionary marshalling.

## Notes

For native code, the new wrappers are intended for API fields/properties such as:

```cpp
API_FIELD() ScriptingObjectInterfaceReference<IMyInterface> Target;
API_FIELD() SoftObjectInterfaceReference<IMyInterface> SoftTarget;
API_FIELD() Array<ScriptingObjectInterfaceReference<IMyInterface>> Targets;
```

For C# scripts, interface object references use attributes on interface-typed fields/properties:

```csharp
[ScriptingObjectInterfaceReference]
public IMyInterface Target;

[ScriptingObjectInterfaceReference]
public List<IMyInterface> Targets;
```

### DIFF
--- a/Source/Editor/CustomEditors/CustomEditorsUtil.cs
+++ b/Source/Editor/CustomEditors/CustomEditorsUtil.cs
@@ -58,11 +58,13 @@ namespace FlaxEditor.CustomEditors
             if (targetType.Type == typeof(object) && values.Count > 0 && values[0] != null && !values.HasDifferentTypes)
                 return CreateEditor(TypeUtils.GetObjectType(values[0]), canUseRefPicker);
 
-            // Special case if property is interface but the value is implemented as Scripting Object that should use reference picker
-            if (targetType.IsInterface && canUseRefPicker && values.Count > 0 && values[0] is FlaxEngine.Object)
-                return new DummyEditor();
-
             // Use editor for the property type
+            if (canUseRefPicker &&
+                targetType.IsInterface &&
+                values.GetAttributes().Any(x => x is ScriptingObjectInterfaceReferenceAttribute || x is SoftObjectInterfaceReferenceAttribute))
+            {
+                return new FlaxObjectRefEditor();
+            }
             return CreateEditor(targetType, canUseRefPicker);
         }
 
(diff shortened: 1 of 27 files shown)


xxSeys1:
Closes #2746

xxSeys1:
Is the `[ScriptingObjectInterfaceReference]` attribute really needed?

Usually Flax will just show public fields in the properties panel without the user need to explicitly mark them as visible.

ifromstone:
> Is the `[ScriptingObjectInterfaceReference]` attribute really needed?
> 
> Usually Flax will just show public fields in the properties panel without the user need to explicitly mark them as visible.

@xxSeys1, hello
Yes, the attribute is intentional. Public fields are shown automatically, but an interface type alone does not say whether it should be treated as a scene object reference or as a regular managed interface value.

The attribute is only used to select the object-reference behavior: editor picker, GUID-based JSON serialization, undo/reload handling, and native interface reference marshalling. Without an explicit marker, every public IInterface or List\<IInterface\> would risk being interpreted as a Flax object reference, which could break normal managed polymorphic/interface fields.

mafiesto4:
Thanks! I merged it with some changes:
* Added `ScriptingObjectInterfaceReference<T>` to C# and use it instead of attribute for references,
* Removed `SoftObjectInterfaceReference<T>` for now,
* Used `MarshalAs=ScriptingObject*` for marshaling that ne winterface via object pointer instead of using custom codegen in bindings,
* Added more code to unit tests,
* Updated docs.

PR #4123, opened 2026-05-28
Fix: macOS Delete key handling (Backspace key remaping)

ifromstone:
## Summary

Fix macOS Delete key handling so the main keyboard Delete key works with editor shortcuts and runtime input checks.

On macOS, the main keyboard Delete key is reported by the platform as `Backspace`. Several editor paths bind deletion to `KeyboardKeys.Delete`, so pressing the physical Delete key could trigger Backspace-specific behavior, such as Content window history navigation, instead of deleting the current selection.

## Changes

- Treat macOS `Backspace` as an alias for `KeyboardKeys.Delete` in editor input bindings.
- Treat macOS `Backspace` as an alias for `KeyboardKeys.Delete` in keyboard polling (`GetKey`, `GetKeyDown`, `GetKeyUp`).
- Forward runtime `KeyDown`/`KeyUp` delegates for `KeyboardKeys.Delete` when macOS reports the physical Delete key as `Backspace`.
- Preserve text editing behavior by keeping the original key event as `Backspace` for UI controls.
- Check the Delete binding before Content view Backspace navigation so deleting a selected content item wins over folder history navigation.


PR_DESCRIPTION.md
mafiesto4, line 1: This file is not needed.

ifromstone: Sorry, I accidentally pushed it while preparing. Deleted the file.

PR #4125, opened 2026-05-29
Fix: Vulkan - preserve valid swapchain surfaces during resize

ifromstone:
## Background
This PR was motivated by very frequent Vulkan device-lost crashes on macOS while interacting with editor windows, especially when opening secondary windows or resizing them. The crashes were observed around swapchain/window handling paths, which led to reviewing how the Vulkan backend recreates swapchains and native surfaces.

## Summary
This change makes Vulkan swapchain recreation more precise and avoids destroying the native `VkSurfaceKHR` during normal resize and `VK_ERROR_OUT_OF_DATE_KHR` handling.

Previously, `GPUSwapChainVulkan::CreateSwapChain()` used the full `ReleaseGPU()` path when a surface already existed. That released the backbuffers and swapchain, but also destroyed the Vulkan surface. For ordinary resize this is unnecessary: the platform surface represents the native window/layer, while the swapchain represents the current set of presentable images for that surface.

The new code separates those lifetimes:

- `ReleaseSwapChain(bool releaseSurface)` releases swapchain-owned resources and synchronization state.
- Normal resize recreates only the swapchain and backbuffers.
- `VK_ERROR_OUT_OF_DATE_KHR` recreates only the swapchain.
- `VK_ERROR_SURFACE_LOST_KHR` recreates both the surface and swapchain.
- Full GPU resource release still destroys the surface.

The swapchain also now records the native window handle used to create the surface. If `_window->GetNativePtr()` changes, the existing surface is treated as invalid and recreated. This is important for Android lifecycle events where `ANativeWindow*` may be replaced, and also makes iOS safer if the underlying `UIView` or `CAMetalLayer` changes.

This approach is especially important for MoltenVK on Apple platforms. Recreating the native surface during secondary window creation or resize can be a heavier and less stable path than recreating only the swapchain images. Keeping the surface alive during ordinary resize reduces native presentation churn while still handling real surface invalidation.

The acquire-image synchronization was also tightened:

- Any pending active command buffer is flushed before adding the acquired-image semaphore wait.
- The acquired-image semaphore is waited on at `VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT`, which matches the first use of the swapchain image as a color attachment.

### DIFF
--- a/Source/Engine/GraphicsDevice/Vulkan/GPUSwapChainVulkan.cpp
+++ b/Source/Engine/GraphicsDevice/Vulkan/GPUSwapChainVulkan.cpp
@@ -41,6 +41,7 @@ void BackBufferVulkan::Release()
 GPUSwapChainVulkan::GPUSwapChainVulkan(GPUDeviceVulkan* device, Window* window)
     : GPUResourceVulkan(device, StringView::Empty)
     , _surface(VK_NULL_HANDLE)
+    , _surfaceWindowHandle(nullptr)
     , _swapChain(VK_NULL_HANDLE)
     , _currentImageIndex(-1)
     , _semaphoreIndex(0)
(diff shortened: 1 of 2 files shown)


mafiesto4:
It does work well on Android.

PR #4135, opened 2026-06-03
No path in window title

xxSeys1:
[Someone suggested](https://discord.com/channels/437989205315158016/438033006016462849/1508958411000250368) removing the project path from the main window title for privacy reason in screenshots/ screen recordings and I agree.

It also looks more clean.

This pr also adds the project path into the main window icon tooltip. It also adds info about if the current engine build is large worlds enabled. I frequently run into the problem of not really knowing if the engine build I have the project opened in has large world enabled or not, so this is a nice and quick way to check that.

image

image

### DIFF
--- a/Source/Editor/Modules/UIModule.cs
+++ b/Source/Editor/Modules/UIModule.cs
@@ -829,9 +829,9 @@ namespace FlaxEditor.Modules
             projectPath = projectPath.Replace('/', '\\');
 #endif
 
-            string largeWorld = "Large Worlds Enabled: false";
+            string largeWorld = "Large Worlds Disabled";
 #if USE_LARGE_WORLDS
-            largeWorld = "Large Worlds Enabled: true";
+            largeWorld = "Large Worlds Enabled";
 #endif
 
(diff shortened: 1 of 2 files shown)


Source/Editor/Modules/UIModule.cs
```diff
@@ -824,10 +824,20 @@ private void InitWindowDecorations(RootControl mainWindow)
                 driver = $" ({driver})";
 #endif
 
+            var projectPath = Globals.ProjectFolder;
+#if PLATFORM_WINDOWS
+            projectPath = projectPath.Replace('/', '\\');
+#endif
+
+            string largeWorld = "Large Worlds Disabled";
```
mafiesto4: I think we could skip this information if Large Worlds are unused so it's only when active. Just liek we do for SDL driver name that is only shown on Linxu where it matters.

xxSeys1: Done in https://github.com/FlaxEngine/FlaxEngine/pull/4135/commits/d103ccdbc3296543a1521294e6e356e324906af7

PR #4141, opened 2026-06-06
Fix Vulkan swapchain synchronization for acquired backbuffers

ifromstone:
## Summary

This fixes Vulkan swapchain backbuffer synchronization by waiting for the command buffer associated with the image returned by `vkAcquireNextImageKHR`, instead of waiting for the previously presented image before acquire.

The change also tracks the submitted command buffer fence counter per backbuffer, so stale/reused command buffer objects do not cause unnecessary waits. On iOS, the display link frame-rate hint is updated to use `preferredFrameRateRange` when available, with `CADisableMinimumFrameDurationOnPhone` enabled for ProMotion-capable devices.

## Why

The previous implementation waited in `GPUSwapChainVulkan::Begin()` using `_currentImageIndex`. With multi-buffered swapchains, that index can refer to the last presented image, not necessarily the image that will be acquired for the next frame. This can serialize rendering across swapchain images and defeat multi-buffering.

On iOS/MoltenVK this caused a stable FPS drop and camera stutter after a few seconds, even in a minimal scene. Moving the wait after acquire ties synchronization to the actual backbuffer that will be rendered into.

### DIFF
--- a/Source/Engine/GraphicsDevice/Vulkan/GPUSwapChainVulkan.cpp
+++ b/Source/Engine/GraphicsDevice/Vulkan/GPUSwapChainVulkan.cpp
@@ -25,16 +25,23 @@ void BackBufferVulkan::Setup(GPUSwapChainVulkan* window, VkImage backbuffer, Pix
     ImageAcquiredSemaphore = New<SemaphoreVulkan>(Device);
 }
 
+void BackBufferVulkan::WaitForSubmit()
+{
+    if (SubmitCmdBuffer)
+    {
+        if (SubmitCmdBufferFenceCounter == SubmitCmdBuffer->GetFenceSignaledCounter())
+            SubmitCmdBuffer->Wait();
+        SubmitCmdBuffer = nullptr;
+        SubmitCmdBufferFenceCounter = 0;
+    }
+}
+
 void BackBufferVulkan::Release()
 {
+    WaitForSubmit();
     Handle.Release();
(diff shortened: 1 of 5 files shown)


mafiesto4:
I'm getting some warnings from Vulkan layer when using this:

```
[Vulkan] Validation Error:1461184347(VUID-vkAcquireNextImageKHR-semaphore-01779) vkAcquireNextImageKHR(): Semaphore must not have any pending operations.
The Vulkan spec states: If semaphore is not VK_NULL_HANDLE, it must not have any uncompleted signal or wait operations pending (https://docs.vulkan.org/spec/latest/chapters/VK_KHR_surface/wsi.html#VUID-vkAcquireNextImageKHR-semaphore.
```

But those are gone after few frames. Now sure why.

mafiesto4:
The issue is gone if I merge #4125 first which does wait on the submit properly.

ifromstone:
> The issue is gone if I merge #4125 first which does wait on the submit properly.

Yes, that's correct.
I made an edit on my local branch, which already contained a #4125 edit. And then I split those changes off to master for PR.
So it's better to merge them all at the same time.

ifromstone:
> The issue is gone if I merge #4125 first which does wait on the submit properly.

Yes, that's correct.
I made an edit on my local branch, which already contained a #4125 edit. And then I split those changes off to master for PR.
So it's better to merge them all at the same time.

PR #4144, opened 2026-06-07
Add Nvidia PerfSDK support

ThePhantomMask:
Another Niko PR that integrates NVIDIA PerfSDK into Flax Engine to provide access to GPU performance counters and profiling data on supported NVIDIA hardware.

The implementation introduces the required PerfSDK initialization and data collection workflow, enabling the engine to gather detailed GPU performance metrics during runtime and profiling sessions. The integration is designed to work alongside the existing profiling infrastructure while ensuring graceful fallback behavior when PerfSDK is unavailable or unsupported.

This addition improves performance analysis capabilities, helping developers identify GPU bottlenecks, optimize rendering workloads, and gain deeper insight into graphics performance directly within the engine.

### DIFF
--- a/Source/Editor/Editor.Build.cs
+++ b/Source/Editor/Editor.Build.cs
@@ -44,6 +44,8 @@ public class Editor : EditorModule
         options.ScriptingAPI.SystemReferences.Add("System.Diagnostics.Process");
         if (Profiler.Use(options))
             options.ScriptingAPI.Defines.Add("USE_PROFILER");
+        if (RenderPerf.Use(options))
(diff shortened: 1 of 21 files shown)


Tryibion:
Would this be better as a plugin project?

ThePhantomMask:
> Would this be better as a plugin project?

I was thinking the same thing—that it might work better as a plugin editor, because it’s impossible on a Mac.
I created this PR based on Niko’s suggestion.

ThePhantomMask:
In the end, Niko will create a separate plugin, so I'm closing this PR.

PR #4155, opened 2026-06-14
Fix NodesCache desync after hotreload

Alewinn-TerizKoming:
Issue:
- Visject types popup context menu becomes desynchronized after multiple hotreloads (2)
- async cache not invalidated correctly when types are cleared

Fix:
- Clear cache 

Impact:
- No global system changes
- Only affects Visject NodesCache lifecycle


mafiesto4:
Good catch! Actual fix location should be `ClearTypes` method in `CodeEditingModule` that is missing a call to that `AllWithStd` cache clear.

PR #4172, opened 2026-07-03
1.13 linux perf sdk

niko1point0:
Adds necessary code to load PerfSDK .so files on linux, just like how Windows loads .dll files. In Flax 1.14 we should find a way to migrate this system to the plugin instead of the engine, but for now it'll do


niko1point0:
Wrong branch, whoops

PR #4178, opened 2026-07-07
Fix floating editor window ordering and focus restoration

ifromstone:
## Summary

This change fixes platform window ordering and focus behavior for parented floating editor windows.

On macOS, minimizable `Regular` windows with a logical parent are now kept as independent top-level Cocoa windows instead of native child windows. This prevents AppKit child-window side effects such as parent/sibling minimization, incorrect z-ordering, and floating windows being pushed behind their parent. Popup, tooltip, utility, and non-minimizable regular windows continue to use native parent attachment where that behavior is expected.

The macOS backend also restores focus after hiding or closing a parented regular window by selecting another visible Flax regular sibling window with the same parent, falling back to the parent only when no sibling is available. This keeps input working after closing a floating window without blindly raising the parent above the remaining floating windows.

On Windows, regular windows are explicitly moved to the top of the z-order when focused or mouse-activated, which makes clicking between multiple floating windows bring the selected window forward consistently.

## Changes

- Keep minimizable macOS `Regular` windows as top-level `NSWindow` instances while preserving their logical `Parent` in `CreateWindowSettings`.
- Continue using Cocoa native parent/child behavior for transient or non-minimizable windows where parent ordering is expected.
- Restore macOS focus after hide/close to another visible parented regular Flax window before falling back to the parent window.
- Avoid selecting arbitrary AppKit windows during focus restoration by mapping `NSWindow` instances back to Flax `Window` objects.
- Update Windows `BringToFront()` so regular windows also receive an explicit `SetWindowPos()` z-order update.
- Call `BringToFront()` from Windows `Focus()`.
- Handle `WM_MOUSEACTIVATE` for Windows regular windows to bring clicked floating windows forward.

## Motivation

Floating editor windows created with client-side decorations can be regular, minimizable windows with a logical parent. Treating those windows as native child windows on macOS gives AppKit too much control over z-order and minimization behavior. As a result, selecting, minimizing, or closing one floating window can affect the parent window or sibling floating windows incorrectly.

The fix separates logical parenting from native platform parenting for the affected window class. This keeps parent-relative behavior available to the engine while avoiding AppKit child-window behavior that is not appropriate for independent floating editor windows.

### DIFF
--- a/Source/Engine/Platform/Mac/MacWindow.cpp
+++ b/Source/Engine/Platform/Mac/MacWindow.cpp
@@ -39,6 +39,62 @@ inline bool IsWindowInvalid(Window* win)
     return !hasWindow || !win;
 }
 
+inline Window* GetEngineWindow(NSWindow* nativeWindow)
+{
+    WindowsManager::WindowsLocker.Lock();
+    for (auto* window : WindowsManager::Windows)
+    {
+        if (window && window->GetNativePtr() == nativeWindow)
+        {
+            WindowsManager::WindowsLocker.Unlock();
+            return window;
+        }
+    }
+    WindowsManager::WindowsLocker.Unlock();
+    return nullptr;
+}
+
+inline bool ShouldAttachToParent(const CreateWindowSettings& settings)
+{
+    // Cocoa child windows inherit parent z-order and minimize behavior.
+    // Minimizable regular windows need independent top-level window semantics.
+    return settings.Parent && (settings.Type != WindowType::Regular || !settings.AllowMinimize);
+}
+
+inline bool CanMakeKeyWindow(NSWindow* window)
+{
+    return window && [window isVisible] && ![window isMiniaturized] && [window canBecomeKeyWindow];
+}
+
+inline bool IsRegularWindowWithParent(Window* window, Window* parent)
+{
+    return window &&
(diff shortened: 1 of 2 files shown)


mafiesto4:
Just tested macOS part and it does improve Editor usage. Thanks!

mafiesto4:
I had to merge macOS changes only as it breaks window focus on Windows:

* Before:

https://github.com/user-attachments/assets/e5d50b91-789c-4345-8bf4-b0f5843cdb34
 
* After:

https://github.com/user-attachments/assets/8ebe66e2-fc19-4eb9-9385-3275c8b8e66c

ifromstone:
> I had to merge macOS changes only as it breaks window focus on Windows:
> 
> 
> 
> * Before:
> 
> 
> 
> https://github.com/user-attachments/assets/e5d50b91-789c-4345-8bf4-b0f5843cdb34
> 
>  
> 
> * After:
> 
> 
> 
> https://github.com/user-attachments/assets/8ebe66e2-fc19-4eb9-9385-3275c8b8e66c
> 
> 

Sorry about the Windows issue. I didn’t have access to a Windows machine at the time, so I wasn’t able to test it properly.

I have one now, though, so I’ll try to take another look at the issue when I get a chance.

Issue #294, opened 2021-03-03
Color in Debug.Log


honzapatCZ:
It would be nice to have some rich text formatting in the debug.log, for like highlighting etc.
related #293

stefnotch:
I have to admit, I usually don't see tools that have formatted debug log output. So maybe I'm missing the use-case for this.

Though, for what it's worth, I have seen tools that have proper object inspectors in the log output. Which is neat.

honzapatCZ:
> I have to admit, I usually don't see tools that have formatted debug log output. So maybe I'm missing the use-case for this.
> 
> Though, for what it's worth, I have seen tools that have proper object inspectors in the log output. Which is neat.

Debug.LogWarning("I couldnt deserialize prop: <color=ocean-blue>" + someProp + "</color> during deserialization an error has been thrown <color=red>" + e.Message + "</color>");

PR #4179, opened 2026-07-07
Added colors for Debug.log

CodeForBeauty:
Added color formatting in Debug.Log, as requested in #294 

The implementation is based on RichTextBox. And has the same syntax.
Formatting is only added on the LogEntry and is not applied to the expanded log info on the bottom of the window when log is selected. Let me know if it is needed there.

Some maybe important quirks:
- The formatting is implemented directly inside the LogEntry. There is no ability to disable it. I don't think anyone will need to have text: <color> in their logs. But if that's a concern let me know.
- Any other tags beside color are left in.
- The implementation ended up being a simpler version of the RichTextBox. So there is some duplicate code. If this is an issue, let me know how it should be resolved.
- Performance is worse. I'm not sure by how much. On my Windows machine I didn't feel any difference.

Tested on Windows11 machine.

### DIFF
--- a/Source/Editor/Windows/DebugLogWindow.cs
+++ b/Source/Editor/Windows/DebugLogWindow.cs
@@ -70,22 +70,61 @@ namespace FlaxEditor.Windows
 
         private class LogEntry : Control
         {
+            /// <summary>
+            /// Text block with only color available, no other styling.
+            /// </summary>
             private struct TextColorBlock
             {
+                /// <summary>
+                /// Color of this block.
+                /// </summary>
                 public Color TextColor;
+
+                /// <summary>
+                /// Range of the text.
+                /// </summary>
                 public TextRange Range;
-                public Float2 Location;
-                public Float2 Size;
+
+                /// <summary>
+                /// The text location and size.
+                /// </summary>
+                public Rectangle Bounds;
             }
 
+            /// <summary>
+            /// Rich text parsing context.
+            /// </summary>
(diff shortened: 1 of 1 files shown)


CodeForBeauty:
Here's the script I've used for testing:

```
using FlaxEngine;

namespace TestProj;

public class TestLog : Script
{
    public override void OnStart()
    {
        Debug.Log("Normal text, <color=red>red text</color>, <color=blue>blue text</color>");
        Debug.Log("Normal text, red text, blue text");
        Debug.Log("Test \ntext, <color=yellow>yellow</color>");
        Debug.Log("Test \ntext \ntext \ntext \ntext");
        Debug.Log("Really, really, really, really, really, really, really, really, really, really, really, really, really, really, really long log" +
            "Really, really, really, really, really, really, really, really, really, really, really, really, really, really, really long log" +
            "Really, really, really, really, really, really, really, really, really, really, really, really, really, really, really long log" +
            "Really, really, really, really, really, really, really, really, really, really, really, really, really, really, really long log");
        Debug.Log("<Color=pink>pink</color>, <b>otherTag</b>");
        Debug.Log("<color>empty</color><color=orange>orange</color>normal");
        Debug.LogWarning("<color=red>red</color>testing");
        Debug.Log("<color=red>red <color=yellow>yellow nested</color>");
    }
    
    public override void OnUpdate()
    {
        Debug.Log("Multiple log collapse test");
    }
}

```

PR #4181, opened 2026-07-14
Resolves undefined behavior when passing in a non locked mutex into `pthread_cond_wait`

### LINKED ISSUE #4170: [Bug] [Linux]: Single core pegged after entering/leaving Play Mode
## Description
One Linux (haven't tested on other OS's), after entering and leaving play mode a single core stays pinned at ~100% usage. The core will stay pinned until the editor is restarted.

## Environment

- Flax 1.12, built from source (Linux x64, Development)
- Arch Linux (kernel 7.0.13)
- Vulkan
- Wayland

### Steps to reproduce

1. Open a project in the editor on Linux
2. Hit Play to enter Play mode
3. Exit Play mode
4. Leave the editor idle and watch CPU usage

### Version

1.12

### Relevant logs

```shell

```

TyGrze:
## Problem
Closes #4170

After a build the cooker thread is kept alive and parks in `GameCookerImpl::ThreadFunction`'s loop calling `ThreadCond.Wait(mutex)` on a `mutex` that is never locked
POSIX requires `pthread_cond_wait` to be called with the mutex locked by the caller
https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_cond_wait.html
On glibc it returns `EPERM` immediately instead of blocking, so the idle thread busy-loops

## Solution
Lock the mutex before the loop so it is held whenever `Wait` is called (`Wait` releases and re-acquires it internally)
The idle thread now sleeps and is woken by `NotifyOne` for the next build

## Testing
- Built the editor (Linux x64, Development)
- Ran Cook and Run, let it finish, left the editor idle
- Confirmed the Game Cooker thread no longer pins a core at ~100% (`top -H`) and CPU returns to idle
- Reran a build afterwards to confirm the thread still wakes and cooks normally
- Had AI write a minimal repro (glibc 2.43) since Flax's `Wait` wrapper is `void` and discards `pthread_cond_wait`'s return code, so the error is invisible in-engine. The repro confirms: waiting on an unlocked recursive mutex returns EPERM ~11.8M times/sec on my machine instead of blocking. Locking first makes it block.

## Note

I am on vacation RN and do not have access to a Windows system to test this change on

### DIFF
--- a/Source/Editor/Cooker/GameCooker.cpp
+++ b/Source/Editor/Cooker/GameCooker.cpp
@@ -795,16 +795,17 @@ int32 GameCookerImpl::ThreadFunction()
     IsThreadRunning = true;
 
     CriticalSection mutex;
+    mutex.Lock();
     while (Platform::AtomicRead(&CancelThreadFlag) == 0)
     {
         if (IsRunning)
         {
             Build();
         }
-
         ThreadCond.Wait(mutex);
     }
 
+    mutex.Unlock();
     IsThreadRunning = false;
     return 0;
 }


TyGrze:
And by on vacation RN i mean until Nov so I cannot test this on Windows at all

mafiesto4:
Fixed in 82ee8713f509e52fbc006a25fb3f0f1361a79583. (will add you to Release Notes thos as a contributor)

PR #4209, opened 2026-08-10
Disable eye adaptation and motion blur by default

Duroxxigar:
This PR disables the two effects mentioned by default. This is based on discussions in the Discord. Disabling the eye adaptation by default was a very popular choice. Disabling the motion blur received some upvotes as well, some being from core devs - so I included it as well.

### DIFF
--- a/Source/Engine/Graphics/Enums.h
+++ b/Source/Engine/Graphics/Enums.h
@@ -1152,7 +1152,7 @@ API_ENUM(Attributes="Flags") enum class ViewFlags : uint64
     /// <summary>
     /// Default flags for materials/models previews generating.
     /// </summary>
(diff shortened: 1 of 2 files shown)


mafiesto4:
I think it should stay included inside View Flags, otherwise, users will need to turn it twice in order to make it work, which is problematic. So whne someone changed postfx settings the effect will work.

Later, I can add auto-upgrade for old projects so they remain old default value to avoid mismatches when loaded existing ames/projects.

Duroxxigar:
Alright, put the view flags back to how they were, however, I did leave Eye Adaptation off for the asset preview. I don't know of a scenario where someone wants that effect when in the asset window.

PR #4223, opened 2026-08-17
Add ASTC texture compression support on Linux

TyGrze:
Enables COMPILE_WITH_ASTC on Linux x64 so the editor can cook ASTC textures, which Android builds require.

The dep build follows the existing Mac case, since both produce a static .a archive

Differences for Linux:

- Drops the Apple-only universal build flags
  (ASTCENC_UNIVERSAL_BUILD/UNIVERSAL_BINARY/INVARIANCE).
- Adds ASTCENC_CLI=OFF, as the Windows case already does, so only the library is built.
- Adds CMAKE_POSITION_INDEPENDENT_CODE=ON, required because the archive is linked into a shared editor module.
- Registers x64 only

### DIFF
--- a/Source/ThirdParty/astc/astc.Build.cs
+++ b/Source/ThirdParty/astc/astc.Build.cs
@@ -19,6 +19,8 @@ public class astc : EngineDepsModule
             return true;
         case TargetPlatform.Mac:
             return options.Architecture == TargetArchitecture.ARM64;
+        case TargetPlatform.Linux:
+            return options.Architecture == TargetArchitecture.x64;
         default:
             return false;
         }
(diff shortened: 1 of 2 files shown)


TyGrze:
I don't think the `libastcenc.a` actually uploaded anywhere because I had to use `git push --no-verify` as I don't have access to the Flax LFS server.

I think someone with access to the LFS server needs to regenerate the binary and upload it to the LFS server for the tests to pass.

TyGrze:
Removed the binary as suggested by Goa on the discord
https://discord.com/channels/437989205315158016/509056735844106251/1538948414467997808

Issue #4100, opened 2026-05-15
[Bug]: Minimized window won't open and come back to front when it is supposed to after reloading scripts


xxSeys1:
### Description

See "Steps to reproduce".

This is more problematic than it might seem at first, because, by default editor settings, all assets that the user tries to open will open in that window that is no inaccessible.

https://github.com/user-attachments/assets/a7ac2247-8141-411e-8ea2-26271cf88af6

### Steps to reproduce

1. Open any asset of which the editor will close and reload during/ after script reload (I believe this applies to all assets, but I know for certain that this happens with behavior tree)
1.1 Make sure the asset editor is undocked in a separate window from the main editor
2. Minimize the asset editor window
3. Reload scripts (go to File -> Recompile Scripts)
4. Try to open the asset again (via Content Panel or any asset reference) or try to bring the asset window back to the front by clicking on it in the taskbar

### Version

master branch

### Relevant logs

```shell

```

PR #4227, opened 2026-08-17
Keep asset windows alive during scripts reload

GoaLitiuM:
Recreated asset editor windows are now reattached to the native window after scripts reload to keep the current state and Z-order of the windows untouched on platforms where window state can't be changed without user interaction (Wayland).

### DIFF
--- a/Source/Editor/CustomEditorWindow.cs
+++ b/Source/Editor/CustomEditorWindow.cs
@@ -1,8 +1,10 @@
 // Copyright (c) Wojciech Figat. All rights reserved.
 
+using System;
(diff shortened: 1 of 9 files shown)


xxSeys1:
Could this potentially fix #4100?

GoaLitiuM:
> Could this potentially fix #4100?

It is possible if the issue has something to do with the minimized state of the window (this PR doesn't attempt to restore those states anymore), though I have only tested these changes over Linux/Wayland.

GoaLitiuM:
Confirmed to fix #4100

PR #4230, opened 2026-08-18
Fix unsafe `StringView` of managed string

GoaLitiuM:
Accessing pinned managed string data is unsafe after pinning is over, but runtime doesn't necessarily move the data immediately, so it remains alive for a short time but (hopefully) long enough for us to make a unmanaged copy of the string. During long Wayland drag'and'drop operations, the original string data passed from managed runtime was more likely to not remain valid which often lead to crashes.

### DIFF
--- a/Source/Engine/Platform/SDL/SDLPlatform.Linux.cpp
+++ b/Source/Engine/Platform/SDL/SDLPlatform.Linux.cpp
@@ -203,8 +203,11 @@ namespace WaylandImpl
             if (inputData->GetType() == IGuiData::Type::Text)
             {
                 UnixFile file(fd);
-                StringAnsi text = StringAnsi(inputData->GetAsText());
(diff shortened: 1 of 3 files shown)


mafiesto4:
Also, it does fix issue while executing debug command where GC comes in between and collects that string:

https://github.com/FlaxEngine/FlaxEngine/blob/1146620d4f965f106ff25f3a08b7fbec1a33b45b/Source/Engine/Debug/DebugCommands.cpp#L442-L444

I just have one idea: could we make it a bit of conditional to use unsafe version (`MCore::String::GetChars`) for some performance-critical bindings? For example, `Render2D.DrawText` is used to draw UI texts and it will produce lots of garbage if we merge it. Or maybe should we use some thread-local buffer for such memory ops? Is it worth trying?

PR #4231, opened 2026-08-18
Add `Start/EndTextInput` for control to toggle character input events

### LINKED ISSUE #4198: [Bug]: WASD input axis partially broke between version 1.11 and 1.12 (on Linux at least)
### Description

Between version 1.11 and 1.12, it appears that something partially broke the WASD input as the default binding for the ASD keys no longer do anything, neither in the Viewport, nor when accessing the input axis via script. The W key works for some reason though

I tested the same project on Flax 1.11 and the issue was not there yet, so some change between 1.11 and 1.12 broke the input.
Given that this essentially breaks any traditional WASD movement, this issue is rather critical

### Steps to reproduce

For this issue, I am running Fedora 44 KDE and using the Flax Engine version 1.12.

To test for this issue, I have uploaded the project file here:
[Experiments.zip](https://github.com/user-attachments/files/30754572/Experiments.zip)

Even though this issue also affects the free cam default script and viewport that an empty project ships by default, this character Controller serves as yet another example

### Version

1.12

### Relevant logs

```shell

```

GoaLitiuM:
Some platforms may show virtual keyboard popups when text input is expected (like in KDE Plasma's Wayland compositor), which may alter how the input is passed to the applications.

This PR tries to improve the support with virtual keyboard by requiring character input events to be explicitly enabled when the UI control expects text input from the user by toggling the events with `StartTextInput` and `EndTextInput` methods. When the text input mode is disabled, keyboard input is expected to be passed down to the controls and input events with no interference from any system virtual keyboards.

Fixes #4198.

### DIFF
--- a/Source/Editor/Content/GUI/ContentView.cs
+++ b/Source/Editor/Content/GUI/ContentView.cs
@@ -674,6 +674,20 @@ namespace FlaxEditor.Content.GUI
         }
 
         /// <inheritdoc />
+        public override void OnGotFocus()
+        {
+            StartTextInput();
+            base.OnGotFocus();
+        }
+
+        /// <inheritdoc />
+        public override void OnLostFocus()
+        {
+            EndTextInput();
+            base.OnLostFocus();
+        }
+
+        /// <inheritdoc />
         public override bool OnMouseDown(Float2 location, MouseButton button)
         {
             if (base.OnMouseDown(location, button))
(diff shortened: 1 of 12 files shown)


mafiesto4:
Can we use it in the future for virtual keyboard display on mobile and consoles?

GoaLitiuM:
> Can we use it in the future for virtual keyboard display on mobile and consoles?

Should be possible.

There was one small issue with this I forgot to report, but in some rare case the text input doesn't end properly when I was changing focus from control to control while in play mode, but I couldn't reproduce it.

PR #4267, opened 2026-09-19
Unix filesystem: long-path safety

envision3d:
Replace the fixed 256-byte path buffers (and hard ASSERT on overflow) in DeleteUnixPathTree, GetChildDirectories, getFilesFromDirectoryTop, and getFilesFromDirectoryAll with 4096-byte buffers and a graceful skip, so deep content paths no longer trip the assertion.


envision3d:

PR #4268, opened 2026-09-19
MaterialInstance: save virtual assets via temp chunk

envision3d:
Virtual assets don't own a storage container, so SetChunk/GetOrCreateChunk are a no-op for them. Wire a temporary chunk into the header for the duration of MaterialInstance::Save (mirrors ModelBase::Save) and release it afterwards, so virtual material instances save correctly.


envision3d:

PR #4269, opened 2026-09-19
ModelInstanceActor: stash material override when model has no slots

envision3d:
When the model has no slots (not loaded/unresolved, e.g. replaying material overrides during deserialization before a procedural model is rebuilt), stash the override so a later model load reconciles it instead of dropping it.


envision3d:

PR #4270, opened 2026-09-19
GPU: benign async-task cancellation on resource release

envision3d:
Add GPUTask::CancelByRelease() to distinguish 'cancelled because the tracked GPU resource was released' (benign) from a real failure. The 4 GPU tasks (CopyResource, CopySubresource, UploadBuffer, UploadTextureMip) use it on release, GPUTasksContext suppresses the 'canceled before a sync' warning in that case, and GPUResourcePropertyBase::OnReleased now fires the Released action so bound tasks actually get notified. Stops spurious warnings / MissingResources failures when a resource is released while an async upload/copy is in flight.


envision3d:

